Issue #14, opened 2020-03-30
Geo2rdr returns nan line-of-sight when provided a state vector


jlmaurer:
Example RAiDER command: 
raiderDelay.py --date 20200103 --time 23:00:00 -s state.txt -modelbb 40 -79 36 75 --model ERA5 --zref 15000 -v 

where state.txt is a text file containing a single line: 

20200103 311674.487622 2469192.551542 -6630254.862875 2388.91490 -6765.188049 -2408.216845

(i.e., date x y z vx vy vz)

The code of interest is in losreader.py, in the function state_to_los. **Note**: #15 only ensures that the inputs to state_to_los are correct, but it does not fix this issue. 

The line "los_x, los_y, los_z = geo2rdr_obj.get_los()" is where the NaNs first appear.

def state_to_los(t, x, y, z, vx, vy, vz, lats, lons, heights):
    from RAiDER import Geo2rdr

    real_shape = lats.shape
    lats = lats.flatten()
    lons = lons.flatten()
    heights = heights.flatten()

    geo2rdr_obj = Geo2rdr.PyGeo2rdr()
    geo2rdr_obj.set_orbit(t, x, y, z, vx, vy, vz)

    loss = np.zeros((3, len(lats)))
    slant_ranges = np.zeros_like(lats)

    for i, (lat, lon, height) in enumerate(zip(lats, lons, heights)):
        height_array = np.array(((height,),))

        # Geo2rdr is picky about the type of height
        height_array = height_array.astype(np.double)

        import pdb; pdb.set_trace()
        geo2rdr_obj.set_geo_coordinate(np.radians(360-lon),np.radians(lat),1, 1,height_array)

        # compute the radar coordinate for each geo coordinate
        geo2rdr_obj.geo2rdr()

        # get back the line of sight unit vector
        los_x, los_y, los_z = geo2rdr_obj.get_los()
        loss[:, i] = los_x, los_y, los_z

        # get back the slant ranges
        slant_range = geo2rdr_obj.get_slant_range()
        slant_ranges[i] = slant_range

    los = loss * slant_ranges

    # Have to think about traversal order here. It's easy, though, since
    # in both orders xs come first, followed by all ys, followed by all
    # zs.
    return los.reshape(real_shape + (3,))

piyushrpt:
Don't we need at least 4 state vectors to interpolate the orbit? This should be enforced when translating the state vectors to an orbit. Looks like this example uses a single state vector?

[...]

Issue #13, opened 2020-03-30
Should cython be a required dependency for RAiDER? 


jlmaurer:
As opposed to just writing some nice .cpp files and including those. Maybe it makes it easier, or not?

dbekaert:
even if they are c-code programs, would you not use cython wrappers to call them as python functionality?

piyushrpt:
Yeah - we would need either cython / pybind11 to make python talk to C++. Numpy now uses cython to build part of itself - so we should be getting cython functionality at no cost.

dbekaert:
Requirements in README reflect to keep cython and numpy.
Per discussion keeping cython and closing this issues.

PR #15, opened 2020-03-30
General code clean-up and setting up the ability to use state vectors

jlmaurer:
- Current implementation fails at line-of-sight generation from the orbit using Geo2rdr (see #14).
- Cleans up a number of deprecated comments and file reads in preparation for chunked data access
- Rearranges some utility functions to be compatible with using a state file and for grouping similar fcns
- removes the .cpp version of Geo2rdr from the repo, as it changes with every build and isn't required because we specify cython as a RAiDER dependency (this could be fixed; see #13).


dbekaert:
@jlmaurer do you want to wait to include the state-vector component too?

Issue #28, opened 2020-04-17
working zenith, model nodes, fixed height level example


dbekaert:
# Scenario:
Example: Running ERA5 at weather model nodes at fixed height over BBoX

Code used:
```
raiderDelay.py --date 20190415 --time 1200 --heightlvs 0 --model ERA5 --modelBBOX 36 -121 34 -119
```

# Errors to be fixed:
1. re-run of hgt verus dem enabled
**not an error:**  the code runs though and uses DEM if it was pre-existing. Does not use fixed height level.
**Solution:** By removing the "geom" folder the issue got removed.
see also #27. It mght be good to track what is in the geom folder an compare with a new call if its to be used or not.

2. Chunks size not of equal size
**error:** ValueError: "chunks" must have same rank as dataset shape
<details>

```raiderDelay.py --date 20190415 --time 1200 --heightlvs 0 --model ERA5 --modelBBOX 36 -121 34 -119 
HDF5 must be used with height levels
Weather model already exists, please remove it ("weather_files/ERA5_2019-04-15 12:00:00.h5") if you want to create a new one.
Traceback (most recent call last):
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/bin/raiderDelay.py", line 13, in <module>
    parseCMD()
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/runProgram.py", line 146, in parseCMD
    (_,_) = tropo_delay(los, lats, lons, heights, flag, weather_model, wmLoc, zref,
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/delay.py", line 182, in tropo_delay
    writePnts2HDF5(lats, lons, hgts, los, pnts_file, in_shape)
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/utilFcns.py", line 766, in writePnts2HDF5
    start_positions = f.create_dataset('Rays_SP', (len(x),3), chunks = los.chunks)
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
[...]

Issue #27, opened 2020-04-17
working Zenith, model node, DEM example


dbekaert:
# Scenario:
Example: Running ERA5 at weather model nodes at DEM height over BBoX

Code used:
```
raiderDelay.py --date 20190415 --time 1200  --model ERA5 --modelBBOX 36 -121 34 -119 
```

# Errors to be fixed:
1. re-run of hgt verus dem enabled
**Error:**  not very clear for a user to know what the issue is, and its caused by a user driver error:
**Solution:** By removing the "geom" folder the issue got removed.
<details>

```raiderDelay.py --date 20190415 --time 1200  --model ERA5 --modelBBOX 36 -121 34 -119 
Weather model already exists, please remove it ("weather_files/ERA5_2019-04-15 12:00:00.h5") if you want to create a new one.
/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/delayFcns.py:267: H5pyDeprecationWarning: dataset.value has been deprecated. Use dataset[()] instead.
  lengths = np.linalg.norm(f['LOS'].value, axis=-1)
Traceback (most recent call last):
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/bin/raiderDelay.py", line 13, in <module>
    parseCMD()
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/runProgram.py", line 146, in parseCMD
    (_,_) = tropo_delay(los, lats, lons, heights, flag, weather_model, wmLoc, zref,
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/delay.py", line 186, in tropo_delay
    computeDelay(weather_model_file, pnts_file, zref, out,
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/delay.py", line 111, in computeDelay
    wet, hydro = interpolateDelay(weather_model_file_name, pnts_file_name, zref = zref,
  File "/Users/dbekaert/Software/python/miniconda37/envs/RAiDER/lib/python3.8/site-packages/RAiDER/delay.py", line 69, in interpolateDelay
    RAiDER.delayFcns.calculate_rays(pnts_file_name, stepSize, verbose = verbose)
[...]

PR #34, opened 2020-04-29
Update the argparser and fix some bugs 

### LINKED ISSUE #26: Updates to raiderDelay.py argparse
##Update for user-friendliness needed in argparse:

1. Provide example for each input. E.g. ISO 8601 format not clear to all users. (e.g. multiple variants for parsing day and time exist online, clarify with e.g. YYYMMDD)
2. Specify to which reference the DEM and reference the height levels are. i.e. ellipsoid, geoid and  corrected for ellipsoid etc
3. The options used to be a bit more nested. i.e. LOS or statevector or zenith; there is no mention anymore of zenith option. Correct in argparse visualization
4. The options used to be a bit more nested: i.e. DEM or levels. Correct in argparse visualization
5. Unclear what the pickle file option is, lacks description and might be only relevant for testing.  
6. If the parallel option does not work, comment it out.
7. Also add which models and how to specify them. e.g. its caps sensitive. 
   For now only list the once work itself.
8. im not sure if –modelBBOX is the right word. Rather see it as a region bbox, where internally the model bbox is extended such interpolation works. 
9. Add in multiple run example
- Zenith for ERA5 at model grid nodes at topo
- Zenith for ERA5 at model grid nodes at fixed height level
- los for ERA 5 using Sentinel-1 orbit at model grid nodes at topo
- los for ERA 5 using ISCE los file at model grid nodes at topo

jlmaurer:
* Update the argument parser help message for clarity
* Use ERA5 superclass for ERA5T
* Fix bug where DEM downloads even when not needed
* Ensure argparse inputs are typed correctly
* Remove antiquated code
* Ensure that argparse inputs are case-insensitive
* Initial stab at ensuring CF-1.8 HDF5 compliance
* Should fix #26, also addresses issues raised in #27 and #28

### DIFF
--- a/tools/RAiDER/checkArgs.py
+++ b/tools/RAiDER/checkArgs.py
@@ -27,14 +27,10 @@ def checkArgs(args, p):
             if args.outformat.lower() != 'hdf5':
                 print('HDF5 must be used with height levels')
                 args.outformat= 'hdf5'
-    if args.model not in AllowedModels():
-       raise NotImplementedError('Model {} has not been implemented'.format(args.model))
-    if args.model == 'WRF' and args.files is None:
-       p.error('Argument --files is required with --model WRF')
 
     ## Area
     # flag depending on the type of input
-    if args.area is not None:
+    if args.latlon is not None:
         flag = 'files'
     elif args.bounding_box is not None:
         flag = 'bounding_box'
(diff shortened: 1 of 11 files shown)


dbekaert:
For some reason this did not trigger the circle CI to run a test. Will need to look into it.

jlmaurer:
Yes, I saw that, looked for it to start but didn't see anything. We can look into it tomorrow.

dbekaert:
circle ci is running on it

PR #42, opened 2020-06-04
Fixes a bug that deleted geometry files prior to writing them

jlmaurer:


jlmaurer:
No, I think it's actually not necessary because they get cleared anyway after the function exits. This was a holdover from when I was having memory issues so was keeping things cleared out, but the memory issue has been dealt with using the HDF5 implementation.

Issue #32, opened 2020-04-27
Change main delay calculation from simple for-loop to parallel for-loop


jlmaurer:
Code is tools/RAiDER/delayFcns.py: 
https://github.com/dbekaert/RAiDER/blob/8b397b153505b47ba98b3655fc92a33be40c9e1c/tools/RAiDER/delayFcns.py#L123

The code is chunked but it processes each chunk serially. A potentially major speed-up would be to add a parallel implementation here. Only reads are required; no data is written, but I'm not sure what happens if you try to open an HDF5 file in read-only mode (see perhaps [here](https://stackoverflow.com/questions/34906652/does-hdf5-support-concurrent-reads-or-writes-to-different-files) for a start). My hope is that we can just pass the file name to different processes (and if the file is chunked properly so we never try to read the same chunk) we can just use multiprocessing and do it that way. I.e. something like: 

```
zipped_args = zip([filename]*len(other_args), other_arg_1, other_arg_2,...)
for arg_set in zipped_args:
    open the file in read-only mode and read the appropriate chunk
    do the delay calculation
```
But I don't know if HDF5 will actually allow that to work.

jlmaurer:
PR #45 addresses this issue.

PR #45, opened 2020-06-15
hdf5 parallel along with a few clean-ups

leiyangleon:


test/test_hdf5_parallel.py
```diff
@@ -0,0 +1,54 @@
...
+    # test error messaging
+    def test_get_delays_wet_accuracy(self):
+        self.assertTrue(np.allclose(self.delays_wet_1, self.delays_wet_4))
+
+    
+    def test_get_delays_hydro_accuracy(self):
+        self.assertTrue(np.allclose(self.delays_hydro_1, self.delays_hydro_4))
+
+    
+    def test_get_delays_runtime(self):
```
jlmaurer, line 44: I'm thinking that instead of testing for the parallel time being less than the serial time, we simply print out the speedup.

tools/RAiDER/delay.py
```diff
@@ -102,9 +102,12 @@ def computeDelay(weather_model_file_name, pnts_file_name, useWeatherNodes = Fals
         import h5py
         # Get the weather model data
         with h5py.File(weather_model_file_name, 'r') as f:
-            zs_wm = f['z'].value.copy()
-            total_wet=f['wet_total'].value.copy()
-            total_hydro=f['hydro_total'].value.copy()
+#             zs_wm = f['z'].value.copy()
```
jlmaurer, line 105: Just delete these three commented lines

tools/RAiDER/delayFcns.py
```diff
@@ -63,56 +67,68 @@ def testTime(t, ray, N):
...
+    t0 = time.time()
 
     # Get the weather model data
     with h5py.File(wm_file, 'r') as f:
-        xs_wm = f['x'].value.copy()
-        ys_wm = f['y'].value.copy()
-        zs_wm = f['z'].value.copy()
-        wet=f['wet'].value.copy()
-        hydro=f['hydro'].value.copy()
+#        xs_wm = f['x'].value.copy()
```
jlmaurer, line 79: Similar to above, go ahead and delete these commented lines

tools/RAiDER/delayFcns.py
```diff
@@ -124,9 +140,53 @@ def get_delays(stepSize, pnts_file, wm_file, interpType = '3D',
 #        pass
 
 #    return wet, hydro
+    time_elapse = (time.time() - t0)
```
jlmaurer, line 143: The tqdm function in the above for-loop tracks time usage, so I suggest to remove the explicit time-checking.

tools/RAiDER/delayFcns.py
```diff
@@ -124,9 +140,53 @@ def get_delays(stepSize, pnts_file, wm_file, interpType = '3D',
...
+    time_elapse_hr = int(np.floor(time_elapse/3600.0))
+    time_elapse_min = int(np.floor((time_elapse - time_elapse_hr*3600.0)/60.0))
+    time_elapse_sec = (time_elapse - time_elapse_hr*3600.0 - time_elapse_min*60.0)
+    print("Delay estimation cost {0} hour(s) {1} minute(s) {2} second(s) using {3} cpu threads".format(time_elapse_hr,time_elapse_min,time_elapse_sec,cpu_num))
+#    sio.savemat('test.mat',{'wet_delay':wet_delay,'hydro_delay':hydro_delay})
+    sio.savemat('get_delays_time_elapse.mat',{'time_elapse':time_elapse})
```
jlmaurer, line 149: Should remove the savemat call

tools/RAiDER/delayFcns.py
```diff
@@ -124,9 +140,53 @@ def get_delays(stepSize, pnts_file, wm_file, interpType = '3D',
...
+        map() only handles functions that take a single argument, and (2)
+        this function can generally be imported from a module, as required
+        by map().
+        """
+    
+    from pyproj import Transformer, CRS
+    
+    k, chunkInds, SP, SLV, in_shape, stepSize, ifWet, ifHydro, max_len, wm_file = tup
+    
+    # Transformer from ECEF to weather model
```
jlmaurer, line 168: It might be more efficient to create this outside the for-loop and pass a copy of the transformer with each tup.

tools/RAiDER/delayFcns.py
```diff
@@ -207,7 +267,8 @@ def getProjFromWMFile(wm_file):
     '''
     from pyproj import CRS 
     with h5py.File(wm_file, 'r') as f:
-        wm_proj = CRS.from_json(f['Projection'].value)
+#        wm_proj = CRS.from_json(f['Projection'].value)
```
jlmaurer, line 270: Delete this commented line

tools/RAiDER/delayFcns.py
```diff
@@ -249,7 +310,8 @@ def lla2ecef(pnts_file):
 
     t = Transformer.from_crs(4326,4978, always_xy =True) # converts from WGS84 geodetic to WGS84 geocentric
     with h5py.File(pnts_file, 'r+') as f:
-        sp = np.moveaxis(np.array(t.transform(f['lon'].value, f['lat'].value, f['hgt'].value)), 0, -1)
+#        sp = np.moveaxis(np.array(t.transform(f['lon'].value, f['lat'].value, f['hgt'].value)), 0, -1)
```
jlmaurer, line 313: delete this commented line

tools/RAiDER/delayFcns.py
```diff
@@ -258,7 +320,8 @@ def getUnitLVs(pnts_file):
     '''
     get_lengths(pnts_file)
     with h5py.File(pnts_file, 'r+') as f:
-        slv = f['LOS'].value / f['Rays_len'].value[...,np.newaxis]
+#        slv = f['LOS'].value / f['Rays_len'].value[...,np.newaxis]
```
jlmaurer, line 323: delete this commented line

tools/RAiDER/delayFcns.py
```diff
@@ -274,7 +337,8 @@ def get_lengths(pnts_file):
                      meters of the top of the atmosphere from the ground pnt. 
     '''
     with h5py.File(pnts_file, 'r+') as f:
-        lengths = np.linalg.norm(f['LOS'].value, axis=-1)
+#        lengths = np.linalg.norm(f['LOS'].value, axis=-1)
```
jlmaurer, line 340: delete this commented line

test/test_hdf5_parallel.py
leiyangleon: For now, each time get_delays is executed, the runtime is printed out. So for this unit test, the runtimes for both serial and parallel time have already been printed out. Do you still want to print out the speedup?

jlmaurer: I think we can put the time printout in the get_delay fcn under the verbose option (or logging once we get it). If we do that, then we can just delete this unit test. I'm unsure whether it's a good idea to have a unit test fail if there is no speedup, since we don't know all possible system setups.

tools/RAiDER/delayFcns.py
jlmaurer: I think we can put the time printout in the get_delay fcn under the verbose option (or logging once we get it).

tools/RAiDER/delayFcns.py
jlmaurer: Based on my conversation with Yang, it looks like the transformer object must be created inside each process; otherwise it crashes.

tools/RAiDER/delayFcns.py
jlmaurer: Instead of a .mat file, we'll save a .txt file to the local dir

tools/RAiDER/delayFcns.py
jlmaurer: Re Yang and my conversation, tqdm isn't relevant for the parallel for-loop, so we'll just use the start and end times but them under a if verbose flag.

leiyangleon:
@dbekaert @jlmaurer All the changes required by @jlmaurer have been fixed. Please review/merge and feel free to let me know if questions.

jlmaurer:
@dbekaert I approved the updates for my part; you can review and merge if you wish.

jlmaurer:
This PR is the solution to #32.

PR #58, opened 2020-06-25
Basic code formatting

Askaholic:
Only addressing the most basic code formatting issues:
1. Sort imports (removing unused imports)
2. Fix whitespace issues (with the following command)
```
autopep8 tools/RAiDER/ --recursive --select E231,E225,E251,E111,E128,E127,E265,E303,E261,E301 --in-place
```

I did not rename any functions or move anything around. One step at a time.


tools/RAiDER/models/ecmwf.py
```diff
@@ -1,15 +1,16 @@
 import datetime
+
 import numpy as np
 from pyproj import CRS
 
-from RAiDER.utilFcns import round_date
```
jlmaurer, line 5: I'm not sure why this is removed, since it looks to me like the function is called in line 123 (in my version) in the _get_from_ecmwf function?

Askaholic: I see what you mean. On that line I get a warning "Undefined name 'util'". I can add `import RAiDER.utilFcns as util` at the top of the file.

PR #60, opened 2020-06-26
Rename test files to remove colons from the file path

### LINKED ISSUE #59: Windows does not like files with `:` in the name
These files are giving me "invalid path" errors on Windows:
```
wrfout_d01_2010-06-25_00:00:00
wrfplev_d01_2010-06-25_00:00:00 
```

and I think it's because of the colons. https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

image

We might want to rename these if we want to support Windows users in the future.

Askaholic:
They're not being used by any code yet.

Fixes #59


tools/RAiDER/utilFcns.py
```diff
@@ -401,20 +405,35 @@ def Chunk(iterable, n):
...
+    )
 
     hydro_file_name = os.path.join(out, hydroname)
     wet_file_name = os.path.join(out, wetname)
     return wet_file_name, hydro_file_name
 
 
+def make_weather_model_filename(name, time, ll_bounds):
+    return '{}_{}_{}N_{}N_{}E_{}E.h5'.format(
+        name, time.strftime("%Y-%m-%d %H_%M_%S"), *ll_bounds
```
jlmaurer, line 433: @Askaholic can you make this "%Y-%m-%dT%H_%M_%S"?

Askaholic: Done

jlmaurer:
@dbekaert can you check this PR as I need it before pushing the unit tests for checking values. Re #59 the WRF file naming convention the problem is only that we cannot have files with colons in the repo; it is irrelevant for files that a user brings.

dbekaert:
@Askaholic independent of a checkout, can the code run with : in the filename on a windows system.

dbekaert:
see my comments on #59

Issue #46, opened 2020-06-15
raiderStats help menu needs cleaned up


jlmaurer:
The help message is currently a bit cluttered and should be grouped into "plot types," "input options," etc. 

# Current functionality:

<details>

```
usage: raiderStats.py [-h] -f FNAME [-c COL_NAME] [-fmt PLOT_FMT]
                      [-cb CBOUNDS CBOUNDS] [-w WORKDIR] [-b BBOX] [-sp SPACING]
                      [-dt DENSITYTHRESHOLD] [-sg] [-dg]
                      [-cp COLORPERCENTILE COLORPERCENTILE] [-ti TIMEINTERVAL]
                      [-si SEASONALINTERVAL] [-station_distribution]
                      [-station_delay_mean] [-station_delay_stdev]
                      [-grid_heatmap] [-grid_delay_mean] [-grid_delay_stdev]
                      [-variogramplot] [-binnedvariogram] [-plotall] [-verbose]

Function to generate various quality control and baseline figures of the spatial-
temporal network of products.

optional arguments:
  -h, --help            show this help message and exit
  -f FNAME, --file FNAME
                        csv file
  -c COL_NAME, --column_name COL_NAME
                        Name of the input column to plot. Input assumed to be in
                        units of meters
  -fmt PLOT_FMT, --plot_format PLOT_FMT
                        Plot format to use for saving figures
  -cb CBOUNDS CBOUNDS, --color_bounds CBOUNDS CBOUNDS
                        List of two floats to use as color axis bounds
  -w WORKDIR, --workdir WORKDIR
                        Specify directory to deposit all outputs. Default is
                        local directory where script is launched.
  -b BBOX, --bbox BBOX  Provide either valid shapefile or Lat/Lon Bounding SNWE.
                        -- Example : '19 20 -99.5 -98.5'
  -sp SPACING, --spacing SPACING
                        Specify spacing of grid-cells for statistical analyses.
                        By default 1 deg.
  -dt DENSITYTHRESHOLD, --densitythreshold DENSITYTHRESHOLD
                        A given grid-cell is only valid if it contains this
                        specified threshold of stations. By default 10 stations.
  -sg, --stationsongrids
                        In gridded plots, superimpose your gridded array with a
                        scatterplot of station locations.
  -dg, --drawgridlines  Draw gridlines on gridded plots.
  -cp COLORPERCENTILE COLORPERCENTILE, --colorpercentile COLORPERCENTILE COLORPERCENTILE
[...]

PR #61, opened 2020-06-30
Added option to download zenith delays virtually

sssangha:
Added end-to-end option to download/virtually access zenith delay information from the UNR Geodetic Laboratory and derive/plot statistics using the ```raiderStats.py``` program: 
1. To access zenith delay information, run ```raiderDownloadGNSS.py``` and refer to child scripts ```downloadGNSSDelays.py``` and ```getStationDelays.py```. By default, URLs from the UNR Geodetic Laboratory repository are passed to CSV named ```gnssStationList_overbbox_withpaths.csv```, before zenith delay information is accessed virtually and appended to series of CSVs deposited under ```GPS_delays``` subdirectory and then in turn appended to a master CSV entitled ```CombinedGPS_ztd.csv``` that records all zenith delay information for a specified time. This latter CSV may be directly passed to ```raiderStats.py``` to derive statistics for zenith delay information, or alternatively ```gnssStationList_overbbox.csv``` alone may be passed to just display distribution of GNSS stations. 

2. ```—download``` flag must be specified to physically download data, which is no longer necessary as data may be accessed virtually from the UNR Geodetic Laboratory. 

3. Input list of stations supported with ``` --station_file``` flag. In case both the bounding box and list of stations are specified, stations which fall outside the bounding box are printed in a warning message. 

Series of up updates made to ```statsPlot.py``` program: 
1. Gray background plotted behind grid-cells in case gridded arrays are plotted, and clean white backgrounds plotted behind stations in case just stations are plotted. 

2. Following issue #46, help menu/parser better organized. 

3. “PlateCarree” projection passed consistently in all map plot calls to address issues with irregular grid-cells in plots. 

4. Units of column for which statistics will be performed may now be specified in the commandline with argument –unit. 

Example end-to-end subroutine: 
#Call to download all data acquired at midnight UTC over CA spanning between 2016-2018 ```raiderDownloadGNSS.py -y '2016,2018' -b '38.1 38.3 237.4 237.6' -v --out products --returntime 00:00:00'``` 

#Call to plot distribution of stations and gridded station mean delay in a specific time interval: ```raiderStats.py -f products/CombinedGPS_ztd.csv -grid_delay_mean -station_distribution -ti '2016-01-01 2018-01-01'``` 

Changes yet to be implemented: 
1. Add option to explicitly retrieve spatial information for quick station plotting vs also retrieve zenith delay information. Currently the program forces the download of both. 

2. Parallelization for building CSV from remote data and plotting. Advice here in particular appreciated. @dbekaert, I wish to implement something along the lines of what is done to sort through products in the ARIA-tools ```ARIAProduct.py``` script 

3. Allow for shapefile input for spatial subset for download and/or plotting

### DIFF
--- a/tools/RAiDER/downloadGNSSDelays.py
+++ b/tools/RAiDER/downloadGNSSDelays.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python3
+import os
+import sys
+import requests
+import argparse
+import itertools
+import pandas as pd
+from RAiDER.getStationDelays import getStationData
+
+
+#base URL for UNR repository
+_UNR_URL = "http://geodesy.unr.edu/"
+
+
+def download_UNR(statID, year, writeDir = '.', baseURL = _UNR_URL, download = False, verbose = False):
+    '''
+    Download a zip file containing tropospheric delays for a given station and year
+    The URL format is http://geodesy.unr.edu/gps_timeseries/trop/<ssss>/<ssss>.<yyyy>.trop.zip
+    Inputs:
+        statID   - 4-character station identifier
+        year     - 4-numeral year
+    '''
+    URL = "{0}gps_timeseries/trop/{1}/{1}.{2}.trop.zip".format(baseURL, statID.upper(), year)
+    if download:
+        saveLoc = os.path.abspath(os.path.join(writeDir, '{0}.{1}.trop.zip'.format(statID.upper(), year)))
+        filepath = download_url(URL, saveLoc, verbose=verbose)
+    else:
+        filepath = check_url(URL, verbose=verbose)
+    return filepath
+
+
+def getStatsByllh(llhBox = None, baseURL = _UNR_URL, userstatList = None):
+    '''
+    Function to pull lat, lon, height, beginning date, end date, and number of solutions for stations inside the bounding box llhBox. 
+    llhBox should be a tuple with format (lat1, lat2, lon1, lon2), where lat1, lon1 define the lower left-hand corner and lat2, lon2 
+    define the upper right corner. 
+    '''
+    import pandas as pd
+    from urllib.request import urlopen 
+
+    if llhBox is None:
+        llhBox = [-90, 90, 0, 360]
+
+    stationHoldings = '{}NGLStationPages/DataHoldings.txt'.format(baseURL)
+    data = urlopen(stationHoldings) # it's a file like object and works just like a file
+    stations = []
+    for ind, line in enumerate(data): # files are iterable
+        if ind == 0:
+            continue
+        statID, lat, lon = getID(line)
+        # Only pass if in bbox
+        # And if user list of stations specified, only pass info for stations within list
+        if inBox(lat, lon, llhBox) and (not userstatList or statID in userstatList):
+            #convert lon into range [-180,180]
+            lon = fix_lons(lon)
+            stations.append({'ID': statID, 'Lat': lat, 'Lon': lon})
+    
+    print('{} stations were found'.format(len(stations)))
+    stations = pd.DataFrame(stations)
+    # Report stations from user's list that do not cover bbox
+    if userstatList:
+        userstatList=[i for i in userstatList if i not in stations['ID'].to_list()]
+        if userstatList:
+            print("Warning: The following user-input stations are not covered by the input bounding box {0}: {1}" \
+                .format(str(llhBox).strip('[]'),str(userstatList).strip('[]')))
+        
+    return stations
+
+
+def check_url(url, verbose = False):
+    '''
+    Check whether a file exists at a URL. Modified from 
+    https://stackoverflow.com/questions/9419162/download-returned-zip-file-from-url
+    '''
+    r = requests.get(url, stream=True)
+    if r.status_code==404:
+        return ''
+    else:
+        return url
+
+
+def download_url(url, save_path, verbose = False, chunk_size=128):
+    '''
+    Download a file from a URL. Modified from 
+    https://stackoverflow.com/questions/9419162/download-returned-zip-file-from-url
+    '''
+    r = requests.get(url, stream=True)
+    if r.status_code==404:
+        return ''
+    else:
+        if verbose:
+            print('Beginning download of {} to {}'.format(url, save_path))
+        with open(save_path, 'wb') as fd:
+            for chunk in r.iter_content(chunk_size=chunk_size):
+                fd.write(chunk)
+        if verbose:
+            print('Completed download of {} to {}'.format(url, save_path))
+        return save_path
+
+
+def readStationFile(filename):
+    '''
+    Read a list of GNSS station names from a plain text file
+    '''
+    statList = []
+    with open(filename, 'r') as f:
+        for line in f:
+            statList.append(line.strip())
+    return statList 
+
+
+def writeStationList(statList, filename = 'gnssStationList_overbbox.txt'):
+    '''
+    Write a python list of GNSS station names to a simple text file
+    '''
+    with open(filename, 'w') as f:
+        for stat in statList:
+            f.write('{}\n'.format(stat))
+
+
+def getStationList(bbox = None, writeLoc = None, userstatList = None):
+    '''
+    Creates a list of stations inside a lat/lon bounding box from a source
+    Inputs: 
+        bbox    - length-4 list of floats that describes a bounding box. Format is 
+                  S N W E
+    '''
+    if writeLoc is None:
+        writeLoc = os.path.join(os.getcwd(), 'gnssStationList_overbbox.csv')
+    else:
+        writeLoc = os.path.join(writeLoc, 'gnssStationList_overbbox.csv')
+
+    if userstatList:
+        with open(userstatList) as f:
+            userstatList = [line.rstrip('\n') for line in f]
+
+    statList = getStatsByllh(llhBox = bbox, userstatList = userstatList)
+
+    # write to file and pass final stations list
+    statList.to_csv(writeLoc, index=False)
+    stations = list(statList['ID'].values)
+
+    return stations, writeLoc
+
+
+def inBox(lat, lon, llhbox):
+    '''
+    Checks whether the given lat, lon pair are inside the bounding box llhbox
+    '''
+    lattest = lat < llhbox[1] and lat > llhbox[0]    
+    lontest = lon < llhbox[3] and lon > llhbox[2]
+    if lattest and lontest:
+        return True
+    else:
+        return False
+
+
+def fix_lons(lon):
+    """
+    Fix the given longitudes into the range ``[-180, 180]``.
+
+    """
+    fixed_lon = ((lon + 180) % 360) - 180
+    # Make the positive 180s positive again.
+    if fixed_lon == -180 and lon > 0:
+        fixed_lon *= -1
+
+    return fixed_lon
+
+
+def getID(line):
+    '''
+    Pulls the station ID, lat, and lon for a given entry in the UNR text file
+    '''
+    statID = str(line.split()[0].decode('utf-8'))
+    lat = float(line.split()[1].decode('utf-8'))
+    lon = float(line.split()[2].decode('utf-8'))
+
+    return statID, lat, lon
+
+
+def downloadTropoDelays(stats, years, writeDir = '.', download = False, verbose = False):
+    '''
+    Check for and download GNSS tropospheric delays from an archive. If download is True then 
+    files will be physically downloaded, which again is not necessary as data can be virtually accessed. 
+    '''
+
+    # argument checking
+    if not isinstance(stats, list):
+        if not isinstance(stats, str):
+            raise TypeError('stats should be a string or a list of strings')
+        stats = [stats]
+    if not isinstance(years, list):
+        if not isinstance(years, int):
+            try:
+                years = list(int(years))
+            except TypeError:
+                raise TypeError('years should be an int or a list of ints')
+        years = [years]
+
+    # Iterate over stations and years and check or download data
+    results = []
+    stat_year_tup = itertools.product(stats, years)
+    for s, y in stat_year_tup:
+        if verbose:
+            print('Currently checking station {} in {}'.format(s, y))
+        fileurl = download_UNR(s, y, writeDir = writeDir, download = download, verbose = verbose)
+        # only record valid path
+        if fileurl:
+            results.append({'ID': s, 'year': y, 'path': fileurl})
+    return pd.DataFrame(results).set_index('ID')
+            
+
+def parse_years(input):
+    '''
+    Takes string input and returns a list of years as integers
+    '''
+    test = len(input.split(',')) > 1
+    if test:
+        years = [int(y) for y in input.split(',')]
+        if years[1] > years[0] + 1:
+            years = [years[0] + k for k in range(years[1] - years[0] + 1)]
+    else:
+        years = [int(input)]
+    return years
+    
+
+def parse_args():
+    """Parse command line arguments using argparse."""
+    p = argparse.ArgumentParser(
+          formatter_class=argparse.RawDescriptionHelpFormatter,
+        description="""
+Check for and download tropospheric zenith delays for a set of GNSS stations from UNR
+Usage examples: 
+downloadGNSSdelay.py -y 2010 --check
+downloadGNSSdelay.py -y 2010 -b 40 -79 39 -78 -v
+downloadGNSSdelay.py -y 2010 -f station_list.txt --out products
+""")
+ 
+    p.add_argument(
+        '--years', '-y', dest='years', nargs='+',
+        help="""Year to check or download delays (format YYYY).
+Can be a single value or a comma-separated list. If two years non-consecutive years are given, download each year in between as well. 
+""", type=parse_years, required=True)
+
+    # Stations to check/download
+    area = p.add_argument_group('Stations to check/download. Can be a lat/lon bounding box or file, or will run the whole world if not specified')
+    area.add_argument(
+        '--station_file', '-f', default = None, dest='station_file',
+        help=('Text file containing a list of 4-char station IDs separated by newlines'))
+    area.add_argument(
+        '--BBOX', '-b', dest='bounding_box', type=str, default=None,
+        help="Provide either valid shapefile or Lat/Lon Bounding SNWE. -- Example : '19 20 -99.5 -98.5'")
+
+    misc = p.add_argument_group("Run parameters")
+    misc.add_argument(
+        '--outformat', 
+        help='GDAL-compatible file format if surface delays are requested.',
+        default=None)
+
+    misc.add_argument(
+        '--out', dest='out',
+        help='Directory to download products', 
+        default='.')
+
+    misc.add_argument(
+        '--returntime', dest='returnTime',
+        help="Return delays closest to this specified time. If not specified, the GPS delays for all times will be returned. Input in 'HH:MM:SS', e.g. '16:00:00'", 
+        default=None)
+
+    misc.add_argument(
+        '--download',
+        help='Physically download data. Note this is not necessary to proceed with statistical analysis, as data can be handled virtually in the program.',
+        action='store_true',dest='download', default = False)
+
+    misc.add_argument(
+        '--verbose', '-v',
+        help='Run in verbose (debug) mode? Default False',
+        action='store_true',dest='verbose', default = False)
+
+    return p.parse_args(), p
+
+
+def parseCMD():
+    """
+    Parse command-line arguments and pass to tropo_delay
+    We'll parse arguments and call delay.py.
+    """
+    args, p = parse_args()
+
+    # Create specified output directory if it does not exist.
+    if not os.path.exists(args.out):
+        os.mkdir(args.out)
+
+    # Setup bounding box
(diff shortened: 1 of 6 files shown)


tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    Pulls the station ID, lat, and lon for a given entry in the UNR text file
+    '''
+    statID = str(line.split()[0].decode('utf-8'))
+    lat = float(line.split()[1].decode('utf-8'))
+    lon = float(line.split()[2].decode('utf-8'))
+
+    return statID, lat, lon
+
+
+def downloadTropoDelays(stats, years, writeDir = '.', download = False, verbose = False):
```
jlmaurer, line 182: Hey @sssangha can you modify the codes so that the main functions are at the top of the file, with helpers such as getID, fix_lons, etc. toward the bottom? Something like parseCMD at the top, followed by downloadTropoDelays, followed by download_UNR, etc. This way these scripts will be consistent with the updates from @Askaholic.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+
+    misc.add_argument(
+        '--verbose', '-v',
+        help='Run in verbose (debug) mode? Default False',
+        action='store_true',dest='verbose', default = False)
+
+    return p.parse_args(), p
+
+
+def parseCMD():
```
jlmaurer, line 284: Since this function appears to be the main function for this script, perhaps we should name it something more descriptive? I think this function was originally in the version I wrote so it's my own fault for calling it this!

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # check for missing times
+        true_times = list(range(0,86400, 300))
+        if len(timesList)!=len(true_times):
+           missing = [True if t not in timesList else False for t in true_times]
+           mask = np.array(missing)
+           delay, sig, east_grad, north_grad = [np.full((288,), np.nan)]*4 
+           delay[~mask] = d
+           sig[~mask] = Sig
+           east_grad[~mask] = egrad
```
jlmaurer, line 104: We may want to add a note in the file as per Geoff Blewitt's comment that they have the north gradient and east gradient swapped in their files. From his email: 

"Thanks David.   At least we don't have any data loss, which is the most important thing.

Before I forget to mention it, please note that, due to a formatting error, our tropospheric gradient columns are interchanged.    That is, the column labeled as the East gradient is actually the North, and vice versa.   But you may be only using ZTD in which case, no problem."

If in the future anyone wants to use the gradients we could leave a comment directing them to check with the UNR folks about the proper format.

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
+#
+import copy
+import datetime as dt
+import numpy as np
+import os
+import pandas as pd
+
+def getDate(stationFile):
```
jlmaurer, line 9: Same comment here about ordering the functions from most to least important/first to last called.

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+    '''
+
+    if outDir is None:
+        outDir = os.getcwd()
+
+    pathbase = os.path.join(outDir, 'GPS_delays')
+    if not os.path.exists(pathbase):
+        os.mkdir(pathbase)
+
+    returnTime=(int(returnTime.split(':')[0])*3600)+(int(returnTime.split(':')[1])*60)+int(returnTime.split(':')[2])
```
jlmaurer, line 150: Might be a little cleaner to call this part that parses the returnTime as its own function

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
+#!/usr/bin/env python3
```
jlmaurer, line 1: @sssangha can you also run the following command on each of the modified files to take care of whitespaces consistent with PR#58:
autopep8 --select E231,E225,E251,E111,E128,E127,E265,E303,E261,E301 --in-place --recursive  <filename>

tools/RAiDER/getStationDelays.py
dbekaert: @jlmaurer @sssangha perhaps a broader question would be do we even need to gradient in east, north, up. We leverage the delays themselves nto the gradients. This might just add more overhead to our files that is not used by RAiDER.

tools/RAiDER/getStationDelays.py
Askaholic: Lots of ways to do this. Definitely extract it to a function though. Probably the most straight forward is:
```suggestion
    h, m, s = returnTime.split(":")
    returnTime = int(h * 3600 + m * 60 + s)
```

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+        if verbose:
+            print('Currently checking station {} in {}'.format(s, y))
+        fileurl = download_UNR(s, y, writeDir = writeDir, download = download, verbose = verbose)
+        # only record valid path
+        if fileurl:
+            results.append({'ID': s, 'year': y, 'path': fileurl})
+    return pd.DataFrame(results).set_index('ID')
+            
+
+def parse_years(input):
```
Askaholic, line 214: Not exactly sure how the edge cases are supposed to be handled, but I feel like you can make use of the 2 argument version of `range` i.e.
```python3
assert list(range(10, 10)) == []
assert list(range(10, 11)) == [10]
assert list(range(10, 15)) == [10, 11, 12, 13, 14]
assert list(range(20, 10)) == []
```

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    args, p = parse_args()
+
+    # Create specified output directory if it does not exist.
+    if not os.path.exists(args.out):
+        os.mkdir(args.out)
+
+    # Setup bounding box
+    if args.bounding_box:
+        if isinstance ([str(val) for val in args.bounding_box.split()], list) and not os.path.isfile(args.bounding_box):
```
Askaholic, line 297: I don't think it's possible for this `isinstance` check to fail. You are constructing a list, then checking if it is a list, and then throwing the list away again?

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
+#
+import copy
```
Askaholic, line 3: Unused import

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+import os
+import pandas as pd
+
+def getDate(stationFile):
+    '''
+    extract the date from a station delay file
+    '''
+
+    # find the date info
+    year = int(os.path.basename(stationFile).split('.')[1])
```
Askaholic, line 15: Again, split once and unpack to your variables

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+    at the closest times to to returnTime. 
+    Inputs: 
+         stationFile - a .gz station delay file 
+         returnTime  - specified time of GPS delay
+    Outputs:
+         a dict and CSV file containing the times and delay information
+         (delay in mm, delay uncertainty, delay gradients)
+    '''
+    #import functions
+    import gzip, zipfile, gzip, requests, io
```
Askaholic, line 35: Put at top of file

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+        f = gzip.open(ziprepo.open(j),'rb')
+        # get the date of the file
+        time, yearFromFile, doyFromFile = getDate(j)
+        # initialize variables
+        d, ngrad, egrad, timesList, Sig = [], [], [],[],[]
+        flag = False
+        for line in f.readlines():
+            try:
+                line = line.decode('utf-8')
+            except:
```
Askaholic, line 63:
```suggestion
            except UnicodeDecodeError:
```

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+                    trotot, trototSD, trwet, tgntot, tgntotSD, tgetot, tgetotSD, wvapor, wvaporSD, mtemp = \
+                        [float(t) for t in line.split()[2:]]
+                except:
+                    continue
+                site = line.split()[0]
+                year, doy, seconds = [int(n) for n in line.split()[1].split(':')]
+                # Break iteration if time from line in file does not match date reported in filename
+                if doy != doyFromFile:
+                   print('WARNING: time %s from line in conflict with time %s from file %s, will continue reading next tarfile(s)' \
+                       %(doy,doyFromFile,j))
```
Askaholic, line 81: use `.format` instead of `%`

tools/RAiDER/getStationDelays.py
```diff
@@ -0,0 +1,184 @@
...
+                ngrad.append(tgntot)
+                egrad.append(tgetot)
+                timesList.append(seconds)
+                Sig.append(trototSD)
+            if 'TROP/SOLUTION' in line:
+                flag = True
+        del f
+        # Break iteration if file contains no data.
+        if d == []:
+            print('WARNING: file %s for station %s is empty, will continue reading next tarfile(s)'%(j, j.split('.')[0]))
```
Askaholic, line 93: same

tools/RAiDER/statsPlot.py
```diff
@@ -400,11 +437,27 @@ def _getExtent(self):  # dataset, spacing=1, userbbox=None
...
+        return val*SI[unit_in]/SI[unit_out]
+
     def _reader(self):
         '''
         Read a input file
         '''
         data = pd.read_csv(self.fname)
+        #check if user-specified key is valid
+        if self.col_name not in data.keys():
+            raise Exception('User-specified key %s not found in inpuit file %s. Must specify valid key.'%(self.col_name,self.fname))
```
Askaholic, line 455: Use `.format`. Also typo in word `input`

tools/RAiDER/statsPlot.py
```diff
@@ -494,19 +547,24 @@ def __call__(self, gridarr, plottype, workdir='./', drawgridlines=False, colorba
...
-        axes.set_extent(self.plotbbox, ccrs.Geodetic())
-        axes.add_image(basemap, resValue, cmap='gray')
-        axes.coastlines()
+        fig, axes = plt.subplots(subplot_kw={'projection': ccrs.PlateCarree()})
+        # by default set background to white
+        axes.add_feature(cfeature.NaturalEarthFeature('physical', 'land', '50m', facecolor='white'), zorder=0)
+        # reset in case bounds too large
+        try:
+            axes.set_extent(self.plotbbox, ccrs.PlateCarree())
+        except:
```
Askaholic, line 563: Should either specify the error type or explicitly catch all exceptions with
```suggestion
        except Exception:
```

tools/RAiDER/getStationDelays.py
jlmaurer: @dbekaert True but the overhead seems likely to be minimal; as long as we're doing nothing but passing the values; seems like the info it could be useful to some users, but perhaps not.

tools/RAiDER/downloadGNSSDelays.py
jlmaurer: @Askaholic agreed, this is something we would like to do eventually with the other scripts as well. Perhaps next PR though. :)

tools/RAiDER/statsPlot.py
```diff
@@ -180,7 +216,7 @@ def resid(x, d, v, m):
         if Nparm is not None:
             lb = np.zeros(Nparm)
             x0 = (ub-lb)/2
-
+
```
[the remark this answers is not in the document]
jlmaurer: @Askaholic if you know a way to do this in vim you'll be my hero...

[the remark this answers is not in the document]
Askaholic: In VIM it's probably a bit more involved than ticking a checkbox, but if you're using VIM then you know what you signed up for :P.

https://vi.stackexchange.com/questions/454/whats-the-simplest-way-to-strip-trailing-whitespace-from-all-lines-in-a-file

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+                  S N W E
+    '''
+    if writeLoc is None:
+        writeLoc = os.path.join(os.getcwd(), 'gnssStationList_overbbox.csv')
+    else:
+        writeLoc = os.path.join(writeLoc, 'gnssStationList_overbbox.csv')
+
+    if userstatList:
+        with open(userstatList) as f:
+            userstatList = [line.rstrip('\n') for line in f]
```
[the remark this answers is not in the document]
sssangha: Nope, ```rstrip``` was a vestige, and ironically the read file function was treated as one too. Will replace this with a call to the read file function instead.

tools/RAiDER/downloadGNSSDelays.py
sssangha: True, restructured accordingly.

tools/RAiDER/getStationDelays.py
sssangha: Ok, added note/reference to that formatting error and made sure to flip the order of the initialized variables to account for this. I agree with @jlmaurer that we should just pass instead of omitting altogether

tools/RAiDER/getStationDelays.py
sssangha: Make sense, moved out to a separate function

tools/RAiDER/statsPlot.py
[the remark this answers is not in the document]
sssangha: I use gedit. Old habits die hard!

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def getStatsByllh(llhBox = None, baseURL = _UNR_URL, userstatList = None):
+    '''
+    Function to pull lat, lon, height, beginning date, end date, and number of solutions for stations inside the bounding box llhBox. 
+    llhBox should be a tuple with format (lat1, lat2, lon1, lon2), where lat1, lon1 define the lower left-hand corner and lat2, lon2 
+    define the upper right corner. 
+    '''
+    import pandas as pd
+    from urllib.request import urlopen
```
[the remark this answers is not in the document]
sssangha: Yeah, pandas was redundant here anyways.

And I don't believe there was a particular reason for choosing ```urlib``` over ```requests```. Any reason to proceed with the latter though @Askaholic?

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    https://stackoverflow.com/questions/9419162/download-returned-zip-file-from-url
+    '''
+    r = requests.get(url, stream=True)
+    if r.status_code==404:
+        return ''
+    else:
+        return url
+
+
+def download_url(url, save_path, verbose = False, chunk_size=128):
```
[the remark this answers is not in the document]
sssangha: These are smaller files anyways, but I will set to ```2048``` and test it out to see if it makes a difference.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    '''
+    Download a file from a URL. Modified from 
+    https://stackoverflow.com/questions/9419162/download-returned-zip-file-from-url
+    '''
+    r = requests.get(url, stream=True)
+    if r.status_code==404:
+        return ''
```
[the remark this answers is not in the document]
sssangha: This is intentional. I want to pass empty fields to the master CSV file so we can see specifically which years data from a given station are not available.

tools/RAiDER/downloadGNSSDelays.py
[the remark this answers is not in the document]
Askaholic: Mostly because `requests` is a higher level library. It will be able to handle stuff like redirects, cookies, etc out of the box. Even the documentation of `urllib` recommends using `requests` instead https://docs.python.org/3/library/urllib.request.html#module-urllib.request

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    else:
```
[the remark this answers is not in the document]
sssangha: True, but since I want to return empty fields in case the URL isn't valid (see my above comment) I suppose I could set the url variable to an empty string. I.e.:
``
    if r.status_code == 404:
        url = ''
    return url
``

tools/RAiDER/downloadGNSSDelays.py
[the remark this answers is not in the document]
Askaholic: Ok, IMO it would be cleaner to do the error handling in the calling function. Since `download_url` seems quite generic to me.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+
+def readStationFile(filename):
+    '''
+    Read a list of GNSS station names from a plain text file
+    '''
+    statList = []
+    with open(filename, 'r') as f:
+        for line in f:
+            statList.append(line.strip())
+    return statList
```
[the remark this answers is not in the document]
sssangha: Yeah, probably ```readTextFile``` should suffice. We should avoid ```readFile``` as that conflicts with a python utils module

tools/RAiDER/downloadGNSSDelays.py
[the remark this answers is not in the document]
Askaholic: If you simply delete the `else:` (And dedent the contents of the Else block) the function will behave exactly the same way. It’s a nice way of saving a level of indentation, which in turn makes it easier to read.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -0,0 +1,330 @@
...
+    '''
+    Creates a list of stations inside a lat/lon bounding box from a source
+    Inputs: 
+        bbox    - length-4 list of floats that describes a bounding box. Format is 
+                  S N W E
+    '''
+    if writeLoc is None:
+        writeLoc = os.path.join(os.getcwd(), 'gnssStationList_overbbox.csv')
+    else:
+        writeLoc = os.path.join(writeLoc, 'gnssStationList_overbbox.csv')
```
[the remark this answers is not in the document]
sssangha: Always hit my head when I miss an opportunity for a single-liner! :p

tools/RAiDER/downloadGNSSDelays.py
sssangha: Yeah, should've clarified that this is a placeholder for an option whereby a user may pass a valid shapefile, as is done in the ARIA-tools package. Will require more work to integrate, hence why I left off without implementing it yet.

tools/RAiDER/downloadGNSSDelays.py
sssangha: Ok, I was thinking of taking this and leveraging map as so:
```
def parse_years(timestr):
    '''
    Takes string input and returns a list of years as integers
    '''
    years = list(map(int, timestr.split(',')))
    # If two years non-consecutive years are given, query for each year in between
    if len(years) == 2:
         years = list(range(years[0],years[1]+1))
    return years
```

In this way, inputs non-consecutive pairs of years are expanded (e.g. '1994,1997' --> [1994 1995 1996 1997]), while inputs of multiple years are left alone (e.g. '2014,2015,2016' --> [2014 2015 2016])

tools/RAiDER/statsPlot.py
```diff
@@ -147,10 +180,13 @@ def _binnedVario(self, hEff, rawVario, xBin=None):
...
         for iBin in range(nBins):
             iBinMask = np.logical_and(xBin[iBin] < hEff, hEff <= xBin[iBin+1])
-
+            #circumvent indexing
             try:
-                hExp.append(np.nanmean(hEff[iBinMask]))
-                expVario.append(np.nanmean(rawVario[iBinMask]))
+                #supress mean of empty slice warning
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", category=RuntimeWarning)
```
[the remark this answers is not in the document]
sssangha: Yeah, ```warnings.filterwarnings("ignore", message="Mean of empty slice")``` suffices.

sssangha:
> Thanks for the contributions.
> @sssangha can you also update setup.py accordingly to install the new script?

Done in latest commit.

sssangha:
Ok, looks like I addressed all suggestions. As @jlmaurer suggested though, I will update the structure of the parser/main function in a separate PR, along with parallelization of downloading/plotting.

tools/RAiDER/downloadGNSSDelays.py
Askaholic: Very nice!

PR #62, opened 2020-07-01
Change gdalwarp command to gdaltranslate in demdownlaod

jlmaurer:


tools/RAiDER/demdownload.py
```diff
@@ -56,28 +57,29 @@ def download_dem(lats, lons, outLoc=None, save_flag='new', checkDEM=True,
         return hgts
 
     # Specify filenames
+    if verbose:
+        print('Getting the DEM')
+        import time
```
Askaholic, line 62: Just import it at the top of the file, it's a std library package, it will always exist.

tools/RAiDER/demdownload.py
```diff
@@ -56,28 +57,29 @@ def download_dem(lats, lons, outLoc=None, save_flag='new', checkDEM=True,
...
+        st = time.time()
+
     memRaster = '/vsimem/warpedDEM'
     inRaster = '/vsicurl/{}'.format(_world_dem)
-
-    ds = gdal.Open(inRaster)
-    gdalNDV = ds.GetRasterBand(1).GetNoDataValue()
```
dbekaert, line 63: @jlmaurer you removed the no-data definition but its used later in the code.

dbekaert:
@jlmaurer can you review the changes made.

tools/RAiDER/demdownload.py
dbekaert: pushed the change

tools/RAiDER/demdownload.py
dbekaert: pushed the change

tools/RAiDER/demdownload.py
```diff
@@ -19,11 +20,12 @@
 
 
 def download_dem(lats, lons, outLoc=None, save_flag='new', checkDEM=True,
-                 outName='warpedDEM.dem', ndv=0.):
+                 outName='warpedDEM.dem', ndv=0., verbose = False):
```
Askaholic, line 23:
```suggestion
                 outName='warpedDEM.dem', ndv=0., verbose=False):
```

jlmaurer:
@Askaholic @dbekaert this should fix the formatting/typos

dbekaert:
@jlmaurer Can you re-issue to  PR; existing changes in DEV should not be seen as new commits.

jlmaurer:
Well that was weird. It turns out the PR doesn't keep track of changes to the target branch. I rebased my branch before updating the PR, so that's why @sssangha's commit was showing up there. Changing the target branch and then back again fixes the problem.

PR #65, opened 2020-07-08
Fixed some bugs, cleaned up examples, and updated name of main functions

sssangha:
Implemented the following bug-fixes:
1. In ```statsPlot.py```, rescaled colorbar in cases lower and maximum bounds are the same, which would raise a ValueError when the colorbar is set up.
2. In ```statsPlot.py```, make sure the gridded array over the area of interest is an even multiple of the user-specified spacing (in degrees). If it is not, spacing defaults to 1 degree and the area of interest is dilated (if necessary) to ensure 1x1 degree grid-cells can span its entire area.
3. The topline of a function in ```getStationDelays.py``` was missing a colon.

Implemented the following enhancements:
1. As suggested in last PR, scripts are restructured with distinct, aptly named main functions ```queryRepos``` and ```statsAnalyses``` (alternative suggestions are welcome)!
2. Examples listed in ```downloadGNSSDelays.py``` updated and expanded.

### DIFF
--- a/tools/RAiDER/downloadGNSSDelays.py
+++ b/tools/RAiDER/downloadGNSSDelays.py
@@ -14,12 +14,12 @@ import sys
 
 import pandas as pd
 import requests
-from RAiDER.getStationDelays import getStationData
+from RAiDER.getStationDelays import get_station_data
 
 # base URL for UNR repository
 _UNR_URL = "http://geodesy.unr.edu/"
 
-def createParser():
+def create_parser():
     """Parse command line arguments using argparse."""
     p = argparse.ArgumentParser(
         formatter_class=argparse.RawDescriptionHelpFormatter,
@@ -48,12 +48,6 @@ geographic bounding box :
 downloadGNSSdelay.py --download --out products -y '2010,2014' --returntime '00:00:00' -b '40 -79 39 -78'
 """)
 
-    p.add_argument(
-        '--years', '-y', dest='years', nargs='+',
-        help="""Year to check or download delays (format YYYY).
-Can be a single value or a comma-separated list. If two years non-consecutive years are given, download each year in between as well. 
-""", type=parse_years, required=True)
-
     # Stations to check/download
     area = p.add_argument_group(
         'Stations to check/download. Can be a lat/lon bounding box or file, or will run the whole world if not specified')
@@ -66,14 +60,15 @@ Can be a single value or a comma-separated list. If two years non-consecutive ye
 
     misc = p.add_argument_group("Run parameters")
     misc.add_argument(
-        '--outformat',
-        help='GDAL-compatible file format if surface delays are requested.',
-        default=None)
-
-    misc.add_argument(
         '--out', dest='out',
         help='Directory to deposit outputs',
         default='.')
+
+    misc.add_argument(
+        '--years', '-y', dest='years', nargs='+',
+        help="""Year to check or download delays (format YYYY).
+Can be a single value or a comma-separated list. If two years non-consecutive years are given, download each year in between as well. 
+""", type=parse_years, required=True)
 
     misc.add_argument(
         '--returntime', dest='returnTime',
@@ -93,12 +88,12 @@ Can be a single value or a comma-separated list. If two years non-consecutive ye
     return p
 
 
-def cmdLineParse(iargs = None):
-    parser = createParser()
+def cmd_line_parse(iargs=None):
+    parser = create_parser()
     return parser.parse_args(args=iargs)
 
 
-def getStationList(bbox=None, writeLoc=None, userstatList=None):
+def get_station_list(bbox=None, writeLoc=None, userstatList=None):
     '''
     Creates a list of stations inside a lat/lon bounding box from a source
     Inputs: 
@@ -108,9 +103,9 @@ def getStationList(bbox=None, writeLoc=None, userstatList=None):
     writeLoc = os.path.join(writeLoc or os.getcwd(), 'gnssStationList_overbbox.csv')
 
     if userstatList:
-        userstatList = readTextFile(userstatList)
+        userstatList = read_text_file(userstatList)
 
-    statList = getStatsByllh(llhBox=bbox, userstatList=userstatList)
+    statList = get_stats_by_llh(llhBox=bbox, userstatList=userstatList)
 
     # write to file and pass final stations list
     statList.to_csv(writeLoc, index=False)
@@ -119,7 +114,7 @@ def getStationList(bbox=None, writeLoc=None, userstatList=None):
     return stations, writeLoc
 
 
-def getStatsByllh(llhBox=None, baseURL=_UNR_URL, userstatList=None):
+def get_stats_by_llh(llhBox=None, baseURL=_UNR_URL, userstatList=None):
     '''
     Function to pull lat, lon, height, beginning date, end date, and number of solutions for stations inside the bounding box llhBox. 
     llhBox should be a tuple with format (lat1, lat2, lon1, lon2), where lat1, lon1 define the lower left-hand corner and lat2, lon2 
@@ -135,10 +130,10 @@ def getStatsByllh(llhBox=None, baseURL=_UNR_URL, userstatList=None):
     for ind, line in enumerate(data.text.splitlines()):  # files are iterable
         if ind == 0:
             continue
-        statID, lat, lon = getID(line)
+        statID, lat, lon = get_ID(line)
         # Only pass if in bbox
         # And if user list of stations specified, only pass info for stations within list
-        if inBox(lat, lon, llhBox) and (not userstatList or statID in userstatList):
+        if in_box(lat, lon, llhBox) and (not userstatList or statID in userstatList):
             # convert lon into range [-180,180]
             lon = fix_lons(lon)
             stations.append({'ID': statID, 'Lat': lat, 'Lon': lon})
@@ -156,7 +151,7 @@ def getStatsByllh(llhBox=None, baseURL=_UNR_URL, userstatList=None):
     return stations
 
 
-def downloadTropoDelays(stats, years, writeDir='.', download=False, verbose=False):
+def download_tropo_delays(stats, years, writeDir='.', download=False, verbose=False):
     '''
     Check for and download GNSS tropospheric delays from an archive. If download is True then 
     files will be physically downloaded, which again is not necessary as data can be virtually accessed. 
@@ -231,7 +226,7 @@ def check_url(url, verbose=False):
     return url
 
 
-def readTextFile(filename):
+def read_text_file(filename):
     '''
     Read a list of GNSS station names from a plain text file
     '''
@@ -239,7 +234,7 @@ def readTextFile(filename):
         return [line.strip() for line in f]
 
 
-def inBox(lat, lon, llhbox):
+def in_box(lat, lon, llhbox):
     '''
     Checks whether the given lat, lon pair are inside the bounding box llhbox
     '''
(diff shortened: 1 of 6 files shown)


tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -309,3 +274,64 @@ def parse_years(timestr):
...
+                   outDir=inps.out, returnTime=inps.returnTime)
+
+    if inps.verbose:
+        print('Completed processing')
+
+
+if __name__ == "__main__":
+    inps = cmdLineParse()
+
+    queryRepos(inps)
```
Askaholic, line 337: I would unpack the arguments here, so that if this were to get called by another python program, that program would not need to somehow construct an object with attributes for the parameters. 
```suggestion
    queryRepos(
        inps.bounding_box,
        inps.verbose,
        ...
    )
```

tools/RAiDER/statsPlot.py
```diff
@@ -444,7 +443,7 @@ class raiderStats(object):
     # import dependencies
     import glob
 
-    def __init__(self, filearg, col_name, workdir='./', bbox=None, spacing=1, timeinterval=None, seasonalinterval=None, stationsongrids=False, colorpercentile='25 95', verbose=False):
+    def __init__(self, filearg, col_name, unit='m', workdir='./', bbox=None, spacing=1, timeinterval=None, seasonalinterval=None, stationsongrids=False, colorpercentile='25 95', verbose=False):
```
Askaholic, line 446: Class names should be in `PascalCase`.

tools/bin/raiderDownloadGNSS.py
```diff
@@ -7,9 +7,9 @@
...
 
-from RAiDER.downloadGNSSDelays import parseCMD
+from RAiDER.downloadGNSSDelays import cmdLineParse,queryRepos
 
 if __name__ == "__main__":
+    inps = cmdLineParse()
 
-    # Main engine
-    parseCMD()
+    queryRepos(inps)
```
Askaholic, line 15: Same comment as above

tools/bin/raiderStats.py
```diff
@@ -7,9 +7,10 @@
...
-from RAiDER.statsPlot import parseCMD
+#from RAiDER.statsPlot import parseCMD
+from RAiDER.statsPlot import cmdLineParse,statsAnalyses
 
 if __name__ == "__main__":
+    inps = cmdLineParse()
 
-    # Main engine
-    parseCMD()
+    statsAnalyses(inps)
```
Askaholic, line 16: Same here

tools/RAiDER/statsPlot.py
```diff
@@ -710,16 +713,23 @@ def __call__(self, gridarr, plottype, workdir='./', drawgridlines=False, colorba
             if cbounds is None:
                 cbounds = [np.nanpercentile(gridarr, self.colorpercentile[0]), np.nanpercentile(
                     gridarr, self.colorpercentile[1])]
+                # if upper/lower bounds identical, overwrite lower bound as 75% of upper bound to avoid plotting ValueError
+                if cbounds[0]==cbounds[1]:
```
Askaholic, line 717: Use a linter and/or autoformatter
```suggestion
                if cbounds[0] == cbounds[1]:
```

tools/RAiDER/statsPlot.py
```diff
@@ -710,16 +713,23 @@ def __call__(self, gridarr, plottype, workdir='./', drawgridlines=False, colorba
...
+                    cbounds[0]*=0.75
```
Askaholic, line 718:
```suggestion
                    cbounds[0] *= 0.75
```

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -19,16 +19,33 @@
 # base URL for UNR repository
 _UNR_URL = "http://geodesy.unr.edu/"
 
-def parse_args():
+def createParser():
```
Askaholic, line 22: BTW the standard in python is to use `snake_case` for function and variable names.

tools/RAiDER/statsPlot.py
```diff
@@ -620,17 +628,17 @@ def __call__(self, gridarr, plottype, workdir='./', drawgridlines=False, colorba
...
         '''
-
+        import cartopy.crs as ccrs
+        import cartopy.feature as cfeature
+        import cartopy.io.img_tiles as cimgt
         import matplotlib as mpl
+        import matplotlib.ticker as mticker
+        from cartopy.mpl.ticker import LatitudeFormatter, LongitudeFormatter
+        from mpl_toolkits.axes_grid1 import make_axes_locatable
+        from pandas.plotting import register_matplotlib_converters
```
dbekaert, line 638: @sssangha can you update the conda requirements file for any dependencies needed by the stats class?

tools/RAiDER/downloadGNSSDelays.py
jlmaurer: Agree with @Askaholic

tools/RAiDER/statsPlot.py
```diff
@@ -465,21 +464,30 @@ def __init__(self, filearg, col_name, workdir='./', bbox=None, spacing=1, timein
...
-                extent = [np.floor(self.bbox[2]-(self.spacing/2)), np.ceil(self.bbox[-1]+(self.spacing/2)),
-                          np.floor(self.bbox[0]-(self.spacing/2)), np.ceil(self.bbox[1]+(self.spacing/2))]
+                extent = [np.floor(self.bbox[2]), np.ceil(self.bbox[-1]),np.floor(self.bbox[0]), np.ceil(self.bbox[1])]
             else:
-                print(
-                    "WARNING: User-specified bounds do not overlap with dataset bounds, proceeding with the latter.")
+                print("WARNING: User-specified bounds do not overlap with dataset bounds, proceeding with the latter.")
             del dfextents_poly, userbbox_poly
 
+        # ensure that extents do not exceed -180/180 lon and -90/90 lat
```
jlmaurer, line 480: Should we be changing extents or raising exception? Seems that if points are requested outside the bounds might be a bigger problem.

tools/RAiDER/statsPlot.py
jlmaurer: Also I'm nervous about the cartopy dependency; do you know if there are any conflicts with GDAL?

tools/bin/raiderDownloadGNSS.py
sssangha: Since ```queryRepos``` is a function though, shouldn't it follow this convention ```query_repos```

tools/bin/raiderStats.py
sssangha: Same as above ```statsAnalyses``` is a function

tools/RAiDER/statsPlot.py
sssangha: @jlmaurer in my experience no, I haven't experienced any conflicts with GDAL.

tools/RAiDER/statsPlot.py
sssangha: @jlmaurer  This is not just a means to dummy-proof it for user error, but also deal with rounding/precision errors which give an invalid value like -91 lat or 182 lon. Would it be better to raise an exception, or just smooth out such errors under the hood?

sssangha:
> I don't think we have any unit tests for any of the raiderStats functionality. This is not a problem so much as an important step we should try to add at some point.

Yeah, I was thinking about pulling some of the examples from the parser and integrating into the unit tests. Maybe more appropriate in a separate PR, as goes for another planned addition I plan to make concerning input of shapefiles for querying the UNR database.

dbekaert:
> > I don't think we have any unit tests for any of the raiderStats functionality. This is not a problem so much as an important step we should try to add at some point.
> 
> Yeah, I was thinking about pulling some of the examples from the parser and integrating into the unit tests. Maybe more appropriate in a separate PR, as goes for another planned addition I plan to make concerning input of shapefiles for querying the UNR database.

Agreed

tools/RAiDER/statsPlot.py
jlmaurer: Personally I think we should raise an exception and then handle the exceptions externally. If @Askahloic or @dbekaert have other thoughts it's not a strong opinion

jlmaurer: *@Askaholic

tools/bin/raiderDownloadGNSS.py
Askaholic: This was in reference to the comment about unpacking arguments. But you are right, the convention would be to call it `query_repos` so that could be changed too.

tools/bin/raiderStats.py
Askaholic: Also, about unpacking arguments

tools/RAiDER/statsPlot.py
Askaholic: Maybe split the difference. If the user passes bad bounds in then raise an exception, but if you are computing the bounds then fixup the rounding errors.

sssangha: I agree with @Askaholic, went ahead and now raise an exception for cases where the user-specified bounds are bad.

tools/RAiDER/statsPlot.py
```diff
@@ -454,32 +453,47 @@ def __init__(self, filearg, col_name, workdir='./', bbox=None, spacing=1, timein
...
             userbbox_poly = Polygon(np.column_stack((np.array([self.bbox[2], self.bbox[3], self.bbox[3], self.bbox[2], self.bbox[2]]),
                                                      np.array([self.bbox[0], self.bbox[0], self.bbox[1], self.bbox[1], self.bbox[0]]))))
             if userbbox_poly.intersects(dfextents_poly):
-                extent = [np.floor(self.bbox[2]-(self.spacing/2)), np.ceil(self.bbox[-1]+(self.spacing/2)),
-                          np.floor(self.bbox[0]-(self.spacing/2)), np.ceil(self.bbox[1]+(self.spacing/2))]
+                extent = [np.floor(self.bbox[2]), np.ceil(self.bbox[-1]),np.floor(self.bbox[0]), np.ceil(self.bbox[1])]
             else:
-                print(
-                    "WARNING: User-specified bounds do not overlap with dataset bounds, proceeding with the latter.")
+                print("WARNING: User-specified bounds do not overlap with dataset bounds, proceeding with the latter.")
```
Askaholic, line 481: No longer accurate

Askaholic: Maybe?

sssangha: No, this is a separate issue w.r.t. my previous commit. Here, I am capturing instances where a user-specified bounding box does not intersect with the data bounds. E.g. let's say a user passed '39 40 120 121' (again south, north, west, and east respectively), but the data bounds span '41 42 122 124'. In these cases, the native data bounds are passed. 

What I just committed concerns invalid user inputs that don't correspond to legitimate coordinates. E.g. -91 south, -182 west. 

But on second though, perhaps the former case should also exit out with an exception instead of just printing a warning and proceeding with the native data bounds.

Makes sense?

Askaholic: To me it would make sense to error out if I specify a bounding box that doesn't overlap with the data at all. If I specify a bounding box that is too big, then it makes sense to me that the program would shrink it down to the available data.

sssangha: Yes I agree. Will raise an exception in the cases where the user-defined bounding box does not overlap with the data bounds.

The other exception concerning user-defined bounds that supersede valid bounds remains, as that is a separate issue.

PR #67, opened 2020-07-11
Switch to pytest

Askaholic:
Pytest has lots of really great features (for example temp directories as seen in this PR) that make writing high quality tests really easy. Probably the best feature is how it handles assert statement introspection and displays the output. For example here are some screenshots of running `py.test test` on this repo.

First, you can notice that it gathers both the `pytest` and the `unittest` tests together and shows a colored summary which is really easy to read.
image

For the test that failed it tries to show as much detail about the failure as it can. The unittest equivalent of this test would only give you `assert False` as the reason for the failure, but pytest actually prints the values of the arguments to `np.allclose` so you can actually see WHY the test failed. In case of numpy arrays, the formatting isn't great because numpy tries to format arrays for normal console output, but it at least enables you to figure out which values you're getting without having to go back and add a whole gob of print statements everywhere.
image

At the bottom you get a short summary including any warnings that were generated and a 1 line error message for each failed test.
image

There are also some amazing plugins for pytest that give you features such as code coverage, and my personal favorite, input fuzzing: https://hypothesis.readthedocs.io/en/latest/quickstart.html


dbekaert:
@Askaholic this looks good and has nice features. I like it.

- Could you also add the system calls into purest for testing; our build one which test start to end with a know example?
- I also noticed in the build log this (can you have a look if this is an error `:` seems to be rising?):
```file "/root/miniconda/envs/RAiDER/lib/python3.8/site-packages/RAiDER/getStationDelays.py", line 204
    def secondsOfDay(returnTime)
                               ^
SyntaxError: invalid syntax
```

jlmaurer:
I have new versions of test_scenario_1 and test_scenario_2 that I will wait to push until this gets merged to avoid conflicts.

Askaholic:
>     * Could you also add the system calls into Pytest for testing; our build one which test start to end with a know example?
I'm not sure what system calls you are referring to. I like to split up tests into several types: Unit tests which test small parts of the code on their own, and integration tests which run a full example end to end.

>     * I also noticed in the build log this (can you have a look if this is an error `:` seems to be rising?):
> 
> 
> ```
>     def secondsOfDay(returnTime)
>                                ^
> SyntaxError: invalid syntax
> ```

Ah yes. Sim fixed this in his PR, so I rebased and it should be good now.

dbekaert:
```
 I'm not sure what system calls you are referring to. I like to split up tests into several types: Unit tests which test small parts of the code on their own, and integration tests which run a full example end to end.
```

In our current circleCI we are having the integration test (end-end) ran separate from the Pytest.

jlmaurer:
@dbekaert not sure if this is what you're after but I have created two integration tests that run a full command and compare to known results in the PR . Was just waiting for this PR to clear so that there are no conflicts.

Issue #37, opened 2020-04-30
Fixing Zentith delay for all variants (point, grid, cube)


jlmaurer:

jlmaurer:
*Update*: Minimal working code should be able to compute zenith delays for the following inputs: 
- lat/lon file specified
- station list specified
- bounding box specified
- each of the above options with fixed height levels specified

jlmaurer:
Hey @dbekaert @Askaholic @leiyangleon I have a question. The current HEAD breaks when passing the --station_files option because of the following line: 
https://github.com/dbekaert/RAiDER/blob/eadc4627632dfcd76f2bd4f3c33de380e1cf0921/tools/RAiDER/delayFcns.py#L167

## Apparent Problem
The problem here is that the input data can be a 1-D list of stations, a 2-D raster, or a 3-D data cube (e.g. weather model nodes). Ideally we would like a single function that is agnostic of the input data shape, but right now the code assumes a 2D raster.

## Underlying problem
The real problem here is that we read the input data from an HDF5 file by chunk, and if the data dimension changes then the chunk size dimension also changes. Because of Python's GIL, we have to pass the HDF5 file pointers to each one of the subprocesses like so: 
https://github.com/dbekaert/RAiDER/blob/eadc4627632dfcd76f2bd4f3c33de380e1cf0921/tools/RAiDER/delayFcns.py#L157
where SP and SLV are the relevant HDF5 arrays. This is good because the data only gets read into memory as chunks, but bad because that means the function (unpacking_hdf5_read) doesn't know what the shape of the chunk is, and so has to do different things based on different size chunks. Currently this is not done so the code breaks. 

## Possible solutions
- One option is to leave the code as is but add in flags for various chunk sizes. I've implemented this in one of my branches and it seems to work, but the code is messy
- One option would be to flatten all the input data before writing to the HDF5 file, then perform chunking on the resulting 1D array. This isn't great if anyone needs to access the intermediate HDF5 data, but would make the delay calculation input-size agnostic
- Other ideas? 

I'm posting to this issue because currently the code doesn't pass the "minimal working example" in the comment [above](https://github.com/dbekaert/RAiDER/issues/37#issuecomment-645085885). If you guys have ideas about what is best please share!

dbekaert:
[...]

PR #69, opened 2020-07-16
Fix cmd arguments and chunking for GNSS station inputs and height levels and add unit tests

jlmaurer:
- This addresses #37
- test_scenario_1 and test_scenario_2 should complete successfully now, these are both end-to-end units tests
- **Note**: test_scenario_1 and test_scenario_2 test will test against the test data set, but the test data set has not been validated. In other words, these unit tests will test for changing values against what the code has been providing, but not the absolute accuracy. 
- chunking is still based on cpu count but adds a safety check; if the chunksize would be greater or less than a threshold it takes the threshold
- chunking is fixed for station_file (i.e. 1D) and height_level (i.e. 3D) query points


test/test_scenario_1.py
```diff
@@ -13,13 +13,16 @@
 SCENARIO_DIR = os.path.join(TEST_DIR, "scenario_1")
 
 
-@pytest.mark.skip("Test fails, needs to be fixed")
 def test_tropo_delay(tmp_path):
+    #########################################
+    # Scenario: 
+    # 1: Small area, ERA5, Zenith delay
+    #########################################
```
Askaholic, line 20: Make it a docstring instead of a comment

test/test_scenario_2.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertTrue(np.allclose(est_delay['totalDelay'].values, true_delay['totalDelay'].values, equal_nan=True))
+        self.assertTrue(np.allclose(est_delay['wetDelay'].values, true_delay['wetDelay'].values, equal_nan=True))
+        self.assertTrue(np.allclose(est_delay['hydroDelay'].values, true_delay['hydroDelay'].values, equal_nan=True))
+
+def main():
+    unittest.main()
+   
+if __name__=='__main__':
+
+    unittest.main()
```
Askaholic, line 71: No need. Pytest will discover the tests automatically

test/test_scenario_2.py
```diff
@@ -0,0 +1,72 @@
...
+    # Scenario to use: 
+    # 1: GNSS station list
+    scenario = 'scenario_2'
+    weather_model_name = 'ERA5'
+    los = Zenith
+    download_only = False
+    verbose = True
+    wetName = 'stations_with_Delays.csv'
+
+    t = datetime(2020,1,3,23,0,0)
```
Askaholic, line 25: Making these member variables is total overkill. Just pass the values in directly, using keyword arguments if you want to make it more readable. Generally providing test setup as members of a class like this is bad and leads to very messy tests. If you need to run some prior setup code, use a fixture (https://docs.pytest.org/en/stable/fixture.html), otherwise keep it inside of the test function itself.

tools/RAiDER/delayFcns.py
```diff
@@ -167,7 +155,64 @@ def make_interpolator(xs, ys, zs, data):
     )
 
 
-def process_chunk(k, chunkInds, SP, SLV, in_shape, stepSize, ifWet, ifHydro, max_len, wm_file):
+def chunk(chunkSize, in_shape):
```
Askaholic, line 158: These chunk creation functions are all great candidates for adding unit tests. This will also help document what the function actually does since you will have something like:

```python3
assert chunk(10, (100, )) == [10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
```
or whatever it actually does. I have no idea what the output is supposed to be from just reading the docstring.

tools/RAiDER/delayFcns.py
```diff
@@ -167,7 +155,64 @@ def make_interpolator(xs, ys, zs, data):
...
+    '''
+    Create a set of indices to use as chunks
+    '''
+    startInds =  makeChunkStartInds(chunkSize, in_shape)
+    chunkInds = makeChunksFromInds(startInds, chunkSize, in_shape)
+    return chunkInds
+
+
+def makeChunksFromInds(startInd, chunkSize, in_shape):
```
Askaholic, line 167: Since `makeChunkStartInds` is called first above, I would expect it to be defined first as well. This helps with reading the code, because the way it is now, I have to skip down and back up instead of just being able to read straight down.

tools/RAiDER/delayFcns.py
```diff
@@ -167,7 +155,64 @@ def make_interpolator(xs, ys, zs, data):
...
+            chunks.append([np.array(g) for g in zip(*list(itertools.product(*index)))])
+    else:
+        chunks = indices
+
+    return chunks 
+
+
+def makeChunkStartInds(chunkSize, in_shape):
+    '''
+    Create a list of indices for chunking a 2D array
```
Askaholic, line 197: If it only works for 2d arrays then that should probably be reflected in the function name. Sometimes I also like to assert my preconditions at the start of a function to make sure no programming errors occur.

```python3
assert len(in_shape) == 2
```

tools/RAiDER/delayFcns.py
```diff
@@ -167,7 +155,64 @@ def make_interpolator(xs, ys, zs, data):
...
+    '''
+    if len(in_shape) == 1:
+        chunkInds = [(i,) for i in range(0, in_shape[0], chunkSize[0])]
+
+    elif len(in_shape) == 2:
+        chunkInds = [(i, j) for i, j in itertools.product(range(0, in_shape[0], chunkSize[0]), 
+                                                          range(0, in_shape[1], chunkSize[1]))]
+    elif len(in_shape) == 3:
```
Askaholic, line 205: Ok, so I guess the docstring is wrong.

tools/RAiDER/runProgram.py
```diff
@@ -144,6 +143,7 @@ def parseCMD():
     times, out, download_only, verbose, \
     wetNames, hydroNames = checkArgs(args, p)
 
+    import pdb; pdb.set_trace()
```
Askaholic, line 146:
```suggestion

```

tools/bindings/utils/makePoints.pyx
```diff
@@ -6,11 +6,13 @@ DTYPE = np.float64
...
 
 ctypedef double npy_float64
+ctypedef float npy_float32
 ctypedef npy_float64  float64_t
+ctypedef npy_float32  float32_t
 
 @cython.boundscheck(False)  # turn off array bounds check
 @cython.wraparound(False)   # turn off negative indices ([-1,-1])
-def makePoints1D(double max_len, cnp.ndarray[double, ndim=1] Rays_SP, cnp.ndarray[double, ndim=1] Rays_SLV, double stepSize):
+def makePoints0D(double max_len, cnp.ndarray[double, ndim=1] Rays_SP, cnp.ndarray[double, ndim=1] Rays_SLV, double stepSize):
```
Askaholic, line 15: Add unit tests? I think the other makePoints functions have some tests.

jlmaurer:
Hey @Askaholic I'm having trouble with test_scenario_1. For some reason I'm getting NaNs when I run the unit test, but the same (as far as I can tell) test run in command-line form doesn't return NaNs. For comparison to test_scenario_1, I run the following from within the scenario_1 directory:
```
raiderDelay.py --date 20200103 --time 23:00:00 --latlon geom/lat.dat geom/lon.dat --model ERA5 --zref 20000. --outformat envi --out . -w weather_files
```
The problem evidently happens after the weather model is created because the weather model HDF5 file already exists and is used by both. Running py.test test results in a delay file containing 0's (the default noData value), while running the command line version results in a file without NaNs. 

Any ideas?

test/test_scenario_1.py
```diff
@@ -12,21 +12,23 @@
...
-        SCENARIO_DIR, 'geom', 'ERA5_Lon_2018_01_01_T00_00_00.dat'
+        SCENARIO_DIR, 'geom', 'lon.dat'
     ))
 
     time = datetime.datetime(2020, 1, 3, 23, 0)
 
     _, model_obj = modelName2Module("ERA5")
     wet_file, hydro_file = makeDelayFileNames(
-        time, Zenith, "envi", "ERA5", tmp_path
+        time, Zenith, "envi", "ERA5", SCENARIO_DIR
```
Askaholic, line 31: Test should be isolated to a temp directory to avoid interactions with previous runs.

test/test_util.py
```diff
@@ -102,26 +154,15 @@ def test_writeArrayToRaster(tmp_path):
     assert np.allclose(band.ReadAsArray(), array)
 
 
-def test_makePoints1D_Python(make_points_1d_data):
-    from RAiDER.utilFcns import makePoints1D
+def test_makePoints0D_cython(make_points_0d_data):
+    from RAiDER.makePoints import makePoints0D
```
Askaholic, line 158: We should put the imports at the top of the file. I guess I forgot to do this when I was refactoring before

Askaholic: I would use `as` to differentiate between the cython and python versions

```python
from RAiDER.makePoints import makePoints0D as make_points_0d_cython
```

tools/RAiDER/delayFcns.py
```diff
@@ -167,7 +155,94 @@ def make_interpolator(xs, ys, zs, data):
...
+    startInds =  makeChunkStartInds(chunkSize, in_shape)
+    chunkInds = makeChunksFromInds(startInds, chunkSize, in_shape)
+    return chunkInds
+
+
+def makeChunkStartInds(chunkSize, in_shape):
+    '''
+    Create a list of start indices for chunking a numpy D-dimensional array. 
+    Inputs: 
+        chunkSize - length-D tuple containing chunk sizes
```
Askaholic, line 171: Maybe it should be plural: `chunk_sizes`

test/test_chunking.py
```diff
@@ -0,0 +1,102 @@
...
+        for k2,arr in enumerate(L):
+            if np.allclose(arr, chunk1[k1][k2]):
+                check.append(True)
+            else:
+                check.append(False)
+    return all(check)
+
+def test_makeChunkStartInds1D():
+    chunkStartInds = [(0,), (1,), (2,), (3,), (4,)]
+    assert chunkStartInds == makeChunkStartInds((1,),(5,))
```
Askaholic, line 21: No need for intermediate variable.

I'm not sure whether it's best to do `assert desired == function()` or `assert function() == desired`, but it's definitely important to be consistent, otherwise reading test output can get confusing. I have adopted the convention of always putting the desired output on the right, mostly because it is more intuitive when you are writing the test. 

Since, the other tests I've written already follow this convention, I'd say let's follow it here too. If you have a really good reason for doing it the other way, then you would need to change it for all of the tests in the project. Consistency is key.

```suggestion
    assert makeChunkStartInds((1,),(5,)) == [(0,), (1,), (2,), (3,), (4,)]
```

test/test_chunking.py
```diff
@@ -0,0 +1,102 @@
...
+    in_shape = (4,)
+    chunkSize = (2,)
+    ind = [(0,), (2,)]
+    chunks = [[np.array([0, 1])], [np.array([2, 3])]]
+    assert checkChunks(chunks, makeChunksFromInds(ind,chunkSize,in_shape)) 
+
+def test_makeChunksFromInds2D():
+    in_shape = (4,4)
+    chunkSize = (2,2)
+    ind = [(0, 0), (0, 2), (2, 0), (2, 2)]
```
Askaholic, line 50: No need for all these variables, they just clutter up the code. If you want to make the function call more readable, use keyword arguments.

test/test_scenario_2.py
```diff
@@ -0,0 +1,67 @@
...
+           wmLoc=None, 
+           zref=20000., 
+           outformat='csv', 
+           time=datetime(2020,1,3,23,0,0), 
+           out=tmp_path, 
+           download_only=False, 
+           verbose=True, 
+           wetFilename=wetFile, 
+           hydroFilename=hydroFile
+           )
```
Askaholic, line 57: dedent
```suggestion
       )
```

test/test_chunking.py
```diff
@@ -0,0 +1,102 @@
...
+    chunkStartInds = (
+        [(0, 0, 0),
+	 (0, 2, 0),
+	 (0, 4, 0),
+	 (0, 6, 0),
+	 (2, 0, 0),
+	 (2, 2, 0),
+	 (2, 4, 0),
+	 (2, 6, 0)] 
+	)
```
Askaholic, line 37: Yikes, formatting

test/test_scenario_2.py
```diff
@@ -0,0 +1,67 @@
+# Unit and other tests
```
Askaholic, line 1: No need for this comment. We know it's a test file because it's in the test directory and the name starts with test.

test/test_util.py
jlmaurer: @Askaholic I completely removed the python versions of these codes as I don't think they're needed, they were only used when the cython versions weren't working.

jlmaurer:
I'm not sure whether circleCI updates the unit tests themselves or tries to run the old tests, but all unit tests are passing or skipped on my repo.

Askaholic:
It's failing when it tries to read a file that doesn't exist. Chances are you still have an old version of this file laying around on your filesystem and that's why it's passing for you and not when circleci clones a fresh copy.

jlmaurer:
@Askaholic yes, that is a good point; however, in this case it seems to be the opposite situation. The file it's trying to find is the old name of the geometry file, which suggests it is running the old version of the unit tests, as that file gets named explicitly in the test_scenario_1 script. That file doesn't exist on my system.

jlmaurer:
There was an error where it was trying to write to a directory that didn't exist, which I fixed with d49961bf7286ca73913e6b9e27f4d918ea25d68a.

Askaholic:
It's not failing on the unit tests themselves, it's failing on the setup commands that are run before hand. Specifically I think it's the last one:
```
raiderDelay.py --date 20200103 --time 23:00:00 --latlon test/scenario_1/geom/ERA5_Lat_2018_01_01_T00_00_00.dat test/scenario_1/geom/ERA5_Lon_2018_01_01_T00_00_00.dat --model ERA5 --zref 20000 -v --out test/scenario_1/
```
Notice that the 2018 files are still specified on the command line

jlmaurer:
Ah @Askaholic so that is the issue! Thanks for pointing that out; I was trying to follow the problematic command on the circleCI website but was having trouble. Anyway I'll see if I can update the config file to use the renamed versions.

Askaholic:
Yea it's a little hard to tell what's going on in circleci because it prints all the commands and the top, and then shows the output for the entire sequence of commands beneath it.

Askaholic:
BTW that command is also present in the help string of the argument parser created in `parse_args`.

jlmaurer:
@Askaholic @dbekaert is there a way to make changes to the config file for circleci as a part of the PR? The file doesn't get cloned with the repository. It would be ideal to tie the change with the PR.

dbekaert:
Not sure what you mean. Yes you can edit the .circleci/files in the repo.
Those when edited and pr issues will be the file that gets checked out.

Askaholic:
The file is in a hidden folder: `.circleci/config.yml`. I edited it when I added pytest

dbekaert:
@jlmaurer looks like there is no lon lat files in the test/scenario_1 folder.

tools/RAiDER/utilFcns.py
```diff
@@ -153,7 +149,7 @@ def big_and(*args):
     return result
 
 
-def gdal_open(fname, returnProj=False):
+def gdal_open(fname, returnProj=False, userNDV = None):
```
Askaholic, line 152:
```suggestion
def gdal_open(fname, returnProj=False, userNDV=None):
```

tools/RAiDER/utilFcns.py
```diff
@@ -166,14 +162,17 @@ def gdal_open(fname, returnProj=False):
...
-            ndv = b.GetNoDataValue()
-            d[d == ndv] = np.nan
-        except:
-            print('NoDataValue attempt failed*******')
-            pass
-        val.append(d)
+        data = b.ReadAsArray()
+        if userNDV is not None:
+            print('Using user-supplied NoDataValue')
+            data[data==userNDV]=np.nan
```
Askaholic, line 168:
```suggestion
            data[data == userNDV] = np.nan
```

tools/bindings/utils/makePoints.pyx
```diff
@@ -38,6 +40,40 @@ def makePoints1D(double max_len, cnp.ndarray[double, ndim=1] Rays_SP, cnp.ndarra
...
+        Npts = int(max_len//stepSize)
+
+    cdef int nrow = Rays_SP.shape[0]
+    cdef cnp.ndarray[npy_float64, ndim=3, mode='c'] ray = np.empty((nrow, 3, Npts), dtype=np.float64)
+    cdef cnp.ndarray[npy_float64, ndim=1, mode='c'] basespace = np.arange(0, max_len+stepSize, stepSize) 
+
+    for k1 in range(nrow):
+        for k3 in range(3):
+            for k4 in range(Npts):
+                ray[k1,k3,k4] = Rays_SP[k1,k3] + basespace[k4]*Rays_SLV[k1,k3]
```
Askaholic, line 73:
```suggestion
                ray[k1, k3, k4] = Rays_SP[k1, k3] + basespace[k4] * Rays_SLV[k1, k3]
```

test/test_chunking.py
```diff
@@ -0,0 +1,111 @@
+import numpy as np
+import pytest
+
+from RAiDER.delayFcns import chunk, makeChunksFromInds, makeChunkStartInds
+
+# The purpose of these tests is to verify the chunking done on the query points during is
+# done correctly for various dimensions
+
+def checkChunks(chunk1, chunk2):
+    check = []
```
Askaholic, line 10: Unused variable

test/test_scenario_1.py
```diff
@@ -12,21 +11,27 @@
...
     time = datetime.datetime(2020, 1, 3, 23, 0)
 
+    wmLoc = os.path.join(SCENARIO_DIR, 'weather_files')
+    if not os.path.exists(wmLoc):
+        os.mkdir(wmLoc)
+
     _, model_obj = modelName2Module("ERA5")
     wet_file, hydro_file = makeDelayFileNames(
-        time, Zenith, "envi", "ERA5", tmp_path
+        time, Zenith, "envi", "ERA5", tmp_path
```
Askaholic, line 34:
```suggestion
        time, Zenith, "envi", "ERA5", tmp_path
```

jlmaurer:
Closing this PR and pushing from a new branch.

PR #70, opened 2020-07-16
Speed up interpolation by implementing it in C++

Askaholic:
After several weeks of fighting against outdated documentation, poor unit testing frameworks and of course constant segfaults I finally got the C++ interpolation into a pretty stable and usable state. I was also reminded of all the reasons why C++ sucks and everyone should just switch to Rust already :)!

If you recall from my other PR, I was able to load the HRRR model and compute zenith delays at the weather model nodes in about 285 seconds using the command:
```
raiderDelay.py --date 20180101 --time 23:00:00 -b 40 -110 25 -95 --model HRRR --zref 30000 -v
```
The most expensive call was `interp_along_axis` which was multithreaded in Python so it showed up as the "acquire" method of lock objects:
```
         11804562 function calls (10569262 primitive calls) in 285.811 seconds

   Ordered by: internal time
   List reduced from 4431 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       77  182.879    2.375  182.879    2.375 {method 'acquire' of '_thread.lock' objects}
        3   14.221    4.740   24.008    8.003 dataset.py:326(__getitem__)
```

Since I couldn't find a way to achieve this 1D interpolation along an axis in a single call using existing numpy/scipy interpolators, and I had already implemented 1D interpolation in C++ at that point, I decided to rewrite `interp_along_axis` entirely in C++, to get around the overhead of calling python functions. I also went ahead and implemented the multithreading in C++ as well using std::async so we don't need to worry about managing that on the Python side anymore. After replacing the calls to `interp_along_axis` inside of `_uniform_in_z`, I ended up with the following profile:

```
         11801983 function calls (10566645 primitive calls) in 96.529 seconds

   Ordered by: internal time
   List reduced from 4264 to 15 due to restriction <15>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        3   13.949    4.650   24.460    8.153 dataset.py:326(__getitem__)
      324    9.199    0.028    9.199    0.028 bindings.py:158(wrapper)
      148    7.954    0.054   17.377    0.117 bindings.py:277(codes_get_double_array)
        1    5.671    5.671    5.671    5.671 {built-in method _pickle.dump}
        3    4.670    1.557    4.670    1.557 utilFcns.py:31(cosd)
1219699/1219662    4.170    0.000   28.692    0.000 {built-in method numpy.array}
       82    3.771    0.046    4.289    0.052 function_base.py:4006(trapz)
        1    3.620    3.620    3.620    3.620 weatherModel.py:419(_find_svp)
1219381/571    3.613    0.000   24.255    0.042 {built-in method numpy.core._multiarray_umath.implement_array_function}
        2    3.523    1.761    3.523    1.761 {method 'sort' of 'numpy.ndarray' objects}
   609309    2.898    0.000    2.898    0.000 {built-in method numpy.core._multiarray_umath.interp}
       13    2.696    0.207    2.795    0.215 dataset.py:57(make_new_dset)
        3    2.247    0.749    2.247    0.749 {built-in method RAiDER.interpolate.interpolate_along_axis}
        2    1.879    0.939    9.191    4.596 shape_base.py:267(apply_along_axis)
        8    1.680    0.210    1.680    0.210 {method 'flatten' of 'numpy.ndarray' objects}
```

The interpolation which took 182 seconds before, now takes only 2.2 seconds. As far as I can tell, the results are exactly the same, although I want to write some unit tests to verify this (specifically to verify the "output" of _uniform_in_z).

Now the majority of the time is spent loading the data into memory. This might be something that we can't even speed up very easily, and is probably pretty low hanging fruit anyways.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ pybind_extensions = [
             # Path to pybind11 headers
             get_pybind_include(),
         ],
+        extra_compile_args=['-std=c++11'],
         language='c++'
     ),
 ]
--- a/setup.py
+++ b/setup.py
@@ -6,14 +6,14 @@
 # RESERVED. United States Government Sponsorship acknowledged.
 #
 #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-import numpy as np
-import os
 import glob
-import subprocess as subp
+import os
+import re
 
-from distutils.core import setup
-from distutils.extension import Extension
-from Cython.Build import cythonize
+import numpy as np
+from setuptools import Extension, setup
+
+from Cython.Build import cythonize  # isort:skip
 
 # Parameter defs
 CWD = os.getcwd()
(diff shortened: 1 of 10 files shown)


dbekaert:
@Askaholic this looks very promising! This optimization is definitely paying off.

Issue #36, opened 2020-04-30
Add GMAO model support


jlmaurer:

dbekaert:
see item #35 as there is a lot of similarity between what is needed for MERRA-2.
e.g. likely leverage same OpenDAP mechanism (different base URL) and same loading mechanism (netcdf4) for the variables. In terms of the model levels see below PDF for information

Relevant URLs:
- background information: https://gmao.gsfc.nasa.gov/GMAO_products/documents/GEOS_5_FP_File_Specification_ON4v1_2.pdf
- website for data-access (see left openDAP data): https://fluid.nccs.nasa.gov/weather/
 
Data to be used:
- **preferred:** inst3_3d_asm_Nv: 3Hour outputs are 0.3° lateral sampling and 72 vertical model levels. The information on the model levels is in the above PDF
- inst3_3d_asm_Np: 3Hour outputs at 0.3° Lateral sampling and 42 pressure levels

leiyangleon:
@dbekaert Using the preferred model level [URL](https://opendap.nccs.nasa.gov/dods/GEOS-5/fp/0.25_deg/assim/inst3_3d_asm_Nv
), I am able to get the dataset returned but not able to retrieve the specific data values. The line I used is
`ds['t'].array[10:14,10:14,10:14,10:14].data`
But the error says **_IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices_**
I tried the example on the Pydap Python [Wikipage](https://www.pydap.org/en/latest/client.html) and it worked fine.


Also, as mentioned [here](https://opendap.nccs.nasa.gov/dods/GEOS-5/fp/0.25_deg/assim/inst3_3d_asm_Nv.info), the temporal resolution is 3 hrs. But you wanted to use 7.5 hrs for the analysis data. So does it require a different URL? Or we can code that 7.5 hrs in the reader and still use the same URL?

dbekaert:
@leiyangleon the model sampling is indeed every 3 hours. The 7.5hr latency means a user cannot request any data within 7.5 hour from real-time as it does not yet exist. We should capture that an error out to the user for such requests.

Can you provide more information on what you are doing in terms of getting the data using OpenDAP, for recreation purposes? I assume you might already slicing through the data? E.g. What happens if you ask for the complete array of 't' instead?

leiyangleon:
[...]

PR #71, opened 2020-07-20
added GMAO model support and updated documentation

leiyangleon:


leiyangleon:
@dbekaert @jlmaurer For the GMAO model, I removed the hashed out lines as suggested. For the documentation, I replaced the actual API key with a symbol.

environment.yml
```diff
@@ -28,4 +28,4 @@ dependencies:
  - scipy
  - xarray
  - shapely
- - pydap
+ - pydap=3.2.1
```
dbekaert, line 31: @leiyangleon do we want to force a single version or could we do it such we have a version excluded? I recall it was an intermediate version which is problematic?

https://stackoverflow.com/questions/54847688/how-to-specify-version-ranges-in-conda-environment-yml

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        # assign the regular-grid (lat/lon/h) variables
+        
+        self._p = p_intrpl
+        self._q = q_intrpl
+        self._t = t_intrpl
+        self._lats = _lats
+        self._lons = _lons
+#        self._xs = lons
+#        self._ys = lats
+#        self._zs = np.flip(hs)
```
dbekaert, line 166: if not used would remove it.

README.md
```diff
@@ -50,6 +51,7 @@ Below we list the dependencies for RAiDER
 * numpy
 * pandas
 * xarray
+* pydap 3.2.1
```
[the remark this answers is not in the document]
dbekaert: @jlmaurer i recall there is a version with a bug that @leiyangleon said not to use.

dbekaert:
> @leiyangleon I suggest to run the following from the base RAiDER repo and then commit the updates:
> 
> ```
> autopep8 tools/RAiDER/ --recursive --select E231,E225,E251,E111,E128,E127,E265,E303,E261,E301 --in-place
> ```
> 
> To fix formatting/whitespace issues.

@jlmaurer could you add a template for PR with some guidance like this?
see under setting then scroll down to templates.

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        self._Name = 'GMAO'
+        self._files = None
+        self._bounds = None
+
+        # Projection
+        self._proj = CRS.from_epsg(4326)
+
+    
+    
+
```
Askaholic, line 50: Only 1 blank line between methods of a class

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        self._humidityType = 'q'
+        self._model_level_type = 'ml' # Default, pressure levels are 'pl'
+        
+        self._classname = 'gmao'
+        self._dataset = 'gmao'
+
+        # Tuple of min/max years where data is available. 
+        self._valid_range = (datetime.datetime(2017,12,1),"Present")
+        self._lag_time = datetime.timedelta(hours=0.3125) # Availability lag time in days
```
Askaholic, line 28: Comment says "lag time in days" but "hours" keyword argument is used. Is this correct?

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        '''
+        Get the variables from the GMAO link using OpenDAP
+        '''
+        # calculate the array indices for slicing the GMAO variable arrays
+        lat_min_ind = int((self._bounds[0] - (-90.0)) / self._lat_res)
+        lat_max_ind = int((self._bounds[1] - (-90.0)) / self._lat_res)
+        lon_min_ind = int((self._bounds[2] - (-180.0)) / self._lon_res)
+        lon_max_ind = int((self._bounds[3] - (-180.0)) / self._lon_res)
+        
+        import datetime as dt
```
Askaholic, line 84: It's already imported at the top of the file. IMO you should just do:
```python3
from datetime import datetime
```
once at the top of the file.

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        T0 = dt.datetime(2017,12,1,0,0,0)
+        DT = self._time - T0
+        time_ind = int(DT.total_seconds() / 3600.0 / 3.0)
+        
+        ml_min = 0
+        ml_max = 71
+
+        # open the dataset and pull the data
+        import pydap.client
+        import pydap.cas.urs
```
Askaholic, line 94: I can see some argument for lazy imports when it's an external dependency. At least put it at the top of the function though.

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        
+        q = ds['qv'].array[time_ind,ml_min:(ml_max+1),lat_min_ind:(lat_max_ind+1),lon_min_ind:(lon_max_ind+1)][0]
+        p = ds['pl'].array[time_ind,ml_min:(ml_max+1),lat_min_ind:(lat_max_ind+1),lon_min_ind:(lon_max_ind+1)][0]
+        t = ds['t'].array[time_ind,ml_min:(ml_max+1),lat_min_ind:(lat_max_ind+1),lon_min_ind:(lon_max_ind+1)][0]
+        h = ds['h'].array[time_ind,ml_min:(ml_max+1),lat_min_ind:(lat_max_ind+1),lon_min_ind:(lon_max_ind+1)][0]
+        
+        # calculate the lat, lon and mean h for each layer in the regular grid
+        hs = np.zeros(72)
+        for ii in range(72):
+            hs[ii] = np.mean(h[ii,:,:])
```
Askaholic, line 107: I think you can use the `axis` parameter of [np.mean](https://numpy.org/doc/stable/reference/generated/numpy.mean.html). This will be much faster, less code and doesn't have a hardcoded constant.

```suggestion
        hs = np.mean(h, axis=(1, 2))
```

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        q = np.flip(q, axis = 2)
+        t = np.flip(t, axis = 2)
+        h = np.flip(h, axis = 2)
+        _lats = np.flip(_lats, axis = 2)
+        _lons = np.flip(_lons, axis = 2)
+        _hs = np.flip(_hs, axis = 2)
+        
+
+        # interpolate (p,q,t) along the vertical axis from irregular grid to regular grid
+        from RAiDER.interpolator import interp_along_axis
```
Askaholic, line 149: Import at top of file

tools/RAiDER/models/gmao.py
```diff
@@ -0,0 +1,171 @@
...
+        p_intrpl = interp_along_axis(h, _hs, p, axis = 2)
+        q_intrpl = interp_along_axis(h, _hs, q, axis = 2)
+        t_intrpl = interp_along_axis(h, _hs, t, axis = 2)
```
Askaholic, line 152: This stuff looks very similar to `_uniform_in_z`. Is there some way we can refactor it, instead of having duplicate implementations?

environment.yml
leiyangleon: @dbekaert @jlmaurer fixed this by excluding a version as @dbekaert suggested above.

weather_setup.md
```diff
@@ -0,0 +1,48 @@
...
+                 login <USERNAME>
+                 password <PASSWORD>
+                 
+   _Note the `<USERNAME>` and `<PASSWORD>` represent the actual username and password, and should be replaced with the user's own information correspondingly._
+   
+4. Add the application `NASA GESDISC DATA ARCHIVE` by clicking on the `Applications->Authorized Apps` on the menu after logging into your Earthdata profile, and then scrolling down to the application `NASA GESDISC DATA ARCHIVE` to approve it. _This seems not required for GMAO for now, but recommended to do so for all OpenDAP-based weather models._
+
+5. Install the OpenDAP using pip: 
+
+         pip install pydap
```
Askaholic, line 44:
```suggestion
         pip install pydap==3.2.1
```

tools/RAiDER/models/gmao.py
leiyangleon: @dbekaert all right. Removed them for now. Although I have been waiting for @jlmaurer 's response on why these must be 3-D cubes, which contradicts to what Jeremy told me, i.e. they should be 1-D vectors.

README.md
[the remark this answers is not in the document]
leiyangleon: @dbekaert @jlmaurer fixed already as mentioned above. 

However, for the following command, it did not work for me and the error message says command not found for autopep8. @jlmaurer could you let me know how to use it?


> @leiyangleon I suggest to run the following from the base RAiDER repo and then commit the updates:
> 
> ```
> autopep8 tools/RAiDER/ --recursive --select E231,E225,E251,E111,E128,E127,E265,E303,E261,E301 --in-place
> ```
> 
> To fix formatting/whitespace issues.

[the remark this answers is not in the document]
Askaholic: You'll need to install [autopep8](https://pypi.org/project/autopep8/) with pip:

```
pip install --user autopep8
```

leiyangleon:
@dbekaert @jlmaurer all of the suggested changes by @Askaholic have been implemented except the one on "_uniform_z", which also calls the interp_along_axis function. However, there are some specific use in "_uniform_z", such as

- self._e is used instead of self._p in all weather model readers

- the output of self._uniform_z will enforce self._xs, self._ys and self._zs to be 1-D vectors instead of 3-D cubes, which will cause the program to fail for all weather model readers in the current stage. This has been reported to @jlmaurer ([here](https://github.com/dbekaert/RAiDER/issues/36#issuecomment-658452354))however without reply yet.

leiyangleon:
@dbekaert @jlmaurer @Askaholic I installed `autopep8` and ran the [command](https://github.com/dbekaert/RAiDER/pull/71#pullrequestreview-451953873) as suggested by @jlmaurer without any messages printed out. So I assume it is all set.

leiyangleon:
@dbekaert @jlmaurer the change on pydap version in `environment.yml` has been excluded in this PR, otherwise it keeps issuing conflicts to be resolved. So once this PR is merged, I will issue another one just to update pydap version.

Askaholic:
Did it modify the files? You'll need to commit the changes, as the version currently pushed to github does not have the fixed formatting.

You'll also need to merge in the latest changes to the dev branch. To do this you'll first need to add this repo as a remote:
```
git remote add upstream https://github.com/dbekaert/RAiDER.git
```
and then run `git pull upstream dev`. This will pause on the conflicting file asking you to resolve the conflicts. Once they are resolved you can run `git add environment.yml` and `git commit` (just use the default message). Then you can push the merged changes to your fork.

tools/RAiDER/models/gmao.py
Askaholic: So is the lag time is really supposed to be 18.75 minutes? Then I would use the minutes argument so it's more clear. If it's supposed to be 0.3125 days (7.5 hours) then I would use `hours=7.5`.

tools/RAiDER/models/gmao.py
Askaholic: You also need to update the places where you're using it.

tools/RAiDER/models/gmao.py
Askaholic: Like I said, I think there is some merit to importing these lazily since it's an external dependency that some people might not need. Just do it at the top of the function (not the middle of the function):

```python
   def _load_model_level(self,f):
        '''
        Get the variables from the GMAO link using OpenDAP
        '''
        import pydap.client
        import pydap.cas.urs

        # calculate the array indices for slicing the GMAO variable arrays
        ...
```

leiyangleon:
@dbekaert @jlmaurer I just worked with @Askaholic and went through the above fixes. It is ready to be merged if you don't have further comments.

tools/RAiDER/models/gmao.py
dbekaert: @Askaholic @jlmaurer i assume this should be recurrent in other models too for model levels?

Issue #66, opened 2020-07-10
Speed-up of downloadgnssdelays.py and statsPlot.py


sssangha:
**Proposed Enhancement**
Accessing delays from UNR and storing it in a CSV file through ``` downloadgnssdelays.py```,  and thereafter loading this file and prepping it for plotting through ``` statsPlot.py``` is still rather slow and inefficient, especially over a global scale. The latter in my experience can take over a day, while the former can take ~20 mins. 

Specifically, I wish to explore ways to parallelize certain functions where appropriate (e.g. appending delays to a CSV file and splitting data into grid-cells of a given spacing) or perhaps developing more efficient alternatives to existing routines (e.g. plotting maps of delays directly from a pandas dataframe instead of passing a gridded array meant to serve as a pseudo-raster).

**Suggested baseline tests**
Instead of running global/decadal tests to start out, I suggest running jobs over a more moderately scale spanning California and 2 years. 
Specifically as so:
###Generate CSV of delays across all stations spanning California between 2016 and 2018 at 06:00 UTC
```raiderDownloadGNSS.py --out CA_test -y '2016,2018' --returntime '00:06:00' -b '31 42 -125 -113'```

###Generate figures of the station distribution and delay heatmaps
```raiderStats.py -f CA_test/CombinedGPS_ztd.csv -w CA_test/maps -ti '2016-01-01 2018-12-31' -grid_heatmap -station_distribution -grid_delay_mean -grid_delay_stdev -b '31 42 -125 -113'```

PS, sorry for posting a blank ticket earlier. My network went down for a while and it was somehow submitted prematurely.

dbekaert:
@Askaholic Could you provide some thoughts on this enhancement; perhaps identify current bottlenecks that we also should address in the code, caveats of over hammering a website, and perhaps basic profiling on the download and CSV generation?

Askaholic:
Ok, I can take a look at this today.

Askaholic:
@sssangha and I took a look at this together yesterday. We used `cProfile` to figure out that almost all of the time is spent downloading the data with `requests.get`, which is not really surprising. Since the data hosted by UNR is only available through a bog standard Apache2 HTTP file server, we are stuck doing 1 web request per file. So the only real option for speeding it up is doing multiple web requests at a time. Since it required minimal code changes, we did this with a multiprocessing pool, but there would be more efficient ways to do this using asyncio, depending on how much more time we want to spend on that aspect.

[...]

PR #72, opened 2020-07-21
Implemented multiprocessing

sssangha:
Following discussion in Issue #66  to isolate time-sinks, multiprocessing implemented for the following functions:

1.    Checking for the existence of station zenith delay data across a specified range of years. Done in script ```downloadGNSDelays.py``` and wrapped around function ```download_UNR```. Before multiprocessing was leveraged, the run-time was ~40 secs. With 4-48 CPUs it takes ~5 secs. No tangible difference above 4 CPUs

2.    Accessing station zenith delay data and appending to master CSV file. Done in script ```getStationDelays.py``` and wrapped around function ```get_delays```. Before multiprocessing was leveraged, the run-time was ~432 secs. With 4 CPUs it went down to ~365 secs, 6 CPUs down to ~327 secs, 8 CPUs down to ~200 secs, 10 CPUs down to 193, and at 48 it took only ~55 secs. 

3.    Constraining the intersection of a specific grid-cell (from a gridded array spanning a region of interest) to each station location. Done in script ```statsPloy.py``` and wrapped around function ```_check_stationgrid_intersection``` in the ```RaiderStats``` class. In addition to optimizing this procedure with the ```multiprocessing``` module, indexing is now executed more efficiently using the ```STRtree``` module from ```shapely```. Before indexing was done crudely by checking for shape intersections over the entire array of polygons each iteration using list comprehension. Before multiprocessing was leveraged, the run-time was ~2841 secs. With 8 CPUs it is ~374. I still need to conduct more thorough tests over a larger range like above. 

By default the number of cpus used is set to 8, but the user may manually specify the number of cores they wish to use with the ```—cpus``` flag. They may specify ```all``` to use all available cores. It is interesting that—at least from what I gathered in my tests—there doesn’t appear to be a saturation in the network bandwidth when using all available 48 cores. The speed was improved, not compromised as I expected. But again, this is to be done at the user’s discretion and the default of 8 core remains for now as a conservative default.

To replicate the tests for points 1-2 above, run ```raiderDownloadGNSS.py --out CA_test -y '2017,2018' --returntime '00:06:00' -b '35 40 -120 -115' --cpus 10```

To replicate the test for points 3 above run the following commands in succession 
```
raiderDownloadGNSS.py --out world_test -y '2019' --returntime '00:00:00' -b '-90 90 -180 180' --cpus 10
raiderStats.py -f world_test/CombinedGPS_ztd.csv -w 2019 -ti '2019-01-01 2019-12-31' -grid_delay_stdev -sp 1 -b '-90 90 -180 180' -cb '150 880' --cpus 10
```


tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -80,6 +81,11 @@ def create_parser():
         help='Physically download data. Note this option is not necessary to proceed with statistical analyses, as data can be handled virtually in the program.',
         action='store_true', dest='download', default=False)
 
+    misc.add_argument(
+        '--cpus',
+        help='Specify number of cpus to be used for multiprocessing. May specify "all" at your own discretion.',
+        dest='numCPUs', type=parse_cpus, default=4)
+
```
dbekaert, line 88: @Askaholic how are number of cpu's controlled on the RAiDER size?

tools/RAiDER/statsPlot.py
```diff
@@ -807,11 +827,13 @@ def stats_analyses(inps=None):
 
     # Station plots
     # Plot each individual station
+    start_time = time.time()
     if inps.station_distribution:
         print("- Plot spatial distribution of stations.")
         unique_points = df_stats.df.groupby(['Lon', 'Lat']).size()
         df_stats([unique_points.index.get_level_values('Lon').tolist(), unique_points.index.get_level_values('Lat').tolist(
         )], 'station_distribution', workdir=os.path.join(inps.workdir, 'figures'), plotFormat=inps.plot_fmt)
+    print("station_distribution took", time.time() - start_time)
```
jlmaurer, line 836: I would put time statements under the verbose option, so they are only run if the user requests verbose output.

tools/RAiDER/statsPlot.py
```diff
@@ -831,12 +853,14 @@ def stats_analyses(inps=None):
...
     # Gridded station plots
     # Plot density of stations for each gridcell
+    start_time = time.time()
     if inps.grid_heatmap:
         print("- Plot density of stations per gridcell.")
         gridarr_heatmap = np.array([np.nan if i[0] not in df_stats.df['gridnode'].values[:] else float(len(np.unique(
             df_stats.df['ID'][df_stats.df['gridnode'] == i[0]]))) for i in enumerate(df_stats.gridpoints)]).reshape(df_stats.grid_dim)
         df_stats(gridarr_heatmap.T, 'grid_heatmap', workdir=os.path.join(inps.workdir, 'figures'), drawgridlines=inps.drawgridlines,
                  colorbarfmt='%1i', stationsongrids=inps.stationsongrids, plotFormat=inps.plot_fmt)
+    print("grid_heatmap took", time.time() - start_time)
```
jlmaurer, line 863: Same thing here as above on time statements.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -271,6 +283,22 @@ def parse_years(timestr):
...
+def parse_cpus(cpustr):
+    '''
+    Takes string input and returns integer number of cpus for multiprocessing
+    '''
+    cpus = cpustr.lower()
+    # If user specifies 'all', pass maximum number of CPUs allowed by system. Otherwise, pass specified integer.
+    if cpus == 'all':
+        cpus = os.cpu_count()
+    elif cpus.isdigit():
+        cpus = int(cpus)
```
Askaholic, line 295: Maybe some sanity checks like the number is positive. Don't know if you would want to put an upper bound on it.

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -166,18 +172,22 @@ def download_tropo_delays(stats, years, writeDir='.', download=False, verbose=Fa
...
-        fileurl = download_UNR(s, y, writeDir=writeDir,
-                               download=download, verbose=verbose)
-        # only record valid path
-        if fileurl:
-            results.append({'ID': s, 'year': y, 'path': fileurl})
-    return pd.DataFrame(results).set_index('ID')
+    stat_year_tup = (tup + (writeDir,) + (download,) + (verbose,) for tup in stat_year_tup)
+    # Parallelize remote querying of station locations
+    with multiprocessing.Pool(numCPUs) as multipool:
+        for fileurl in multipool.starmap(download_UNR,stat_year_tup):
```
Askaholic, line 178: BTW this whole thing could become a single list comprehension if you want
```suggestion
        for fileurl in multipool.starmap(download_UNR, stat_year_tup):
```

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -166,18 +172,22 @@ def download_tropo_delays(stats, years, writeDir='.', download=False, verbose=Fa
...
-    for s, y in itertools.product(stats, years):
-        if verbose:
-            print('Currently checking station {} in {}'.format(s, y))
-        fileurl = download_UNR(s, y, writeDir=writeDir,
-                               download=download, verbose=verbose)
-        # only record valid path
-        if fileurl:
-            results.append({'ID': s, 'year': y, 'path': fileurl})
-    return pd.DataFrame(results).set_index('ID')
+    stat_year_tup = (tup + (writeDir,) + (download,) + (verbose,) for tup in stat_year_tup)
```
Askaholic, line 175:
```suggestion
    stat_year_tup = ((*tup, writeDir, download, verbose) for tup in stat_year_tup)
```
or even 
```suggestion
    stat_year_tup = (stat, year, writeDir, download verbose) for stat, year in stat_year_tup)
```

Tuples are immutable so it's generally best to avoid concatenating them together (each + will copy the tuple).

tools/RAiDER/getStationDelays.py
```diff
@@ -134,10 +136,10 @@ def get_delays(stationFile, filename, returnTime=None):
     allstationTarfiles.sort()
     del ziprepo
 
-    return allstationTarfiles
+    return
```
Askaholic, line 139: If it's at the end of the function then you don't need an explicit return.

tools/RAiDER/statsPlot.py
```diff
@@ -48,6 +51,8 @@ def create_parser():
                           help='Specified input unit. Input will be converted into m if not already in m.')
     userinps.add_argument('-w', '--workdir', dest='workdir', default='./',
                           help='Specify directory to deposit all outputs. Default is local directory where script is launched.')
+    userinps.add_argument('--cpus', dest='numCPUs', type=parse_cpus, default=8,
+                          help='Specify number of cpus to be used for multiprocessing. May specify "all" at your own discretion.')
```
Askaholic, line 55: We could split the `create_parser` function into multiple functions so that we only need to define shared arguments in one place. For example:

```python3
def add_common_arguments(parser):
    parser.add_argument('--verbose, ...)
    parser.add_argument('--cpus', ...)
    ...
```

I'm not saying to necessarily do it in this PR, but it's something to keep in mind for the future, so we can avoid copy/pasting the same strings around.

tools/RAiDER/statsPlot.py
```diff
@@ -527,6 +532,21 @@ def _convert_SI(self, val, unit_in, unit_out):
...
+        '''
+        Return index of grid cell which intersects with station
+        Note: Fast, but assumes station locations don't change
+        '''
+        coord = Point((self.unique_points[1][self.unique_points[0].index(
+            stat_ID)], self.unique_points[2][self.unique_points[0].index(stat_ID)]))
+        # Get grid cell polygon which intersect with station coordinate
+        grid_int = self.polygon_tree.query(coord)[0]
+        # Pass corresponding grid cell index
+        if grid_int == []:
```
Askaholic, line 545:
```suggestion
        if not grid_int:
```

tools/RAiDER/statsPlot.py
```diff
@@ -527,6 +532,21 @@ def _convert_SI(self, val, unit_in, unit_out):
...
+            return stat_ID, 'NaN'
+        else:
+            return stat_ID, self.polygon_dict[id(grid_int)]
```
Askaholic, line 548: I would write it like this:
```suggestion
        if grid_int:
            return stat_ID, self.polygon_dict[id(grid_int)]
        return stat_ID, 'NaN'
```

Also I don't think I've ever seen `id` actually used in python before.

tools/RAiDER/statsPlot.py
```diff
@@ -598,22 +618,22 @@ def create_DF(self):
...
-                idtogrid_dict[i] = 'NaN'
+        self.unique_points = self.df.groupby(['ID', 'Lon', 'Lat']).size()
+        self.unique_points = [self.unique_points.index.get_level_values('ID').tolist(), self.unique_points.index.get_level_values(
+            'Lon').tolist(), self.unique_points.index.get_level_values('Lat').tolist()]
+        # Initiate R-tree of gridded array domain
+        self.polygon_dict = dict((id(pt), i) for i, pt in enumerate(append_poly))
+        self.polygon_tree = STRtree(append_poly)
+        # Parallelize check of station intersection with grid cells
+        with multiprocessing.Pool(self.numCPUs) as multipool:
+            for stat_ID,grd_index in multipool.starmap(self._check_stationgrid_intersection, itertools.product(self.unique_points[0])):
```
Askaholic, line 631: I would split the long line like this

```suggestion
            for stat_ID, grd_index in multipool.starmap(
                self._check_stationgrid_intersection, 
                itertools.product(self.unique_points[0])
            ):
```

tools/RAiDER/downloadGNSSDelays.py
```diff
@@ -80,6 +81,11 @@ def create_parser():
         help='Physically download data. Note this option is not necessary to proceed with statistical analyses, as data can be handled virtually in the program.',
         action='store_true', dest='download', default=False)
 
+    misc.add_argument(
+        '--cpus',
+        help='Specify number of cpus to be used for multiprocessing. May specify "all" at your own discretion.',
+        dest='numCPUs', type=parse_cpus, default=4)
```
Askaholic, line 87: Defaulting to 4 instead of 8?

tools/RAiDER/downloadGNSSDelays.py
Askaholic: They are computed based on the number of chunks.

tools/RAiDER/statsPlot.py
sssangha: As we discussed in the meeting, I will leave these out for now. What I think would be more useful is a progress bar. Will brainstorm in a separate issue ticket.

tools/RAiDER/downloadGNSSDelays.py
sssangha: Not sure about the upper bound, the default of 8 should be fine and if a user specifies anything higher it is their own responsibility.

For a negative value, would you think it be better to raise an exception and exit, or just pass the absolute value? I hesitate with the latter because such an input error would be handled without transparency under the hood.

tools/RAiDER/downloadGNSSDelays.py
sssangha: Oops, typo!

tools/RAiDER/statsPlot.py
sssangha: Ok, will move away from multiprocessing for this particular routine anyways though. Will give more detail in my update.

tools/RAiDER/statsPlot.py
sssangha: Yeah, I think it's a clean way of handling it in this case instead of passing a large dictionary filled with polygons spanning the global. Also a new idea I encountered in the depths of stackoverflow!

tools/RAiDER/statsPlot.py
sssangha: Yeah, probably better to address in a separate PR. Will raise an issue ticket.

sssangha:
Implementation of multiprocessing for deducing grid-intersections with station locations in ```statsPlot.py``` stalled out at the end of the loop when leveraged in tangent with the ```STRtree``` module from ```shapely```. Multiprocessing alone looped through the global archive in ~

sssangha:
```statsPlot.py``` is now updated to support multiprocessing of variograms, a time-consuming process that could originally could take dozens of hours if a user wished to general variograms for every year, for every grid-cell.

Several minor changes also integrated in ```statsPlot.py``` such as explicitly suppressing a numpy ```Runtimewarning``` when handling empty arrays, basic reorganization of the variogram class and input arguments, and clean-up of loaded modules.

sssangha:
Ok, tracked down an issue where outputs weren't successfully appended for between parallel jobs. This has been fixed.

Furthermore, some more specific numpy warning messages are suppressed to avoid inundating the user's screen and a textfile record of grid-cell locations are appropriately sorted.

Now everything looks good on my end so please feel free to share any further questions/concerns before we merge.

tools/RAiDER/downloadGNSSDelays.py
Askaholic: Raise an exception. If you specify negative CPU's you should get an error

sssangha: Yes I agree, done.

PR #75, opened 2020-07-22
Issue/#47 Add python logging

Askaholic:
This replaces all print statements with calls to a `logging.Logger` instance. The nice thing about this is we don't need to pass around 'verbose' flags anymore. Instead we can make use of logging levels to handle that automatically.

I also added color to highlight warnings and errors that are printed out to the screen. This should make it really easy to spot potential misconfiguration even in verbose mode:
image

Additionally all warnings and errors are logged to a file which then looks like this:
```
[2020-07-22 13:06:27,806]          tropo_delay:145   WARNING    Weather model already exists, please remove it ("test/scenario_1/weather_files/ERA5_2020-01-03T23_00_00_15.75N_18.25N_-103.25E_-99.75E.h5") if you want to create a new one.
[2020-07-22 13:06:27,806]         download_dem:51    WARNING    DEM already exists in /Users/rweeden/Documents/Repos/RAiDER/test/scenario_1/geom, checking shape
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Makefile
 dummy*
 weather_files
 test/data
+*.log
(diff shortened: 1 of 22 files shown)


dbekaert:
@Askaholic this looks great!

PR #77, opened 2020-07-23
Season interval properly split + postscript warning suppressed

sssangha:
Seasonal interval inputs, which are specified as an earlier date followed by a later date (e.g. '03-21 06-21') were not properly split, raising an error when trying to initialize datetime variables. This issue is fixed.

Furthermore, postscript warnings suppressed when generating variogram plots.

### DIFF
--- a/tools/RAiDER/statsPlot.py
+++ b/tools/RAiDER/statsPlot.py
@@ -21,9 +21,6 @@ import pandas as pd
 from pandas.plotting import register_matplotlib_converters
 from shapely.geometry import Point, Polygon
 from shapely.strtree import STRtree
-# supress matplotlib postscript warnings
-mpl._log.setLevel('ERROR')
-register_matplotlib_converters()
 
 from RAiDER.downloadGNSSDelays import parse_cpus
 from RAiDER.logger import logger
@@ -465,7 +462,7 @@ class VariogramAnalysis():
         plt.xlabel('Distance (°)')
         plt.ylabel('Dissimilarity (m\u00b2)')
         plt.legend(bbox_to_anchor=(1.02, 1),
-                   loc='upper left', borderaxespad=0.)
+                   loc='upper left', borderaxespad=0., framealpha=1.)
         # Plot empirical variogram
         if d_test is None and v_test is None:
             plt.title('Empirical variogram'+title_str)
--- a/tools/RAiDER/statsPlot.py
+++ b/tools/RAiDER/statsPlot.py
@@ -35,7 +35,7 @@ Perform basic statistical analyses concerning the spatiotemporal distribution of
 
 Specifically, make any of the following specified plot(s):
 scatterplot of station locations, total empirical and experimental variogram fits for data in each grid cell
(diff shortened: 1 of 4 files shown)


tools/RAiDER/statsPlot.py
```diff
@@ -624,6 +629,7 @@ def create_DF(self):
 
         # seasonal filter
         if self.seasonalinterval:
+            self.seasonalinterval = self.seasonalinterval.split()
```
dbekaert, line 632: This is not friendly for pre-2000 GPS data. Not that we will do it, but a user might?

Askaholic:
It depends on what the warning is. Maybe we should fix it instead of just ignoring it?

dbekaert:
@Askaholic what would you recommend for debug figures. In the past we used to make then if a users asked for debug.
https://github.com/dbekaert/RAiDER/blob/a3548e9f23d6280629f14bd2a15b75bcb5263f35/tools/RAiDER/statsPlot.py#L341

@sssangha Can you also edit the plotting such no display connection is required?

Askaholic:
Like I wrote in the comment, I would use an argument specifically for generating extra figures. Verbose is expected to not change the behavior of the program, only to print additional information as to what it is doing.

sssangha:
@dbekaert the warning message we get whenever saving figures was this: ```The PostScript backend does not support transparency; partially transparent artists will be rendered opaque.```

This was already suppressed for ```RaiderStats``` calls, but not for the ```VariogramAnalysis``` class, hence why I moved the suppression of the warnings up to the top of the code. I have been unable to pinpoint the source of the warning as I do not explicitly apply any transparencies, other than passing nans for invalid grid-cells.

Will also specify another flag for generating the debugging figures, but will have to then update the notebooks slightly.

Askaholic:
@sssangha From a quick web search I found https://github.com/matplotlib/matplotlib/issues/1740#issuecomment-515639297, maybe it's relevant?

tools/RAiDER/statsPlot.py
sssangha: The year I chose was arbitrary (other than intentionally not being a leap year) and doesn't have any implications on the code as only the time of the day is passed and used as a filter. I am forced to specify a year too, so I cannot go on with just a month and day.

sssangha:
> matplotlib/matplotlib#1740 (comment)

@Askaholic yup, thanks for pointing me to the right track! Apparently by default if ```framealpha=None```, then it passes a transparency value of 0.8 (see https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.legend.html). This initially didn't register to me because I did not realize ```None``` would default to a semitransparent display.

This has been fixed now. Warning suppressions are removed.

sssangha:
Ok, updated the parser, main function call and wrapper in bin to appropriately reflect the parser options. It looks like ```numCPUs``` was missing.

Let me know if the code looks good to go on your end.

dbekaert:
@sssangha did you remove the verbose to another term for plotting the debug figures?

Issue #78, opened 2020-07-23
refractor input parsing to be modular and assert correct inputs


dbekaert:
@Askaholic discussed in other issue tickets to remove duplication on shared input arguments and their syntax between different programs.

In addition we discussed capturing driver errors early one.  So, should we also implement a verification that input arguments are supplied correctly, directly after parsing?
- e.g. assumption on date order for seasonality?
- e.g. verifying dates and time are in correct and recognizable format

Lets discuss below.

Askaholic:
I think a lot of the input validation can be done in the parsing stage directly by supplying parsing functions through the `type` parameter.

dbekaert:
Would you recommend a single location for this for the parser and checks? https://github.com/dbekaert/RAiDER/blob/a3548e9f23d6280629f14bd2a15b75bcb5263f35/tools/RAiDER/statsPlot.py#L23

Few examples for discussion stimulation:
- This does not enforce the date formatting to be checked in the parser: https://github.com/dbekaert/RAiDER/blob/a3548e9f23d6280629f14bd2a15b75bcb5263f35/tools/RAiDER/downloadGNSSDelays.py#L72 

- The checking of the bounding box is done in the code and not in the parser. Parser is requiring a string. @Askaholic i assume you want to do this as part of the parser? https://github.com/dbekaert/RAiDER/blob/a3548e9f23d6280629f14bd2a15b75bcb5263f35/tools/RAiDER/downloadGNSSDelays.py#L326-L337

- this does not check the order of the dates or the formating as part of the parser: https://github.com/dbekaert/RAiDER/blob/a3548e9f23d6280629f14bd2a15b75bcb5263f35/tools/RAiDER/statsPlot.py#L70

Askaholic:
Yea I'm thinking some directory structure like this:

```
RAiDER/
    cli/
        __init__.py
        validators.py
        parser.py
```

We would put the parameter parsing functions in validators.py. I think you're pretty much free to do as much input validation as you want in the parsing functions even checking that files exist etc: https://docs.python.org/3/library/argparse.html#type

jlmaurer:
@dbekaert do you want to close this issue re #82 or leave it open for further work? @Askaholic notes that more could be done.

Askaholic:
Don't close it, it's far from being done.

PR #82, opened 2020-07-24
Refactor some arguments to a cli module

Askaholic:
There really is a ton more that could be done, but it's going to take a lot of work so I'm not sure it's worth it at this point. Everything in `checkArgs` needs to be moved into the argument parser. The `argparse` module supports doing everything that `checkArgs` currently does, but in a clean and reusable way.

I hope that this will at least provide some examples for how to add argument parsing correctly in the future, and how to add unit tests for the argument parser.

Related: #78

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,13 +6,12 @@
 # RESERVED. United States Government Sponsorship acknowledged.
 #
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
 import glob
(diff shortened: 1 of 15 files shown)


jlmaurer:
@Askaholic I like this quite a lot, seems like a very nice way to handle command-line arguments especially with multiple apps like we have. I wasn't familiar with this module (cli) before, do you have a recommended starting point/set of examples website besides the basic documention?

Askaholic:
> do you have a recommended starting point/set of examples website besides the basic documention?

I just use the official documentation https://docs.python.org/3/library/argparse.html which is pretty good. If I have any specific questions about implementation, I look at the source code.

PR #84, opened 2020-07-27
Add coverage report

### LINKED ISSUE #80: Code-coverage though Pytest and coveralls
Per discussion with @Askaholic:

- it woudl be good to include a coverage report as part of pytest.
This allows us to see how much of the unit-test cover the code; and would highlight gabs.

- adding coveralls as application which would then test the coverage for new code to be added in each PR. It would take the coverage report from the py-test and visualized it in the PR like CircleCI. https://coveralls.io/

Askaholic:
Add unit test coverage reporting to the CI build. Because there are some compiled modules, the coverage is recorded on the installed package resulting in coverage paths looking like `/path/to/conda/environment/RAiDER/__init__.py`. To get a proper coverage report we need to strip off the conda environment part, and replace it with the path to the source code from the repo.

All that's left is to set up coveralls and set the `COVERALLS_REPO_TOKEN` environment variable in the circleci build.

Closes #80

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,10 +10,8 @@ jobs:
       - run:
           name: Setup Miniconda
           command: |
-            apt update
-            apt install -y wget
-            apt-get update --yes && apt-get upgrade --yes
-            apt-get install g++ --yes
+            apt update --yes && apt-get upgrade --yes
+            apt install -y --no-install-recommends wget ca-certificates g++
             cd $HOME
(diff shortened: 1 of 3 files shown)


dbekaert:
@jlmaurer @Askaholic Looks above seems to capture it. We are doing an awesome job of 33% code checks! Thats better than 0.

dbekaert:
@Askaholic Any concern for the last git message in the circleCI build on GIT?

PR #95, opened 2020-07-30
Add fill value to c++ interpolate function

Askaholic:
Needed in order to replace calls to the scipy regular grid interpolator during delay calculation.


jlmaurer:
To actually replicate RegularGridInterpolator would mean raising a ValueError unless bounds_error is False. Is this something we want to do? From tests it seems the current default is to do linear extrapolation.

Askaholic:
I guess I leave that up to you. I could easily make it raise an exception, but it seems like everywhere you were calling it you turned off the bounds checking anyways and just had it use the fill value.

Issue #100, opened 2020-08-03
Install fails [BUG]


sssangha:
**Describe the bug**
Build of library bindings fails using existing conda environment recipe because c++17 compliant compiler wasn't include.
Refer to this error at the end of python build:
```
creating build/temp.linux-x86_64-3.7/tools/bindings/interpolate/src
gcc -pthread -B /u/leffe-data/ssangha/conda_installation/stable_july30_2020/envs/RAiDER/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/u/leffe-data/ssangha/conda_installation/stable_july30_2020/RAiDER/.eggs/pybind11-2.5.0-py3.7.egg/pybind11/include -I/u/leffe-data/ssangha/conda_installation/stable_july30_2020/envs/RAiDER/include/python3.7m -c tools/bindings/interpolate/src/interpolate.cpp -o build/temp.linux-x86_64-3.7/tools/bindings/interpolate/src/interpolate.o -std=c++17
gcc: error: unrecognized command line option '-std=c++17'
error: command 'gcc' failed with exit status 1
```

**Desktop (please complete the following information):**
 - RAiDER git tag: ```937ca731dfc5bb2fca676b2f8ddce0d75d7eedd6```
 - OS

dbekaert:
@jlmaurer we likely want to keep it easy but also system independent.

PR #101, opened 2020-08-03
Update environment yml files

sssangha:
## Description
Environment yml files updated to include c++17 compliant compiler prerequisite for successful python build. This addresses issue #100

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -31,6 +31,8 @@ dependencies:
  - xarray
  - shapely
  - pydap>3.2.2|<3.2.2
+ - cxx-compiler
+ - sysroot_linux-64
  # Dev dependencies. Only needed for contributing
  - pytest
  - pytest-cov
--- a/environment.yml
+++ b/environment.yml
@@ -30,3 +30,5 @@ dependencies:
  - xarray
  - shapely
  - pydap>3.2.2|<3.2.2
+ - cxx-compiler
+ - sysroot_linux-64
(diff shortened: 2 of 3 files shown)


environment.yml
```diff
@@ -30,3 +30,5 @@ dependencies:
  - xarray
  - shapely
  - pydap>3.2.2|<3.2.2
+ - cxx-compiler
+ - sysroot_linux-64
```
dbekaert, line 34: does this have any impact depending on the operating system?

environment-dev.yml
```diff
@@ -31,6 +31,8 @@ dependencies:
  - xarray
  - shapely
  - pydap>3.2.2|<3.2.2
+ - cxx-compiler
+ - sysroot_linux-64
```
dbekaert, line 35: i guess the circle CI code has a separate compiler as it completed successfully before.

environment.yml
sssangha: I'm don't believe so, but I'm not completely confident. These packages should be segregated to the RAiDER environment from my understanding and shouldn't be affected by the operating system. Perhaps you could confirm/correct this @Askaholic ?

Askaholic: I am not familiar with conda, I'll have to look it up.

environment-dev.yml
Askaholic: Probably the reason is that we have `apt install g++` in the circle ci config file, so it would be using the system's globally installed c++ compiler, and not anything from conda.

sssangha:
@dbekaert, I confirmed this build is stable on MacOS

PR #112, opened 2020-08-19
weather model update with bug fix

leiyangleon:
- [ ] GMAO worst-case lag time updated to guarantee the data files exist
- [ ] added MERRA-2 model support where the lag time is updated dynamically to guarantee the data files exist
- [ ] interpolation along vertical axis in the weather model base class has been refactored to replace the repeated uses in the GMAO and MERRA-2 readers
- [ ] UTC time zone is now enforced for all weather models

### DIFF
--- a/tools/RAiDER/models/gmao.py
+++ b/tools/RAiDER/models/gmao.py
@@ -1,11 +1,6 @@
 import datetime as dt
-import os
-import re
-
 import numpy as np
 from pyproj import CRS
-
-from RAiDER.interpolate import interpolate_along_axis
 from RAiDER.models.weatherModel import WeatherModel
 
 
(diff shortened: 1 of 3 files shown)


leiyangleon:
@jlmaurer I still cannot trigger the circleCI check. Talked to Rohan. It might be permission issue that can be solved by the owner of the repo, which David. We can still merge without ci check.

dbekaert:
Believe you have an outstanding invite to join the repository team.
Circleci only get triggered for the team collaborators pr.

Cheers D.

leiyangleon:
> Believe you have an outstanding invite to join the repository team. Circleci only get triggered for the team collaborators pr. Cheers D.
> […](#)

I had received one a while ago. I remember I did accept it. Could you send another invite just to make sure? The old one expired after 7 days. Thanks.

PR #133, opened 2020-09-29
Update weather model

leiyangleon:
## Checklist:

- [ ] Refined weather model debugging module
- [ ] Added HRES reader
- [ ] Fixed minor bugs (e.g. data availability range, data rotation/flipping) in weather model readers, e.g. ERA-I, ECMWF model level base class (for use in HRES), ERA-5
- [ ] Updated weather model documentation for using ECMWF api 
- [ ] Updated again the unit test (scenario 1 and 2) files since ERA-5 reader is changed in this PR


tools/RAiDER/models/hres.py
```diff
@@ -0,0 +1,235 @@
...
+        # model constants
+        self._k1 = 0.776   # [K/Pa]
+        self._k2 = 0.233   # [K/Pa]
+        self._k3 = 3.75e3  # [K^2/Pa]
+
+        # 9 km horizontal grid spacing
+        self._lon_res = 9. / 111
+        self._lat_res = 9. / 111
+        self._x_res = 9. / 111
+        self._y_res = 9. / 111
```
dbekaert, line 28: Can you add a comment that this is only used for extending the download-buffer, i.e. not in subsequent processing.

dbekaert (approved):
Looks good to me.

jlmaurer (approved):

PR #141, opened 2020-10-06
Update ncmr model reader

leiyangleon:
## Motivation

Since the downloaded NCMR file contains the global data for each specified date and time, the model reader tried to read huge data volume, most of which are not used. This is time consuming. 

## Description

- [x] Added data slicing to only read the subset within the lat/lon bounds specified by the user. **_This brings the runtime from 20 mins per run to 2 secs._**
- [x] Cleaned the code by removing commented/unused lines.
- [x] Compared the debugging plots of NCMR with GMAO and ERA-5 ones. They look reasonable.


leiyangleon:
@kam3545 Please check out the changes I made to the NCMR model reader.

jlmaurer (approved):
seems to look good to me!

PR #143, opened 2020-10-14
Fix ncmr bug longitude cross zero

leiyangleon:
## Description

NCMR data has longitude ranging from 0 to 360, where 0->180 is normal and 180->360 is equivalent to -180->0. So if the longitude range crosses zero, there will be problems. In this PR, the data slicing are performed in two parts when longitude range crosses zero, i.e. the 1st part is for positive longitude and the 2nd part for negative longitude. The results from the two parts are finally appended together as if the data were sliced from other weather models (e.g. GMAO and MERRA-2) with longitude ranging from -180 to 180.

## Type of change

## Checklist:


jlmaurer (approved):
Looks good to me!

PR #146, opened 2020-10-21
Fix readLL bugs and add unit tests

jlmaurer:
## Description

The readLL function parses the command-line inputs, but there is some funkiness in how arguments get unpacked, because argparse wraps the arguments in a tuple. This wasn't being correctly handled (i.e. it was assumed that the bounding box was a list, instead of a list inside a tuple), which led to the error. 

In addition, I've added unit tests to ensure that this kind of problem gets caught during testing rather than after the PR. 

## Motivation and Context
Fixes #145 

## How Has This Been Tested?
I added additional tests to test_llreader.py to check each type of input option (lat/lon files, bounding box, GPS station file) and successfully passes. 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have added an explanation of what changes do and why they should be included.
- [x] I have written new tests for the changes.
- [x] I have successfully ran tests with changes locally.

PR #147, opened 2020-10-21
Fix logger implementation so that it actually works as intended

jlmaurer:
Currently the logger isn't working as intended. Removed the direct calls to the Python logging module and added the custom logger Rohan built. Does not affect or change any unit tests. With this change the logging files get written as expected.

## Type of change


jlmaurer:
Same comment here as on #146 for the breaking tests.

dbekaert:
@jlmaurer can you rebase.

dbekaert (approved):
LGTM

PR #149, opened 2020-10-22
Fix demdownloader

jlmaurer:
## Description
Fixes the DEM change adopted by OpenTopography

## Motivation and Context
The old DEM URL no longer works. dem.py has been refactored with the new API. See [here](https://portal.opentopography.org/apidocs/#/Public/getGlobalDem) for more details and example.

Screen Shot 2020-10-22 at 2 34 25 PM

## How Has This Been Tested?
Updated the unit test suite to account for the changes. 

## Type of change

## Checklist:


tools/RAiDER/dem.py
```diff
@@ -89,17 +89,18 @@ def forceNDArray(arg):
...
-        outName=None, 
-        ndv=0., 
-        buf = 0.02
-    ):
+    lats,
+    lons,
+    save_flag='new',
+    checkDEM=True,
+    outName=None,
+    ndv=0.,
```
dbekaert, line 97: Can you not pull that from the data meta-data directly?
Like to avoid hard-corded values, or at least have a check if it exist and can be updated.

jlmaurer: Yes, good point. You mean the no-data value? I don't think that the ndv value is even getting used now so I'll remove it.

jlmaurer: @dbekaert got it fixed and pulling the noDataValue now directly from the DEM.

tools/RAiDER/dem.py
```diff
@@ -243,12 +238,50 @@ def isInside(extent1, extent2):
...
+    ySize = ds.RasterYSize
+    geoProj = ds.GetProjection()
+    trans = ds.GetGeoTransform()
+    Nbands = ds.RasterCount
+    if band_num is None:
+        band_num = 1
+        print('Using band one for dataType')
+
+    dType = ds.GetRasterBand(band_num).DataType
+    noDataVal = ds.GetRasterBand(band_num).GetNoDataValue()
```
dbekaert, line 283: This is ok when there is a no-data value associated with the file. 
Should have a catch for the case when that is not the case.

e.g. user provides a DEM as an envi file without that field.

dbekaert (approved):
LGTM

PR #150, opened 2020-10-23
Handle crash raised by empty station

sssangha:
## Description
Program crashes when parsing stations which are empty for a specified time.

The updates circumvent this issue, and an exception is raised in case no valid delays are parsed for a given time/region.


jlmaurer (approved):

jlmaurer:
I'm going to go ahead and merge this in as it clearly shouldn't break any unit tests with the new DEM fix.

Issue #151, opened 2020-10-24
Ability to sample interval in date range


dbekaert:
Add the ability to provide a stepsize for a data range, i.e. 12 day intervals etc in support of NISAR work.

dbekaert:
closed via #152

PR #152, opened 2020-10-24
adding ability to provide a stepsize for the date range

dbekaert:
## Description

Adding the ability to provide a stepsize for the range of dates to be downloaded

## Motivation and Context

If it fixes an open issue: #151 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] I have successfully ran tests with your changes locally.
- [ x] I have updated the documentation accordingly.


dbekaert:
@jlmaurer Perhaps have a look at the PR and see if you like how i did this. In the end, i am allowing for a third option which is the integer day sampling.  If you agree on that, we can update the expected circleCI unit test to allow for that.

jlmaurer (approved):
I think everything looks good! I'm not sure why Geo2rdr.cpp is in there, I think that might be one we need to add to the .gitignore. But it's probably fine for now.

tools/RAiDER/cli/validators.py
```diff
@@ -127,13 +127,20 @@ def __init__(
...
-            raise ArgumentError(self, "Only 1 or 2 dates may be supplied")
+        if len(values) > 3 or not values:
+            raise ArgumentError(self, "Only 1, 2 dates, or 2 dates and interval may be supplied")
 
         if len(values) == 2:
             start, end = values
             values = [start + timedelta(days=k)
-                      for k in range((end - start).days + 1)]
+                      for k in range(0,(end - start).days + 1,1)]
+        elif len(values) == 3:
```
jlmaurer, line 137:
```suggestion
        elif len(values) == 3:
            start, end, stepsize = values
            try:
                if not stepsize.year==0:
                    raise ArgumentError(self, "The stepsize should be in integer days")
             ...
```

jlmaurer: This should get you to pass the currenlty failing unit test

jlmaurer:
@dbekaert I added a suggestion to fix the currently failing unit test. Basically, if you have three arguments but the third is a regular date it will not fail, which it should. I just added an example line that should fail if the year is not zero; you could modify it as needed.

PR #174, opened 2020-11-05
Save pydap2hdf5

leiyangleon:
## Description

Save the OpenDAP/PyDAP-retrieved weather model data (GMAO and MERRA-2) to an HDF5 file

## Type of change

## Checklist:


leiyangleon:
@jlmaurer @dbekaert same problem of missing osr module. I think this should be installed when installing gdal. Not sure what happened exactly.

dbekaert (approved):
@leiyangleon looks good to me.

dbekaert:
fixed the issue of OSR via #176

tools/RAiDER/processWM.py
```diff
@@ -35,6 +35,9 @@ def getWMFilename(weather_model_name, time, outLoc):
         )
     )
 
+    if weather_model_name is 'GMAO' or 'MERRA2':
+        f = f[:-2]+'h5'
+
```
dbekaert, line 40: @leiyangleon i fixed the "osr" issue we have and confirmed it went fine before i merged this one. I believe this change is impacting one of the unit test to fail for ERA5. Can you have a closer look?

Issue #171, opened 2020-11-03
DEM download chunking


dbekaert:
ARIA-tools has included a chunking for DEM download. https://github.com/aria-tools/ARIA-tools/pull/240
We should port this back to address large region issues.
@sssangha  is that something you could do?

dbekaert:
Thanks @sssangha, per #180

PR #180, opened 2020-11-05
Handling of large DEMs supported

sssangha:
Introduce chunking of larger DEMs. Addressed #171 .


sssangha:
@dbekaert ok, I confirmed my patch works, but for large areas the DEM download takes sufficiently long (>10 mins) such that the unit test times out.

dbekaert:
Looks something wrong with the bounds; when you are projecting to local coordiantes

dbekaert (approved):
LGTM @sssangha

PR #196, opened 2020-11-10
Save and load gridded array

sssangha:
Save and load gridded array from TIF files to avoid duplication of sorting/cropping from same initial CSV file for every run.

This allows more for a more efficient means to quickly plot and adjust plots.

Note that the saving of the TIF files works as intended, however I need to fix the part of the code which attempts to load the TIF file and pass global variables expected outside of the main class.

### DIFF
--- a/tools/RAiDER/statsPlot.py
+++ b/tools/RAiDER/statsPlot.py
@@ -134,7 +134,7 @@ def cmd_line_parse(iargs=None):
     return parser.parse_args(args=iargs)
 
 
-def save_gridfile(df, gridfile_type, fname, cbounds, plotbbox, spacing, unit, colorbarfmt='%.3f', stationsongrids=False, gdal_fmt=6):
+def save_gridfile(df, gridfile_type, fname, plotbbox, spacing, unit, colorbarfmt='%.3f', stationsongrids=False, gdal_fmt='float32'):
     '''
         Function to save gridded-arrays as GDAL-readable file.
     '''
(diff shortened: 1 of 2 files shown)


sssangha:
I patched up the loading, appears to work as intended on my end now.

As a recap, I will briefly outline the functionality of these updates through a few example calls:

#Generate and save variogram-derived range and sill to file
```raiderStats.py --file products/UNRcombinedGPS_ztd.csv -w maps -b '36 40 -124 -119' -ti '2018-01-01 2019-12-31' --seasonalinterval '03-21 06-21' -variogramplot -variogram_per_timeslice --cpus all -verbose -grid_to_raster```
*Note specify (-grid_to_raster) to save gridded files to pseudo-raster for easy re-plotting later. Here, since we just specified (-variogramplot), only the gridded sill and range values will be generated and saved to a file.

#Load variogram-derived range for re-plotting
```raiderStats.py --file products/ZTD_range_heatmap.tif -w maps --cpus all -verbose```
*Note, for re-plotting there is no need to specify what type of gridded array you are passing because that info is tracked in the metadata!

#Load variogram-derived sill for re-plotting
```raiderStats.py --file products/ZTD_sill_heatmap.tif -w maps --cpus all -verbose```
*Note, again for re-plotting there is no need to specify what type of gridded array you are passing because that info is tracked in the metadata!

dbekaert (approved):
thank @sssangha will try this out now.

PR #203, opened 2020-11-11
Wgs2utm

### LINKED ISSUE #201: Change variogram analysis from lon/lat distances to x/y

leiyangleon:
## Description
This one fixes #201 where lat/lon were changed to x/y in local UTM coordinates when generating stats plots. @sssangha @jlmaurer @dbekaert further changes of unit from m to km and also the colorbar unit of the range heatmap from deg to km needs to be done.

it would be nice to add a unit test on this converter. Not urgent for now.

## Type of change

## Checklist:


dbekaert (approved):
@leiyangleon this looks good to me.

@sssangha can you make the changes based on the km units?

tools/RAiDER/utilFcns.py
```diff
@@ -532,3 +532,84 @@ def writeWeatherVars2HDF5(lat, lon, x, y, z, q, p, t, proj, outName=None):
...
+    L = np.zeros(lon.shape,dtype='<U1')
+    X = lon.copy()
+    Y = lon.copy()
+    for ind in range(lon.__len__()):
+        longitude = lon[ind]
+        latitude = lat[ind]
+        z, l, x, y = project((longitude,latitude))
+        Z[ind] = z
+        L[ind] = l
+        X[ind] = x
```
dbekaert, line 595: @leiyangleon 

Im not that familiar with UTM. But understadn from reading that its not a contineous XY system across zone boundaries. See http://geokov.com/education/utm.aspx 
` When working across zone boundaries multiple coordinate systems must be used. Coordinates in adjacent zones do not have a mathematical relationship.`

We provide a list of longitude and latitude and then project to UTM based on the lonlat.
If not continuous across zones then the xy coordinates might not be wrt to the same origin. e.g. what would be the different of using the UTM zone of the centroid and then computing the xy wrt to that zone?

tools/RAiDER/utilFcns.py
```diff
@@ -532,3 +532,84 @@ def writeWeatherVars2HDF5(lat, lon, x, y, z, q, p, t, proj, outName=None):
...
+    if z not in _projections:
+        _projections[z] = pyproj.Proj(proj='utm', zone=z, ellps='WGS84')
+    if l < 'N':
+        y -= 10000000
+    lng, lat = _projections[z](x, y, inverse=True)
+    return (lng, lat)
+
+def WGS84_to_UTM(lon, lat):
+    shp = lat.shape
+    lon = np.ravel(lon)
```
jlmaurer, line 583:
```suggestion
    Z = lon.flatten() # .flatten() takes care of raveling and copying in one step. 
```

jlmaurer: You would do the same thing for X, Y

tools/RAiDER/utilFcns.py
```diff
@@ -532,3 +532,84 @@ def writeWeatherVars2HDF5(lat, lon, x, y, z, q, p, t, proj, outName=None):
...
+    lat = np.ravel(lat)
+    Z = lon.copy()
+    L = np.zeros(lon.shape,dtype='<U1')
+    X = lon.copy()
+    Y = lon.copy()
```
jlmaurer, line 588: Is this a typo? X = lon, Y = lat?

jlmaurer (commented):
@leiyangleon can you add a couple unit tests that verify the zone and letter for a couple lat/lon locations around the world? E.g. one point each on hawaii, UK, new zealand? That would just verify that the projection is working correctly.

tools/RAiDER/utilFcns.py
leiyangleon: @jlmaurer This is okay, because lon / lat can be changed (i.e. don't need to copy; returning the same pointer of the array is okay). While Z,L,X,Y are just initialized as new arrays with the same shape of lon (or lat). So X and Y are not typo, either.

leiyangleon (commented):
> @leiyangleon can you add a couple unit tests that verify the zone and letter for a couple lat/lon locations around the world? E.g. one point each on hawaii, UK, new zealand? That would just verify that the projection is working correctly.

@jlmaurer I did verify that already with dozens of zones on my machine. As I outlined in the PR description, unit tests would be great but next step.

tools/RAiDER/utilFcns.py
leiyangleon: @dbekaert Got it. Will look into that next week.

tools/RAiDER/utilFcns.py
dbekaert: @leiyangleon to confirm look suspicious

leiyangleon: @dbekaert same question @jlmaurer asked. See my above response to @jlmaurer. These are not typos but just initialization of arrays(Z,L,X,Y) with same shape as "lon". Values do not matter in the initialization, cuz they finally got assigned with the correct values at the lines 593-596. FYI, I have verified the results for dozens of zones and lat/lon's.

leiyangleon:
This one is closed because a new one (#216) has been issued.

PR #204, opened 2020-11-11
Functions for combining gnss and weather model delays

jlmaurer:
## Description

Adding in functions for combining GNSS ZTD and weather model delay estimates

## How Has This Been Tested?

Has not been tested

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/gnss/processDelayFiles.py
+++ b/tools/RAiDER/gnss/processDelayFiles.py
@@ -79,7 +79,7 @@ def create_parser():
         description=dedent("""\
             Combine delay files from a weather model and GPS Zenith delays
             Usage examples:
-            raiderCombine.py --gnss UNRCombined_gnss.csv --raiderLoc ERA5/ -o Combined_delays.csv 
+            raiderCombine.py --gnss UNRCombined_gnss.csv --raiderLoc ERA5/ --raider ERA5_combined_delays.csv -o Combined_delays.csv --column ERA5
             """)
     )
 
@@ -112,8 +112,8 @@ def create_parser():
     )
 
     p.add_argument(
-        '--name',
-        '-n' 
+        '--column',
+        '-c' 
         dest='column_name',
         help=dedent("""\
             Name of the column containing RAiDER delays.
--- a/tools/RAiDER/gnss/processDelayFiles.py
+++ b/tools/RAiDER/gnss/processDelayFiles.py
@@ -1,3 +1,4 @@
+import argparse
 import datetime
 import glob
 import os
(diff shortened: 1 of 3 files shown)


tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -0,0 +1,73 @@
...
+    '''
+    Merge a combined RAiDER delays file with a GPS ZTD delay file
+    '''
+    dfr = pd.read_csv(raiderFile)
+    dfz = readZTDFile(ztdFile)
+    dfc = dfr.merge(dfz[['ID', 'Datetime', 'ZTD']], how='inner', left_on=['Datetime', 'ID'], right_on=['Datetime', 'ID'], sort=True)
+    if outName is None:
+        return dfc
+    else:
+        dfc.to_csv(outName, index=False)
```
dbekaert, line 61: Does this make the assumption it ingest raider computed delays where at the GNSS locations of matching time and station?

What happens if the GNSS file has every 5 min and raider is for fixed UTC.
I understand that the filename written file will then we still every 5 min, but there will be a lot of blanks.

tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -0,0 +1,73 @@
+import datetime
+import glob
+import os
+
+import pandas as pd
```
dbekaert, line 5: @jlmaurer  Could you ad in a brief run-example. Or add a a parser __main__ for now.

tools/RAiDER/gnss/processDelayFiles.py
jlmaurer: Right now I  merge only on the intersection, so if dates are entirely missing in either file they are dropped. Ie inner join.

tools/RAiDER/gnss/processDelayFiles.py
jlmaurer: Yes, I'll add that tonight after I get out of service!

dbekaert: Ok do you want me to wait to merge it to have the example added?

jlmaurer: Sure that's fine, either way is fine with me.

jlmaurer: I'll add an rough example to the comments

jlmaurer:
Ipython example to get started: 

```
from processDelayFiles import *

loc = '.'
raider_file = 'raider_delays.csv'
ztd_file = 'UNRcombined_gnss.csv'

combineDelayFiles(raider_file, loc=loc)

 mergeDelayFiles(raider_file, ztd_file, 'Combines_delays.csv')
```

dbekaert (approved):
LGTM @jlmaurer  thanks

Issue #202, opened 2020-11-11
Stats Plots commands


dbekaert:

dbekaert:
@sssangha we can use the edit mode above to fill in details as we go along to track.

PR #206, opened 2020-11-12
Series of fixes for statsplot

sssangha:
Supports series of fixes outlined in #202, namely:
- Rounding of the colorbar ticks to two decimal points for floating values (as opposed to integer heatmaps of station locations).
- Variogram range to be reported in km units. Currently crude factor of 1010km applied, to be updated in PR #203 
- Remove the wording heatmap from the colorbar title
- Change "sill" to "Variance" and "range" to "Range"; "latitude" to "Latitude"; "longitude" to "Longitude"
- Ability to improve the dpm for the Figures, now toggled with ```-figdpi``` option
- Ability to change units of loaded arrays.

### DIFF
--- a/tools/RAiDER/statsPlot.py
+++ b/tools/RAiDER/statsPlot.py
@@ -302,6 +302,8 @@ class VariogramAnalysis():
         x, y = self._get_XY(x, y, indpars)
         dists = self._get_distances(
             np.array([[x[:, 0], y[:, 0]], [x[:, 1], y[:, 1]]]).T)
+        # TODO temp fix first multiply by 110,000 m as crude conversion from degrees to km
+        dists = dists*110000
         vario = self._get_variogram(samples[:, 0], samples[:, 1])
 
         return dists, vario
@@ -369,11 +371,6 @@ class VariogramAnalysis():
                                    loss='soft_l1', f_scale=0.1,
                                    args=(d, v, model))
 
-        # convert range to specified output unit
-        # TODO temp fix first multiply by 1,010,000 m as crude conversion from degrees to km
-        res_robust.x[0] = res_robust.x[0]*1010000
-        res_robust.x[0] = convert_SI(res_robust.x[0], 'm', 'km')
-
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", message="All-NaN slice encountered")
             d_test = np.linspace(0, np.nanmax(dists), 100)
@@ -383,7 +380,7 @@ class VariogramAnalysis():
         return res_robust, d_test, v_test
 
     # this would be expontential plus nugget
-    def __exponential__(self, parms, h):
+    def __exponential__(self, parms, h, nugget=False):
         '''
         returns a variogram model given a set of arguments and
         key-word arguments
@@ -392,7 +389,10 @@ class VariogramAnalysis():
         a, b, c = parms
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", message="overflow encountered in true_divide")
-            return b * (1 - np.exp(-h / a)) + c
+            if nugget:
+                return b * (1 - np.exp(-h / a)) + c
+            else:
+                return b * (1 - np.exp(-h / a))
 
     # this would be gaussian plus nugget
     def __gaussian__(self, parms, h):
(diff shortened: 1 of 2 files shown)


tools/bin/raiderStats.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/data2/ssangha/conda_installation/stable_jan10_2020/ariatools_conda/envs/RAiDER/bin/python3
```
jlmaurer, line 1: Looks like you made a temporary change, just need to switch back.

jlmaurer (commented):
Looks fine to me except for the one environment statement.

tools/RAiDER/statsPlot.py
```diff
@@ -351,6 +369,11 @@ def resid(x, d, v, m):
                                    loss='soft_l1', f_scale=0.1,
                                    args=(d, v, model))
 
+        # convert range to specified output unit
+        # TODO temp fix first multiply by 1,010,000 m as crude conversion from degrees to km
+        res_robust.x[0] = res_robust.x[0]*1010000
+        res_robust.x[0] = convert_SI(res_robust.x[0], 'm', 'km')
+
         with warnings.catch_warnings():
```
dbekaert, line 377: Probably the easiest would be that you just compute distance and multiply it there with 110km per degree so you do that change once.

tools/RAiDER/statsPlot.py
```diff
@@ -527,17 +550,25 @@ def plot_variogram(self, gridID, timeslice, coords, workdir='./', d_test=None, v
...
             plt.axhline(y=res_robust[1], color='g',
                         linestyle='--', label='ɣ\u0332\u00b2({}\u00b2)'.format(self.unit))
             plt.axvline(x=res_robust[0], color='c',
                         linestyle='--', label='h\u0332(°)')
         if d_test is not None and v_test is not None:
+            # TODO temp fix first multiply by 1,010,000 m as crude conversion from degrees to km
+            d_test = [i*1010000 for i in d_test]
             plt.plot(d_test, v_test, 'r-', label='experimental fit')
-        plt.xlabel('Distance (°)')
+        plt.xlabel('Distance (km)')
```
dbekaert, line 571: Then in future when a user ask for a different unit. We would change units from SI to requested unit just before making the plot.

i.e. assume all units up to just before plotting to be in SI standard units. 
Then change it just before calling the plotting.

tools/RAiDER/statsPlot.py
```diff
@@ -351,6 +369,11 @@ def resid(x, d, v, m):
                                    loss='soft_l1', f_scale=0.1,
                                    args=(d, v, model))
 
+        # convert range to specified output unit
+        # TODO temp fix first multiply by 1,010,000 m as crude conversion from degrees to km
+        res_robust.x[0] = res_robust.x[0]*1010000
```
dbekaert, line 374: Note that equator 1 degree is roughly 110km so should be * 110,000 to go from degree to m and not 1,010,000.

tools/RAiDER/statsPlot.py
```diff
@@ -351,6 +369,11 @@ def resid(x, d, v, m):
                                    loss='soft_l1', f_scale=0.1,
                                    args=(d, v, model))
 
+        # convert range to specified output unit
+        # TODO temp fix first multiply by 1,010,000 m as crude conversion from degrees to km
```
dbekaert, line 373: the comment is incorrect it is mixing conversion from degree to m and km.

dbekaert:
@sssangha can you also add in an exponential model without sill?

tools/bin/raiderStats.py
sssangha: Ok, fixed.

tools/RAiDER/statsPlot.py
sssangha: Figured out how to change the distance once. The bug I experience yesterday is gone.

tools/RAiDER/statsPlot.py
sssangha: Fixed

tools/RAiDER/statsPlot.py
sssangha: Fixed

tools/RAiDER/statsPlot.py
```diff
@@ -369,7 +389,10 @@ def __exponential__(self, parms, h):
         a, b, c = parms
```
jlmaurer, line 389: With nugget = False will this fail?

sssangha:
> @sssangha can you also add in an exponential model without sill?

You mean no nugget? By default I turned off the nugget flag in the exponential model. Will add in user-flag to toggle later.

jlmaurer (commented):
Just one place where the nugget = False could cause a problem.

tools/RAiDER/statsPlot.py
jlmaurer: I think I've convinced myself it should be ok.

jlmaurer (approved):

dbekaert (approved):
LGTM @sssangha

PR #212, opened 2020-11-14
Fix gnss issue

jlmaurer:
## Description

- Fix a bug in the way pd.concat was called 
- Add a try/except in case the gnss file already has datetime defined
- Add try/except to stats plot so that either "Datetime" or "Date" will work


dbekaert (approved):
LGTM, will test on dev directly.

PR #214, opened 2020-11-15
Insure datetimes are passed as timestamps, as expected by variogram c…

sssangha:
Insure datetimes are passed as timestamps, as expected by variogram class


dbekaert (approved):
thanks @sssangha

PR #215, opened 2020-11-16
Specify custom plot titles

sssangha:
Specify custom title for plot with ```-title``` flag (e.g. ```-title 'year 2009'```).

By default no title is passed in the plot, aside from default labels for the colorbar.


dbekaert (approved):
LGTM @sssangha thank you!

PR #216, opened 2020-11-17
Wgs2utm

### LINKED ISSUE #201: Change variogram analysis from lon/lat distances to x/y

leiyangleon:
## Description
This one fixes #201 where lat/lon were changed to x/y in local UTM coordinates when generating stats plots. 

Added the option of UTM conversion wrt the zone of the center of the region, i.e. by turning on the flag "`common_center=True`". For example, you can do "`WGS84_to_UTM(lon, lat, common_center=True)`". By default, that flag is set to False, i.e. if "`WGS84_to_UTM(lon, lat)`" is used.

Unit tests on the core converter "`project`" (single point) and the wrapper "`WGS84_to_UTM`" (multiple points; both with and without common center) have been included as well. 

## Type of change

## Checklist:


dbekaert:
@leiyangleon this looks good. For now we should make the centroid the default to handle UTC zone crossings (given we have a 5 degree box for our analysis). 

Do you have a rough thought if something weird (geometrically) happens if mix two zones versus one zone by zone?
e.g. perhaps a plot in local coordinates:
- points in Zone A wrt a neighbor zone 
- point in Zone A wrt a neighbor zone

Would suggest we over plot them together on the same figure (xy), but where you would have subtracted the local coordiante from a reference point (xpoint,ypoint). This would then allow you to overlay both figures together and see impact of geometry.
Would suggest to do this once for a larger grid (e.g. 5 degrees) around the equator and perhaps at higher latitude.

dbekaert:
adressed via #226 #225

Issue #219, opened 2020-11-25
[BUG] GNSS station ZTD errors out for 00 UTC and LGC1 station


dbekaert:
GNSS ZTD delays downloaded and then error out after ~12000 stations, killed for LGC1 station.

Command to replicate:
```
raiderDownloadGNSS.py --out products --years '2018' --returntime '00:00:00' --bounding_box '-89.90 89.90 -179.9 179.9'
```

The following error occurred: 
```
raiderDownloadGNSS.py --out products --years '2018' --returntime '00:00:00' --bounding_box '-89.90 89.90 -179.9 179.9'

requests.exceptions.ConnectionError: HTTPConnectionPool(host='geodesy.unr.edu', port=80): Max retries exceeded with url: /gps_timeseries/trop/LGC1/LGC1.2018.trop.zip (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f2fd8024bd0>: Failed to establish a new connection: [Errno -2] Name or service not known'))
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./tools/bin/raiderDownloadGNSS.py", line 23, in <module>
    inps.verbose
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/downloadGNSSDelays.py", line 364, in query_repos
    returnTime=returnTime
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/getStationDelays.py", line 200, in get_station_data
    multipool.starmap(get_delays_UNR, args)
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/multiprocessing/pool.py", line 276, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
requests.exceptions.ConnectionError: None: Max retries exceeded with url: /gps_timeseries/trop/LGC1/LGC1.2018.trop.zip (Caused by None)
```

bbuzz31:
should be fixed with https://github.com/dbekaert/RAiDER/pull/220

jlmaurer:
Closing as fixed with #220

PR #220, opened 2020-12-02
implement automatic retry

bbuzz31:
Addresses https://github.com/dbekaert/RAiDER/issues/219 by automatically retrying 'get' request to UNR

### DIFF
--- a/tools/RAiDER/downloadGNSSDelays.py
+++ b/tools/RAiDER/downloadGNSSDelays.py
@@ -17,6 +17,7 @@ import requests
 from RAiDER.cli.parser import add_cpus, add_out, add_verbose
 from RAiDER.logger import *
 from RAiDER.getStationDelays import get_station_data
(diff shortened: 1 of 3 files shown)


dbekaert:
@kam3545 Can you try if this resolved your issue?

bbuzz31:
As a disclaimer, there is no guarantee that this will fix anything.
In my experience these issue usually arise due to a problem in the server - client connection.
For example, my home internet connections isn’t the greatest, and I usually can’t download anything large without an interruption.

The additional retries I’ve implemented is merely an additional level of redundancy.

jlmaurer (approved):

kam3545:
I tested these changes, still downloading of all GNSS stations are stopped for few stations.

bbuzz31:
recommend you to rerun a few times and use an ethernet connection

PR #221, opened 2020-12-02
Netcdf4

### LINKED ISSUE #155: Netcdf4 instead of HDF5 for storing weather model
Store the data as netcdf4 CF compliant file instead of hdf5 such we can leverage GDAL for chunking of the model file without much manipulation.

leiyangleon:
## Description

This one fixes #155, where the manipulated weather model data (p/q/e) and the refractivity (both wet and hydrostatic) are recast to NetCDF4 format (in compliance with CF 1.6) for better chunking and checking/viewing via gdalinfo and QGIS.

Future work entails:

1. Compare top-left corner vs. pixel center for manually downloaded GMAO raw data and the output using the current function.

2. Save OpenDAP-downloaded new GMAO raw data (2017/12/1 on) as NetCDF4 as well

## Type of change

## Checklist:


leiyangleon:
@jlmaurer test_scenario_1 and _2 failed because they were seeking for h5 files rather than the new nc files. If you can quickly fix that, that would be great; otherwise, I will look at this next week.

leiyangleon:
@jlmaurer @dbekaert the above problem solved and all checks passed.

tools/RAiDER/delay.py
```diff
@@ -123,7 +126,7 @@ def tropo_delay(args):
             return None, None
 
         try:
-            weather_model.write2HDF5(weather_model_file)
+            weather_model.write2NETCDF4(weather_model_file)
```
jlmaurer, line 129: I think it would be good to leave the HDF5 method in there, and change to a generic "write" method that takes a file type as a parameter in order to maintain backwards compatibility with anything else in the codebase.

tools/RAiDER/delayFcns.py
```diff
@@ -100,12 +101,12 @@ def get_delays(
     t0 = time.time()
 
     # Get the weather model data
-    with h5py.File(wm_file, 'r') as f:
-        xs_wm = f['x'][()].copy()
-        ys_wm = f['y'][()].copy()
-        zs_wm = f['z'][()].copy()
-        wet = f['wet'][()].copy()
-        hydro = f['hydro'][()].copy()
+    with Dataset(wm_file, mode='r') as f:
```
jlmaurer, line 104: I know the HDF5 version was already here, but I'm thinking to start getting this massaged into a more modular form we should have a generic reader that also takes a file type, and implement both of these bits for a more generic code.

tools/RAiDER/delayFcns.py
```diff
@@ -286,8 +287,8 @@ def getProjFromWMFile(wm_file):
     '''
     Returns the projection of an HDF5 file
     '''
-    with h5py.File(wm_file, 'r') as f:
-        wm_proj = CRS.from_json(f['Projection'][()])
+    with Dataset(wm_file, mode='r') as f:
```
jlmaurer, line 290: Same thing here. We should have a NETCDF "projection reader" that compliments rather than replaces the HDF5 reader.

tools/RAiDER/models/weatherModel.py
```diff
@@ -581,73 +582,203 @@ def _checkForNans(self):
         self._t = fillna3D(self._t)
         self._e = fillna3D(self._e)
 
-    def write2HDF5(self, outName=None):
+    def write2NETCDF4(self, outName=None, NoDataValue=-3.4028234e+38, chunk=128):
```
jlmaurer, line 585: Here's where I think a generic "write" method with a file type parameter would be better. Suggestion: 
```suggestion
    def write(self, outName=None, fmt='netcdf4'):
        if fmt == 'netcdf4':
            # do your new code
        elif fmt == 'hdf5':
            # do the old code
        else:
            raise ValueError('fmt {} not a valid fmt for writing the weather model results')
            # (or we could simply change the fmt to the default value)
```

jlmaurer (commented):
@leiyangleon Made some comments, mostly about keeping backwards compatibility. No sense deleting functions we still use until we're sure we've got everything worked out. I think we should also function-alize both the original hdf5 code and your new netcdf writers to be more generic so that we can call them elsewhere as well. I'm working on that part, so if you make the changes in my review (mainly just keeping the old hdf5 stuff as an option) we could probably go ahead and merge in then I can work on the functionalizing.

PR #224, opened 2020-12-04
Fix several bugs related to CDS API and some general clean-up

jlmaurer:
## Description

- There were two lines in weatherModel.py that queried the CDS API, with a note that the logger wouldn't work without them. However, I tested removing them and the logger worked just fine. We definitely don't want those lines there because not all models use the CDS API. 
- Refactored to better handle errors when the config file for CDS API is missing or the server can't be reached
- Added a note that WRF or HDF5 are not fully implemented if a user calls those
- Fixed a few misc bugs related to passing height levels for only the weather model nodes that I found while testing above
- Cleaned up some formatting

## Motivation and Context

- Improves error messaging and gets rid of problematic call to CDS API in base weatherModel class

## How Has This Been Tested?

- Unit tests pass locally

## Type of change

## Checklist:

- [x] I have added an explanation of what the changes do and why I'd like us to include them.
- [ ] I have written new tests for my core changes, as applicable.
- [x] I have successfully ran tests with my changes locally.

### DIFF
--- a/tools/RAiDER/checkArgs.py
+++ b/tools/RAiDER/checkArgs.py
@@ -7,14 +7,16 @@
 #
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 import os
-from datetime import datetime
 
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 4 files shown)


jlmaurer:
@bbuzz31 This should address the poor error messages you were getting.

dbekaert (approved):
Looks good to me.

PR #227, opened 2020-12-10
Separate utilFcns into modules that make sense and provide a generic hdf5 writer

jlmaurer:
Generally cleans up the existing I/O functionality 

## Description

- Turns out that creating a generic HDF5 writer that does everything we want is hard. E.g. chunk size varying for different variables, creating empty variables, setting all the possible attributes and handling if they aren't specified, etc. 
- Replaces several existing HDF5 writers with a single generic writer, with placeholder for a future NETCDF writer 
- In line with this, splits out I/O functions from the "utilFcns" module in an ioFcns module for reading/writing functions
- Since I was splitting things out, I went ahead and did the rest of the utilFcns to be more explicit in module names
- After splitting, it became very obvious where fucntions weren't being tested so I added a bunch of unit tests.
- Looks like a lot of changes, but the only real code additions are the generic writer functions. All other changes are unit tests,  updating the module calls (changing utilFcns to ioFcns, etc.), or formatting.

## Motivation and Context

- Prepping for moving to generic NETCDF writers

## How Has This Been Tested?

- All unit tests pass locally

## Type of change

## Checklist:


jlmaurer:
Closing this as will push directly to the netcdef branch

Issue #164, opened 2020-10-29
pixel node verus pixel center registration of models


dbekaert:
Currently we write manually to hdf5.
Need to verify if do not make incorrect assumption for the models.
Once we are in netcdf we can let GDAL do the semantics.

leiyangleon:
Verified that by using the new [netcdf4 branch](https://github.com/dbekaert/RAiDER/tree/netcdf4), the coordinate grid in the NetCDF outputs is pixel centered (not top left corner aligned) for all of the weather models (i.e. ERA-5, ERA-5T, ERA-I, HRES, GMAO, MERRA-2, NCMR), and exactly aligned with the coordinate grid in the native weather model data (that are NetCDF files as well)

PR #229, opened 2020-12-16
Nc abstract

leiyangleon:
## Description

- An abstract/modular NetCDF writer (RAiDER.utilFcns.write2NETCDF4core) has been included and can be refactored for writing other data results to NetCDF files in CF1.6 format. To use this writer, users only need to write a wrapper function that specifies the global attributes and pass the dictionaries of individual data along with attributes to the abstract writer.
- Processed weather model data and refractivity results are now written to NetCDF files using the wrapper function (RAiDER.models.weatherModels.write2NETCDF4) that calls the abstract writer 
- OpenDAP/PyDAP-retrieved weather model data (GMAO and MERRA-2) are now written to NETCDF files using the wrapper function (RAiDER.utilFcns.writeWeatherVars2NETCDF4) that calls the abstract writer
- the two lines of cdsapi initialization have been added back to the weather model base class, because those missing logging info are important for debugging non-era5-type weather models, such as GMAO used here. Without a better way to solve this problem, suggest to keep these two lines here.
- Future work will be to solve #164 based on the NecCDF outputs comparing to the native .nc files downloaded from the GMAO server.

## Motivation and Context

This PR is related to an upgrade of #155 (with a modular NetCDF writer that can be refactored) and also establishes the base for #164.

## How Has This Been Tested?

1. 

Run the following commands and compare the NetCDF results with those saved as hdf5 and make sure the data are exactly the same
```
raiderDelay.py --date 20200103 --time 23:00:00 -b 39 40 -80 -78 --model GMAO --zref 15000 -v
raiderDelay.py --date 20200103 --time 23:00:00 -b 39 40 -80 -78 --model MERRA2 --zref 15000 -v
```

2. 

With the above commands, also compare the NetCDF results with those created using the original NetCDF writer (not the abstract/modular one) and make sure the results are exactly the same, i.e. remember to check both the fetched weather model raw data named as
`GMAO_2020_01_03_T23_00_00.nc`
and the processed weather model data and refractivity results named as
`GMAO_2020_01_03_T23_00_00_39.0N_40.0N_80.0W_78.0W.nc`
 

## Type of change

## Checklist:


tools/RAiDER/models/weatherModel.py
```diff
@@ -27,6 +27,11 @@ class WeatherModel(ABC):
     '''
 
     def __init__(self):
+        # without a better way to solve the problem, suggest to add this back ; otherwise missing a lot of logging info needed for all other models e.g. GMAO (except those call cdsapi directly like ERA-5)  -- Yang (update: Dec 15 2020)
```
jlmaurer: I recommend not to include this here since it is not related to the netcdf. We can include it back in its own PR.

jlmaurer (approved):

PR #230, opened 2020-12-18
Fix multiprocessing issue

jlmaurer:
Adds in a simple for-loop option for processing delays for multiple dates.

## Description

- multiprocessing works when download_only option is called, but the delays are already individually parallelized so the current implementation breaks if the delay calculation is attempted
- This PR implements a simple way to run delay calculations for multiple dates without requiring parallelization
- Is still based on user input, but now raises a RuntimeError if parallel > 1 and download_only is False.

## Motivation and Context

Fixes #206 for now; later enhancements to improve the overall parallelization strategy

## How Has This Been Tested?

- Passes unit tests locally

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -10,6 +10,7 @@ import os
 
 import h5py
 import numpy as np
+from netCDF4 import Dataset
 
 import RAiDER.delayFcns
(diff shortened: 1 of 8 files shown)


jlmaurer:
Closing to re-open with latest dev changes

PR #239, opened 2021-01-25
Subsetting by input bounding box

jlmaurer:
This PR provides logic to subset the HRES weather model by the user-input bounding box at the loading stage, i.e. after file download. 

## Description

- Needed to fix a few bugs
- Adding subseting to _makeDataCubes method in models/hres.py
- Currently adding at the model level, not sure if a more generic/uniform approach is possible given the different APIs

## Motivation and Context

- This loads in a subset of the weather model; required in order to load a subset of HRES models into memory. 

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/checkArgs.py
+++ b/tools/RAiDER/checkArgs.py
@@ -31,8 +31,13 @@ def checkArgs(args, p):
             if args.outformat.lower() != 'hdf5':
                 raise RuntimeError('HDF5 must be used with height levels')
 
+    if args.wmLoc is not None:
+        wmLoc = args.wmLoc
+    else:
+        wmLoc = os.path.join(args.out, 'weather_files')
     if not os.path.exists(wmLoc):
         os.mkdir(wmLoc)
+
     # Query Area
     lat, lon, llproj, bounds, flag = readLL(args.query_area)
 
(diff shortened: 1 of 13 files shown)


jlmaurer:
Hey @dbekaert and @cmarshak I've implemented a basic subsetter for ECMWF-High resolution model using xarray. @cmarshak this is not the implementation we will eventually want, but I just wanted to get something slotted in that can do the subsetting already before we have the full xarray logic worked out, as I think that will take some more significant refactoring. I also added the xarray logic to the ERA5 model just as a check and everything seems to be working fine. Currently our unit test coverage has decreased, but we'll deal with that when we do the fuller implementation.

jlmaurer:
@dbekaert it worked! I was able to process ZTD for the weather model nodes for 1/4 of the world using ECMWF-HRES on my Linux machine (130 GB memory). I peaked out at around 20% memory usage. I seemed to be limited somewhat by the single processor (100% most of the time), but may have also been limited by I/O speed. Made some fixes. Next PR for me is the filename fix.

jlmaurer:
@dbekaert I went back and checked that the output of the original and new versions is the same. There was an issue with the bounding box, the buffer zone around the query points was not getting preserved due to the way I was handling the lat/lon bounds. I fixed that and now the old and new versions give identical arrays.

dbekaert (approved):
Thanks for confirming @jlmaurer! LGTM

PR #243, opened 2021-02-02
Fix some weather model stuff

jlmaurer:
- This is only one part of the required changes. The main thing this does is to fix the file-naming convention and the checking logic so that if the weather model has already been processed, the code doesn't re-run the processing. This was implemented before but somehow got lost. 
- Also fixes a bug or two in GMAO

- All units tests pass locally

## Type of change

## Checklist:


dbekaert (approved):
LGTM @jlmaurer

Issue #251, opened 2021-02-07
memory need: Float64 versus Float32


dbekaert:
Issue ticket for us to assess what precision is really needed for the variables.
e.g. Lon lat should remain float64 for precision purposes at the 10m level.

Need to assess what is needed for the other variables separately:
- P
- T
- q
- e
- hydrostatic refractivity => gets carried in mem for the interpolation object
- wet refractivity => gets carried in mem for the interpolation object
- wet delay
- hydrostatic delay

Below is an example which shows for HRES that currently float64 is used:

```
Traceback (most recent call last):
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/bin/raiderDelay.py", line 4, in <module>
    __import__('pkg_resources').run_script('RAiDER==0.0.1', 'raiderDelay.py')
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/pkg_resources/__init__.py", line 665, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1463, in run_script
    exec(code, namespace, namespace)
  File "/u/leffe0/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/EGG-INFO/scripts/raiderDelay.py", line 12, in <module>
    parseCMD()
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/runProgram.py", line 198, in parseCMD
    _tropo_delay(new_args)
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/runProgram.py", line 219, in _tropo_delay
    (_, _) = tropo_delay(args_copy)
  File "/home/dbekaert/python/python36/envs/RAiDER-NISAR-ST/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/delay.py", line 120, in tropo_delay
    makePlots=True
[...]

PR #253, opened 2021-02-08
Fix two bugs and swap WM read to float32

jlmaurer:
- Fixes a syntax issue
- ds.close() not needed for opendap? 
- Read weather model variables as float32 - a quick fix for now helping with memory load
- Addresses #251 

## How Has This Been Tested?
- ERA-5 runs

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/delayFcns.py
+++ b/tools/RAiDER/delayFcns.py
@@ -95,11 +95,11 @@ def get_delays(
     '''
     # Get the weather model data
     with Dataset(wm_file, mode='r') as f:
-        xs_wm = np.array(f.variables['x'][:])
-        ys_wm = np.array(f.variables['y'][:])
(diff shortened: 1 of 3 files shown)


dbekaert:
@jlmaurer the memory issues are encountered before when computing the hydrostatic and wet delays.
So that would be in the model class itself 

Also, best to verify the precision changes before we make the change.
- Local coordinates in m are like good to be in float 32
- Coordinates in lon lat should stay in float64 as that map to m changes.
- Height i suspect will be fine in float32.

Wrt to the model variables
- the height is derived from the coordinates which is float64. at some point it should be able to convert that to float32. Likely not as part of the calculation. Hence we should do some tests here.

jlmaurer:
Closing this as is dealt with more properly in #256

PR #266, opened 2021-02-19
Address warning + adjust colorbars

sssangha:
## Description
Minor warning when saving INT format files (e.g. heatmaps) was being raised because the nodata value was set to nan. For such INT format files, the nodata value is toggled to 0 to avoid this warning.

Also, if the range of plotted values is too small (e.g. as is usually the case for residual/stdev plots), then the colorbar range must be adjusted to avoid useless colorbars.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/tools/RAiDER/statsPlot.py
+++ b/tools/RAiDER/statsPlot.py
@@ -174,7 +174,7 @@ def convert_SI(val, unit_in, unit_out):
     return val * SI[unit_in] / SI[unit_out]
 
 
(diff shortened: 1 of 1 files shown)


dbekaert (approved):

dbekaert:
LGTM @sssangha

sssangha:
For reference in regards to the latter point, below are example plots -- albeit from different runs -- that demonstrate the problem.

Original forced precision of .2f:
Screen Shot 2021-02-18 at 10 01 23 PM


Dynamic precision adjusted automatically to .3f:
ZTD_grid_delay_stdev

PR #269, opened 2021-02-23
Remove vestigial degrees to m conversion

sssangha:
After @leiyangleon update to convert the coordinates to m in the variogram analysis class, a vestige that crudely converted degrees to m (*110000) got lost in the shuffle.

It has now been removed and range values are on the order of ~a few kms instead of ~a few billions.

Thanks for the catch @bbuzz31 !


jlmaurer (approved):
@sssangha LGTM

PR #278, opened 2021-03-05
Use the correct levels throughout HRES

jlmaurer:

## Checklist:


dbekaert (approved):
L254 is not essential anymore, i believe, what do you think?

There is a check in the code at some point if level and number of a b coefficients is the same.
Might be better to make a check here if the data is the same as what was assumed?

dbekaert:
Beyond the minor comment this looks good to me.

Issue #271, opened 2021-03-01
NCUM in Circle CI


dbekaert:
@kam3545 we would like to add the NCUM model to the circle ci suite. 
It would allow us to test for each code change that the model still works and when it fails allows the team to go in and adress the issue.

For this we would need sample files for NCUM.
We do not need to complete model file, a small subset would be sufficient.

kam3545:
@dbekaert , I am attaching one sample subset file for your reference. 
Last time, Dr. Yang suggested to add filename for pre-downloaded NCUM files.
(_If you’d like to use the pre-downloaded data as I did for the test, you can simply add “filename = ‘.’ + filename” at Line #79 and put the folder “weather_files” one level above RAiDER. That should work temporarily and produce the debugging plots as I sent earlier_.)
This option is not working now.
[testncmr.zip](https://github.com/dbekaert/RAiDER/files/6065319/testncmr.zip)

dbekaert:
@leiyangleon while adding in the scenario 1 test, could you at the same time look at this for NCUM too? 
Similar likely needs a small fix wrt issue #270

leiyangleon:
The error should be fixed by this PR #281.

However, for using a small dataset in CI test, the current NCMR reader does not support it. Because, similar to ERA-5 and etc., NCMR files (in netCDF format) normally cover the whole globe (lat -90-->90 and lon 0-->360). The indices in lat/lon pixels are determined by using the start/end convention and the bounding box user specifies. The small test dataset [above](https://github.com/dbekaert/RAiDER/issues/271#issuecomment-788451119), however, has already been truncated from the whole globe which will mess up the [index calculation](https://github.com/dbekaert/RAiDER/blob/7aa629ea752e01dfa7d8d22e9e8035174db276fb/tools/RAiDER/models/ncmr.py#L80-L90) in the NCMR reader.

FYI, PR #281 guarantees that the NCMR reader works well for the typical NCMR files (the whole globe) NOT a small piece of those files. Similarly, if you provide a small piece of ERA-5 or other model files as input, the reader in the current version does not work either.

leiyangleon:
@kam3545 @dbekaert BTW, the use of the NCMR reader now should be in compliance with other models. Ignore the [temporary method](https://github.com/dbekaert/RAiDER/issues/271#issuecomment-788451119) mentioned before. Just put the pre-downloaded NCMR nc file into `RAiDER/weather_files/` where all other model data are stored, the program will run through without problems.

[...]

Issue #270, opened 2021-03-01
NCUM model filename issue


dbekaert:
See attached screenshot for the reported error by @kam3545 

![Screenshot from 2021-03-01 11-39-46](https://user-images.githubusercontent.com/19240905/109523158-e35ff600-7a63-11eb-86ed-7b57652917e0.png)

dbekaert:
@leiyangleon Could you look into this one while we add unit test for all models in scenario 1?

leiyangleon:
The issue should be fixed by this PR #281.

dbekaert:
Thanks @leiyangleon.
@kam3545 can you give it a try?

PR #281, opened 2021-03-08
Fix ncmr bug

leiyangleon:
## Description

This PR fixes the issues in #270 and partially #271, which reported errors in using the NCMR model reader.

## Type of change

## Checklist:


leiyangleon:
Note that the use of the NCMR reader now should be in compliance with other models. Ignore the [temporary method](https://github.com/dbekaert/RAiDER/issues/271#issuecomment-788451119) mentioned before. Just put the pre-downloaded NCMR nc file into `RAiDER/weather_files/` where all other model data are stored, the program will run through without problems.

jlmaurer (approved):
Seems fine to me

dbekaert (approved):
LGTM @leiyangleon, thank you.

For the unit test, perhaps we could distill the file from @kam3545 down to a small region and store that in our unit test suite?

leiyangleon:
> LGTM @leiyangleon, thank you.
> 
> For the unit test, perhaps we could distill the file from @kam3545 down to a small region and store that in our unit test suite?

Yes, that is what I am trying to do.

Issue #273, opened 2021-03-04
[BUG] second example command (by running the helper command) fails due to losreader problem


leiyangleon:
**Describe the bug**
When running the 2nd raiderDelay helper command, it fails due to array dimension error in the losreader.py file.

**To Reproduce**

1. Command used
`raiderDelay.py --date 20200103 --time 23:00:00 -b 39 40 -79 -78 --model GMAO --zref 15000 --heightlvs 0 100 200 -v
`
2. Error Output
```
Weather model GMAO is available from 2014-02-20 00:00:00-Present
WARNING: Rounded given hour from  23 to 0
Traceback (most recent call last):
  File "/opt/anaconda3/envs/RAiDER/bin/raiderDelay.py", line 4, in <module>
    __import__('pkg_resources').run_script('RAiDER==0.0.1', 'raiderDelay.py')
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/pkg_resources/__init__.py", line 665, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1463, in run_script
    exec(code, namespace, namespace)
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-macosx-10.9-x86_64.egg/EGG-INFO/scripts/raiderDelay.py", line 12, in <module>
    parseCMD()
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-macosx-10.9-x86_64.egg/RAiDER/runProgram.py", line 196, in parseCMD
    _tropo_delay(new_args)
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-macosx-10.9-x86_64.egg/RAiDER/runProgram.py", line 208, in _tropo_delay
    (_, _) = tropo_delay(args_copy)
  File "/opt/anaconda3/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-macosx-10.9-x86_64.egg/RAiDER/delay.py", line 153, in tropo_delay
    los = getLookVectors(los, lats, lons, hgts, zref)
[...]

PR #272, opened 2021-03-04
Merra2 pydap update

leiyangleon:
## Description

MERRA2 recently updated their pydap-retrieved data format, so these fixes are required. The GMAO pydap-retrieved format was not changed.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have added an explanation of what your changes do and why you'd like us to include them.
- [ ] I have written new tests for your core changes, as applicable.
- [x] I have successfully ran tests with your changes locally.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #285, opened 2021-03-10
Add all models scenario 1

leiyangleon:
## Description

This PR adds all other weather models besides ERA5 for test scenario 1. For each model, there is a sub-folder named by that model, i.e. RAiDER/test/scenario_1/GMAO/, containing all the wet and hydrostatic results in both envi and header format. The test_scenario_1.py file has been updated with a number of sub-tests for all of the models: 6 for now and NCMR is to be included soon.

## Type of change

## Checklist:


leiyangleon:
@jlmaurer @dbekaert the CI check passed for my version of RAiDER (accessed on March 2) but not the current one. Would thus recommend wait for @jlmaurer to fix the existing bugs like #273 . If still not passing, then I can debug the unit test code to see what the cause is.

leiyangleon:
@jlmaurer @dbekaert here is probably the issue that caused this CI check. In @jlmaurer 's PR #274, one additional key `pnts_file` was added [here](https://github.com/dbekaert/RAiDER/pull/274/files#diff-35207f7e60ea2f00a8cac0e26cda9bcf7772bbad46aae89c69924ebfa05f440aR47). Will add that line and CI will probably pass this time.

leiyangleon:
@dbekaert @jlmaurer the CI passed for GMAO, ERA5, ERA5T, while failed for MERRA2, ERAI, HRES. This is because:

1. MERRA2 server is down today; @dbekaert be sure to re-run CI check sometime later. it should pass as I did yesterday 
2. ERAI and HRES passed on my laptop without problem; the problem here is associated with the anonymous account of ecmwapi that is being used in the CI check, which does not have access to these two models. @dbekaert be sure to agree the terms and conditions for that dummy user account and then re-run CI check, which should pass without problems either

jlmaurer:
@leiyangleon and @dbekaert this is a good point about account access. Probably we should add a flag on the tests that require account access so that they skip if the credential file is not found.

dbekaert:
Agreed on what @jlmaurer suggested on skipping the test if the account is not found. However it will not be say to do so. e.g. One could for example have ECMWF account but not an HRES license, but ERA-I would work. Can we capture the return from the API to see if one is unauthorized user? 

wrt circle CI same applies the anonymous user will never have HRES, that specific to an individual. if the unit test is small we could capture the credential in secured circle ci as environment variables.

dbekaert:
@leiyangleon  The environment variable are set in circleCI. See the details under circle ci above under environment variables tab. I suspect we are not using using the environment variable correct when trying to make the ~/.ecmwapi and ~/.cdsapi files. Could you perhaps look into this and perhaps try to modify .circleci/config.yml file?

leiyangleon:
@dbekaert @jlmaurer see below the ecmwapi account-related error messages for ERA-I and HRES:

- HRES: `ecmwfapi.api.APIException: "ecmwf.API error 1: User 'anonymous' has no access to services/mars"`
- ERA-I: `ERROR: "ecmwf.API error 1: User 'anonymous' has not access to datasets/interim. Please accept the terms and conditions at http://apps.ecmwf.int/datasets/licences/general"`

Can either of you accept the terms/conditions for the anonymous user account so that ERA-I would work? 

In the meanwhile, I will see if I can skip HRES if no access.

BTW, MERRA-2 server is back on but the pydap data retrieval format changed back to the original version (that reverses my recent PR #272). So, I would need to issue another PR to change it back. What I incline to do in this PR is to keep the two versions of the data retrieval format (by commenting the unused one) as it seems agnostic when the server changes to either one of the two versions.

dbekaert:
Per earlier comment, the .circle is setting the account details and
the .ecmwfapi and .cdapi files.
The account does not have problems getting these files.

I suspect it is going wrong as its not listing it as a user by anonymous.
Perhaps it is an issue on location in the ~/.ecmwfapi for the circle ci run?

leiyangleon:
> Per earlier comment, the .circle is setting the account details and the .ecmwfapi and .cdapi files. The account does not have problems getting these files. I suspect it is going wrong as its not listing it as a user by anonymous. Perhaps it is an issue on location in the ~/.ecmwfapi for the circle ci run?
> […](#)

Yes, I saw the definition of the .ecmwfapi file in CircleCI, but that only defines the account information (url, email, key). My understanding is when the account is created, the user who creates that account has to go to some website and agree the terms/conditions so that specific data can be downloaded. based on the ci error message I attached above, i.e. also copied here [`ERROR: "ecmwf.API error 1: User 'anonymous' has not access to datasets/interim. Please accept the terms and conditions at http://apps.ecmwf.int/datasets/licences/general"`], it is apparent that agreement step is missing, which  is not related to ci definition. As far as I know, one cannot agree the terms/conditions in the ci. But one can do it on the website once and for all.

dbekaert:
In both cases the account can download the data on the local servers.
Which i just verified manually.

```
(RAiDER-NISAR-ST) ***@***.*** test]$ raiderDelay.py --time 00:00
--model ERAI --date 20090101 --bbox -90 90 -180 180
Weather model ERA-I is available from 1979-01-01 00:00:00-2019-08-31
00:00:00
2021-03-15 16:42:44 ECMWF API python library 1.6.1
2021-03-15 16:42:44 ECMWF API at https://api.ecmwf.int/v1
2021-03-15 16:42:45 Welcome David Bekaert
```

I do not think the accounts are being picked up.
The HRES account being fed through the environment variables is the one we
can use for downloading actual HRES data.
I verified the key and account just now is the same on my local server.
If the .ecmwfapi file would be picked up correctly that would have reported
the account details.

leiyangleon:
@dbekaert @jlmaurer after fixing the ci file, both ERA-I and HRES worked fine this time. Please approve my new PR #290 that fixes the MERRA-2 bug and then run this CI check again, which will make this PR's CI check pass.

dbekaert:
Re-ran it but error stays. Let me know what i should do next.
I have merged your PR for the merra that passed in the other run.
You might need to rebase the current PT  to the dev branch?

leiyangleon:
@dbekaert @jlmaurer I have rebased the branch to the dev branch but MERRA2 still won't work in the CI check. Spent a whole day debugging but still no luck. Looking at the CI error messages, it is just the pydap data array slicing that gives errors, i.e. something like dataset[slice], which points to some Conda-installed pydap files. The weird thing is I have been successfully running the same test command on my laptop and I believe it is not an api user account/password issue either. Can you both try downloading this particular branch to your laptop and running "py.test test/test_scenario_1.py" to verify if that works for you guys? At this stage, I cannot see what caused that difference. Any other thoughts?

BTW, if you don't know where the problem comes from for the CI check and also the test command passes on your laptops, I would recommend to comment out MERRA2 scenario 1 test function for now and merge the PR because 
1. all other models worked fine
2. the problem only occurs for CI check, i.e. it runs perfectly on local machines
3. MERRA2 server is not stable as it sometimes shuts down and changes the pydap data retrieval format
4. MERRA2 is not on the list for the NISAR tropo working group

Let me know what you think.

dbekaert:
@leiyangleon fine by me.
@jlmaurer thoughts?

jlmaurer:
Hi @leiyangleon and @dbekaert I found the bugs. The problem has to do with the PyDAP packaging, sometimes you need a ".data" attribute, and sometimes ".data[0]", and sometimes nothing. I added a few try/excepts with explicit catches and it solves the problem. 

The failure was happening because the data download would fail but it wasn't erroring out because the error message got sent to logging and did not re-raise the exception. That's partly why it wasn't clear initially what was actually causing the bug. This can be fixed by explicitly re-raising any unhandled errors, which I implemented in the new branch. 

Incidentally I also fixed the HRES bounds issue, happens because the longitude check I had implemented wasn't correct. Everything passes now fine, except that there are some small differences in the GMAO and HRES results. I looked at these, and they are only a few pixels and mostly of order 10^-3.

jlmaurer:
I added a new PR #294 with my updates.

leiyangleon:
@jlmaurer @dbekaert sounds great! Will look into this next week and verify the updated PR.

leiyangleon:
@jlmaurer @dbekaert I tried the changes @jlmaurer implemented for MERRA-2, which still failed in the CI check. Note that @jlmaurer added one `except` condition with .data format, which I have never encountered for MERRA2. Also note that GMAO and MERRA2 always hhave different pydap formatting, plus GMAO is working fine, i.e. only MERRA2 is the problem here. See below the error messages. Again the root error is exactly the same as [what I found](https://github.com/dbekaert/RAiDER/pull/285#issuecomment-800689509) (marked in bold), which is not related to the saved files nor the the .data format. 

Back to my above suggestion, if @jlmaurer agrees, for now, I would like to comment out MERRA-2 (also incorporate @jlmaurer's changes on HRES) and let the PR pass. Other people could always pick up the MERRA-2 piece and continue to refine it later. Otherwise, I don't have enough time to wrap up the work (incorporating NCMR which is requested my the NISAR working group) by March 31st. Let me know what you guys think.

ERROR: MERRA-2: Unable to read weathermodel data
**Traceback (most recent call last):
  File "/root/miniconda/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/models/merra2.py", line 113, in _fetch
    q = ds['QV'].data[0][time_ind, ml_min:(ml_max + 1), lat_min_ind:(lat_max_ind + 1), lon_min_ind:(lon_max_ind + 1)][0]
  File "/root/miniconda/envs/RAiDER/lib/python3.7/site-packages/pydap/handlers/dap.py", line 139, in __getitem__
    dds, data = r.body.split(b'\nData:\n', 1)
ValueError: not enough values to unpack (expected 2, got 1)**
ERROR: MERRA-2: Unable to save weathermodel to file
  File "/root/miniconda/envs/RAiDER/lib/python3.7/site-packages/RAiDER-0.0.1-py3.7-linux-x86_64.egg/RAiDER/models/merra2.py", line 133, in _fetch
    writeWeatherVars2NETCDF4(self, lats, lons, h, q, p, t, outName=out)
UnboundLocalError: local variable 'h' referenced before assignment

jlmaurer (approved):
Approving this for now, will re-visit MERRA-2 later.

PR #292, opened 2021-03-16
3D DEM intersection for ZTD

### LINKED ISSUE #81: Implement option for convetional slant delay computation
This option would be calculation of slant delay by first computing the zenith delays and projection towards the slant.

jlmaurer:
This fairly significant PR implements the DEM intersection method for calculating delays for ZTD

This change introduces a change in how the delays get calculated; i.e. there are no rays computed for ZTD. This results in several differences with the existing scenario tests, described below. 

## Description

- Refactored and simplified code to handle 3D point intersection going off of ARIA-tools implementation (just a RegularGridInterpolator operation)
- Added a "Conventional" class definition to complement the existing "Zenith" class, for calculating slant delays using the conventional approach. This is now the default behavior when passing a LOS file. 

## Motivation and Context

This puts in place the 3D intersection of the query points with the native weather model grid for calculating ZTD. Eventually this will be implemented for LOS as well.
- Also fixes #81 

## How Has This Been Tested?

- All unit tests except the scenarios pass locally
- Autopep8 ran on changes

- test_delay module added for testing functions in the delay.py module. 

This update breaks current unit tests test_scenario_1 and test_scenario_2. 
- hydrostatic delays are ~5% different and wet delays are ~0.3% different for both tests
- There are fewer NaN-values in the new mode than the original. Presumably this is because there is no DEM and all the weather model points are used, even those below ground.

## Type of change

## Checklist:


jlmaurer:
@dbekaert and @bbuzz31 this PR should be ready to review / merge. It calculates Zenith delays on the 3D cube and does the 3D DEM intersection as per @sssangha's ARIA-tools implementation. I also added a conventional slant delay calculation, which is the default now when passing an ISCE-style line-of-sight file (instead of an orbit vector).

jlmaurer:
@dbekaert I have a copy of this PR with only a single commit if you'd rather do it that way.

bbuzz31 (approved):
I went through the changes from a high level and ran the zenith and conventional slant delays over a subset of the EC. LGTM.

PR #294, opened 2021-03-17
Add all models scenario 1 update

jlmaurer:
This fixes some bugs in the original PR #285 

## Description

This PR adds all other weather models besides ERA5 for test scenario 1. For each model, there is a sub-folder named by that model, i.e. RAiDER/test/scenario_1/GMAO/, containing all the wet and hydrostatic results in both envi and header format. The test_scenario_1.py file has been updated with a number of sub-tests for all of the models: 6 for now and NCMR is to be included soon.

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/models/gmao.py
+++ b/tools/RAiDER/models/gmao.py
@@ -140,9 +140,13 @@ class GMAO(WeatherModel):
 
         try:
             # Note that lat/lon gets written twice for GMAO because they are the same as y/x
+            writeWeatherVars2NETCDF4(self, lats, lons, h.data, q.data, p.data, t.data, outName=out)
+        except AttributeError:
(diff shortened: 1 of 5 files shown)


jlmaurer:
@leiyangleon @dbekaert this PR will fail because of small differences in the "true" files originally saved by @leiyangleon . My recommendation is probably to just update these files, because the differences are small and possibly due to different files actually being used to generate them (which is why I suspect @leiyangleon the tests were passing on your machine).

dbekaert:
@jlmaurer suggest to update the files before we merge. We have a lot of PR so want to make sure we use circleCI

jlmaurer:
@dbekaert agree, I suggest @leiyangleon perhaps you can regenerate the files as needed and make sure they are consistent with what you had done before. Perhaps you can just update your PR with the changes

jlmaurer:
Closing this as no longer needed.

Issue #277, opened 2021-03-05
Clean up bounds reporting (panda's function is shown)


dbekaert:

jlmaurer:
Just need to change the formatting

cmarshak:
I want to confirm I was asked to also use gdal/rasterio to avoid opening the entire netcdf array in memory.

dbekaert:
Does not need to be GDAL, if there is existing array functionality to get the bounds rapidly.
Note that variable names in the datasets will change depending on the model.

jlmaurer:
I think this can be closed now.

PR #295, opened 2021-03-17
Issue 277

cmarshak:
## Description

- Use rasterio to check if bounds are contained within a weather model extent
-  Log the extents
- Update related tests
- Flake8 of modified files
- Use pathlib in tests

https://github.com/dbekaert/RAiDER/issues/277

## Motivation and Context

To speed up bound checking for weather models.

## How Has This Been Tested?

The relevant test (which has been modified) is `test/weather_model/test_processWM.py`. This was tested with `pytest test --ignore test/weather_model/test_downloaders.py`

## Type of change

## Checklist:


tools/RAiDER/processWM.py
```diff
@@ -140,16 +136,33 @@ def prepareWeatherModel(
...
+def checkContainment(weather_model: weatherModel,
+                     outLats: np.ndarray,
+                     outLons: np.ndarray) -> bool:
+    '''Check if weather model contains bounding box of outLats and outLons'''
+    weather_model_path = weather_model.files[0]
+    with rasterio.open(f'netcdf:{weather_model_path}') as ds:
+        datasets = ds.subdatasets
+
+    with rasterio.open(datasets[0]) as ds:
+        bounds = ds.bounds
```
cmarshak, line 148: @dbekaert - I assume the netcdf has a bunch of arrays of similar shape as in they conform to the xarray dataset. I can see this would fail for the ARIA-type netcdfs if there are groups with different shapes/coordinates as well as a shapefile embedded there.

dbekaert: I guess, you are assuming that any first dataset hit would be the right size.

cmarshak: Yes, exactly. If you would like a different approach, fine to do that. Let me know.

dbekaert:
@jlmaurer likely on way to make sure this does not break anything is to run the scenario_1 two times, no? As the code will only enter the bounds checking when the file exists. This would directly capture the assumption of having same dimensions.

tools/RAiDER/processWM.py
dbekaert: Depending how we would go about xarray in future, we could capture this dataset and variable mapping.

dbekaert:
@jlmaurer What is your though on running scenario 1 twice, above above comment?

jlmaurer:
@dbekaert that could work. An alternative would be to write a test file directly and read its bounds; this would be more direct as we could test various variable names etc. without having to run multiple scenario_* tests. I'm going to check quickly to see if it would be easy to implement a single direct test

README.md
```diff
@@ -119,19 +125,23 @@ conda update --force-reinstall gdal libgdal
 To test the installation was successfull you can run the following tests:
 ```
 py.test test/
-raiderDelay.py -h 
+raiderDelay.py -h
 ```
 
+### To enable automatic CircleCI Tests from a pull requests
```
jlmaurer, line 131: Thanks for adding this!

environment-dev.yml
```diff
@@ -34,6 +34,7 @@ dependencies:
  - ecmwf-api-client
  - cxx-compiler
  - sysroot_linux-64
+ - rasterio
```
jlmaurer, line 37: I have a suggested version of this that won't require rasterio but will require numba. 
```suggestion
 - numba
```

test/__init__.py
```diff
@@ -1,17 +1,8 @@
-import inspect
 import os
 from contextlib import contextmanager
+from pathlib import Path
```
jlmaurer, line 3: Thanks for adding this fix as well! I incorporated this into a different PR, so it may or may not show up if that one does/does not get added first, but I used your code here to do it.

tools/RAiDER/processWM.py
```diff
@@ -6,18 +6,14 @@
...
-import sys
-
 import numpy as np
-import matplotlib.pyplot as plt
-import xarray as xr
-
-from datetime import datetime, date
-
-from RAiDER.logger import *
+import rasterio
```
jlmaurer, line 11: Probably we should keep the original syntax (spacing) here to be consistent with the other modules, and update them all together if desired.

tools/RAiDER/processWM.py
```diff
@@ -140,16 +136,33 @@ def prepareWeatherModel(
...
-    in_extent = weather_model._getExtent(outLats, outLons)
+def checkContainment(weather_model: weatherModel,
+                     outLats: np.ndarray,
+                     outLons: np.ndarray) -> bool:
+    '''Check if weather model contains bounding box of outLats and outLons'''
+    weather_model_path = weather_model.files[0]
+    with rasterio.open(f'netcdf:{weather_model_path}') as ds:
+        datasets = ds.subdatasets
+
+    with rasterio.open(datasets[0]) as ds:
```
jlmaurer, line 147: I definitely think this could fail. This check happens to the original weather model data (as opposed to the processed) so we don't know what variables exist and whether they are always the first. My opinion would be to use either gdal, rasterio, or xarray to query the coordinates specifically (e.g. the "coords" attribute in xarray). I will post a time test results later.

tools/RAiDER/processWM.py
```diff
@@ -140,16 +136,33 @@ def prepareWeatherModel(
...
+
+    xmin, ymin, xmax, ymax = tuple(bounds)
+    weather_model_box = box(xmin, ymin, xmax, ymax)
+
+    xmin_input, xmax_input = np.min(outLons), np.max(outLons)
+    ymin_input, ymax_input = np.min(outLats), np.max(outLats)
+    input_box = box(xmin_input, ymin_input, xmax_input, ymax_input)
+
+    # Logger
+    input_box_str = [f'{x:1.2f}' for x in [xmin_input, ymin_input,
```
jlmaurer, line 158: I think we want to avoid logging inside subfunctions, as we don't necessarily want to log when running unit tests etc. You could move this part into the main "prepareWeatherModel" function.

tools/RAiDER/processWM.py
cmarshak: I just want to clarify a few points about when/how this will fail. 

a) This only selects _geo-referenced_ variables in an netcdf files. For example, it would ignore "levels" in an HRES processed weather model file
b) If there are multiple subgroups with different coordinate reference frames then this may fail if the reference frames have different bounding box - in this case you couldn't open the entire dataset with xarray; you would have to specify the group.

There is a third instance (coming from aria) in which a non-raster, georeferenced object is included in the netcdf. I don't think this will be a problem for the data we are considering.

tools/RAiDER/processWM.py
cmarshak: Is there a static analysis tool I should be using? I can just turn off flake8 too.

tools/RAiDER/processWM.py
jlmaurer: @cmarshak ah I see! I was essentially thinking about a levels or similar dataset coming from the weather model file, but evidently that is not an issue. I think this could work fine then.

tools/RAiDER/processWM.py
jlmaurer: There is an autopep8 cmd we've been running (see the template PR), otherwise not.

tools/RAiDER/processWM.py
cmarshak: Could we remove the logging?

jlmaurer: We definitely want to log the bounds, as that is the first thing you look at when there is a problem with the extents. I originally had the function return the extents, rather than return the flag.

jlmaurer (approved):
@cmarshak I like it!

PR #296, opened 2021-03-22
Add helper cmd-line program for just combining files

jlmaurer:
Running ```raiderCombine.py --raiderDir '*/' --raider all_delays.csv``` combines all the delay files together (in this case, without matching to a GNSS delay file)

## Description

- Filenames still need to have the datetime string "YYYYMMDDTHHMMSS" in their name, which all RAiDER-generated files do. 
- Updated the search for the datetime in the filename to be a regex instead of integer indexing
- Made the GNSS file optional so that you can combine RAiDER files without having a GNSS file
- Any set of files you can find by globbing are discoverable
- The output RAiDER delay file is sorted by first station ID then by Datetime

- all unit tests pass locally
- Added unit tests to test/test_gnss.py

## Type of change

## Checklist:


jlmaurer:
@dbekaert and @sssangha this is a slightly modified version of the code that allows for combing RAiDER delay files without needing to have a GNSS file for comparison. You can simply omit the --gnss option and combine the files. Also note that it should work as expected regardless of datetime / region; i.e. it will take what is there and concatenate, and I was already dropping dups so did not need to add that.

jlmaurer:
Just realized that this is still taking the inner merge. Just as a reminder: 
- Inner only keep datetime/ID combos that have both GNSS and RAiDER delays
- left keeps (in this case) all raider delays and matches up GNSS delays as available
So we want to switch to left merge?

dbekaert:
@jlmaurer; yes left merge is probably the way to go. But would be good to report out a stat per year on how much of the GNSS is missing.

tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -143,12 +162,20 @@ def create_parser():
...
     )
 
+    p.add_argument(
+        '--gnss', dest='gnss_file',
+        help=dedent("""\
+            Optional .csv file containing GPS Zenith Delays. Should contain columns "ID", "ZTD", and "Datetime"
+            """),
+        default=None
+    )
+
```
dbekaert, line 172: Can you add run file examples to the helper?

jlmaurer: @dbekaert done.

jlmaurer: There was already an example command. I added another to cover the case of no gnss files.

jlmaurer:
@dbekaert ok I changed the merge style to left. Not sure how to easily check how many GNSS stations are missing without some rather clunky code; related to outer merge but not sure how to only get the right side.

dbekaert:
thanks @jlmaurer

Issue #284, opened 2021-03-09
STATS: local time globally


dbekaert:
Add the ability to make a merged csv file for local time. Have a -+ threshold hour. This does not need to be part of stats, could be in raider combine or so.

PR #298, opened 2021-03-29
Pass local time support

sssangha:
Introduce local-time support, and implement a series of other related fixed.

## Description
Through raiderCombine.py`, introduce optional control to pass only data at local-time WRT user-defined time at 0 longitude (1st argument), and within +/- specified hour threshold (2nd argument). By default UTC is passed as is without local-time conversions. Input in 'HH:MM:SS H', e.g. '16:00:00 1'"

Also I made two other fixes:
1. Bug-fix in `raiderDownloadGNSS.py` relating to floating undefined variable in the case a user doesn’t specify a bounding box (i.e. a global query is initiated). 
2. Lines with NaNs and duplicate lines now dropped as merging is prepped in `raiderCombine.py`. This drastically improves the run-time up to ~20 mins depending on the size of the files.
3. Bug-fix in `raiderStats.py` whereby warning is passed to log (as opposed to raising exception) if the stdev column isn’t found.
4. The user may now specify a directory containing GNSS CSVs`-gnssdir` to concatenate through the `raiderCombine.py` program.

## Type of change

## Checklist:


sssangha:
To elaborate, the logic in the code is as so. 0 longitude is treated as a reference whereby the specified local time and threshold WRT 0 longitude are projected through to the rest of the globe.
This is accomplished by scaling the rotation rate at the equator to the longitude of interest, estimating the great circle distance along a consistent latitude centered at the 0 longitude to a given longitude of interest, and dividing the great circle distance by the rotation rate at that given longitude in order to estimate the time-offset WRT 0 longitude, given the specified UTC time (1st argument for `--localtime`). This time-offset is then used to estimate the expected local time at a given location, and given the specified time threshold in hours (2nd argument for `--localtime`) it is determined whether or not the native UTC time associated with the data is the same as the expected local time, and if it isn't then the datapoint is discarded.

Tests below, using data acquired over a single day for sake of speed:

Original ERA5-GNSS residual at 06 UTC (i.e. no local-time projection):
Screen Shot 2021-03-29 at 3 44 54 PM


Only points at 06 local UTC between the single input ERA5 and GNSS files are passed, we clearly need to concatenate more observations to not be limited over small windows of longitude:
local UTC 06
Screen Shot 2021-03-29 at 6 48 08 PM



GNSS observations approximately every 2-3 hours are concatenated, and merged with ERA5 observations at 06 and 18 UTC. While still only covering a single day, this denser coverage between the two sets of observations ensures precise global coverage at the specified local-time:
local UTC 12
Screen Shot 2021-03-29 at 6 45 39 PM

sssangha:
Addresses #284

tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -104,15 +187,61 @@ def mergeDelayFiles(
...
+    dfz = pd.read_csv(ztdFile, parse_dates=['Datetime'])
+    # only pass common locations and times
+    dfz = pass_common_obs(dfr, dfz)
+    dfr = pass_common_obs(dfz, dfr)
+
+    # If specified, convert to local-time reference frame WRT 0 longitude
+    if localTime is not None:
+        from RAiDER.getStationDelays import seconds_of_day
+
+        localTime_hrs = seconds_of_day(localTime.split(' ')[0])/3600
```
jlmaurer, line 207: Could this be functionalized to make unit testing easier?

tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -224,6 +367,20 @@ def create_parser():
         default='Combined_delays.csv'
     )
 
+    p.add_argument(
+        '--localtime',
```
jlmaurer, line 371: The explanation was a little opaque to me. Rather than specify the details about 0 longitude, perhaps just state that the user can pass a localtime?

jlmaurer (approved):
Have those few comments, otherwise looks fine!

tools/RAiDER/gnss/processDelayFiles.py
sssangha: If I'm understanding correctly, are you suggesting to segregate out this local-time functionality into a separate function?

tools/RAiDER/gnss/processDelayFiles.py
sssangha: Yeah, maybe a bit too superfluous in the parser. How about something like this instead:
"Optional control to pass data at specified, locally-projected time (1st argument) and pass data within specified +/- X hour threshold of local-time (2nd argument).
By default UTC is passed as is without local-time conversions.
Input in 'HH:MM:SS H', e.g. '16:00:00 1'""

sssangha:
Logic behind local-time estimation:
Given a rotation rate of (360deg/23.9333333333hr) = 15.041782729825965 deg/hr, estimate the local time-shift at a given longitude WRT 00UTC at 0 longitude. 

This time-shift is then applied to a given base-time derived from the native datetime of a given observation, but set to the user-specified local hour. E.g. if a user specifies 00 UTC and a given observation’s native date time is 2019-10-01 01 UTC at ~15 longitude, the time-shift WRT 0 longitude is ~1hr (because 15 longitude divided by the rotation rate of 15.041782729825965 deg/hr = 1 hr), then this observation is selected as it is within the expected time bounds (i.e. user-input hour-threshold +/- local-time ). 

Also, for observations sufficiently far west of 0 longitude (i.e. negative time-shift), it may be necessary to shift the base-time by 1 day forward. E.g. if a user specifies 00 UTC and a given observation’s native datetime is 2019-09-30 18 UTC at ~-100 longitude, the time-shift WRT 0 longitude is ~-6hrs (100 longitude divided by rotation rate of 15.041782729825965 deg/hr = 6.6hrs), so we need to add one day to this base-time (i.e. 2019-10-01) and subtract ~6hrs to accurately pick out the appropriate observations.

To clarify, the function ‘update_time’ produces estimated local-times WRT user-input parameters (again, specified UTC local time and time threshold) which are then cross-referenced with the native datetimes of a given observation so that observations outside of the estimated temporal range are discarded.


I have tested this out on observations spanning January 2019 - March 2019 for ERA vs GNSS residuals. Note that this was done using concatenated ERA observations sampled every 3 hours and concentrated GNSS observations sampled every ~2-3 hours. The narrow distribution of results suggest that perhaps it would be a good idea to download hourly, global GNSS observations and/or expand the time threshold beyond +/-1 as I’ve specified.
For each panel, the mean residual for a given station is on the left, and the native datetime hour is plotted on the right:

00 UTC:
Screen Shot 2021-04-26 at 10 49 02 PM

06 UTC:
Screen Shot 2021-04-26 at 10 51 07 PM


16 UTC:
Screen Shot 2021-04-26 at 10 53 14 PM


18 UTC:
Screen Shot 2021-04-26 at 10 56 18 PM


Raider bug-fixes:
Don’t automatically adjust colorbar if range is specified by the user, even if the result will be saturated.


Enhancements to piggy-back off of this PR:
Plot RMSE of seasonal amplitude/phase curve-fit in addition to existing seasonal plots. This is now automatically generated when the user specifies seasonal fitting, no additional option needs to be toggled.

tools/RAiDER/gnss/processDelayFiles.py
sssangha: @jlmaurer done, see `local_time_filter` function

dbekaert:
@sssangha Can you add the cycle colorbars for the differnt plots as well as the one for the phase in here, such we will merge and make features available to @kam3545 in order to update his figures to be consistent with ours.

PR #305, opened 2021-04-16
Hres pl

jlmaurer:
Adds pressure levels to the HRES API

## Description

- moves a/b/zlevels to an external file
- adds fixed height levels
- For some reason the API doesn't like arbitrary floats in the min bounding box? 

## Type of change

## Checklist:


tools/RAiDER/models/era5.py
```diff
@@ -27,59 +28,9 @@ def __init__(self):
...
-            0.111505, 0.124448, 0.138313, 0.153125, 0.168910, 0.185689, 0.203491,
-            0.222333, 0.242244, 0.263242, 0.285354, 0.308598, 0.332939, 0.358254,
-            0.384363, 0.411125, 0.438391, 0.466003, 0.493800, 0.521619, 0.549301,
-            0.576692, 0.603648, 0.630036, 0.655736, 0.680643, 0.704669, 0.727739,
-            0.749797, 0.770798, 0.790717, 0.809536, 0.827256, 0.843881, 0.859432,
-            0.873929, 0.887408, 0.899900, 0.911448, 0.922096, 0.931881, 0.940860,
-            0.949064, 0.956550, 0.963352, 0.969513, 0.975078, 0.980072, 0.984542,
-            0.988500, 0.991984, 0.995003, 0.997630, 1.000000
-        ]
+        self._a = A_ERA5
```
dbekaert, line 31: Catching this one for you ;)

tools/RAiDER/models/era5.py
```diff
@@ -27,59 +28,9 @@ def __init__(self):
...
+        self._b = B_137_ERA5
+        self._zlevels = LEVELS_137_HEIGHTS
```
dbekaert, line 33: missing import

tools/RAiDER/models/hres.py
```diff
@@ -24,14 +32,12 @@ def __init__(self):
...
 
         # 9 km horizontal grid spacing. This is only used for extending the download-buffer, i.e. not in subsequent processing.
-        self._lon_res = 9. / 111
-        self._lat_res = 9. / 111
-        self._x_res = 9. / 111
-        self._y_res = 9. / 111
+        self._lon_res = 0.08108115
+        self._lat_res = 0.08108115
+        self._x_res = 0.08108115
+        self._y_res = 0.08108115
```
dbekaert, line 38: Does this impact us when we have already download all models with 9/111?

tools/RAiDER/models/hres.py
```diff
@@ -265,27 +224,116 @@ def _fetch(self, lats, lons, time, out, Nextra=2):
...
+                'stream': "oper",
+                'type': "an",
+                'levelist': "all",
+                'levtype': "{}".format(self._model_level_type),
+                'param': param,
+                'date': datetime.datetime.strftime(corrected_date, "%Y-%m-%d"),
+                'time': "{}".format(datetime.time.strftime(corrected_date.time(), '%H:%M')),
+                'step': "0",
+                'grid': "{}/{}".format(lon_step, lat_step),
+                'area': "{}/{}/{}/{}".format(lat_max, floorish(lon_min, 0.1), floorish(lat_min, 0.1), lon_max),
```
dbekaert, line 256: If you floor should the max be ceiled?

tools/RAiDER/models/hres.py
jlmaurer: That is a good point; yes I suspect it will. We can wait on implementing this part until after done processing.

tools/RAiDER/models/hres.py
jlmaurer: I don't think it's needed here; the only reason I floor is because I was getting API errors when allowing arbitrary floats. The maximum will already be a fixed integer multiple of the grid size, so ceil isn't needed as we already have a buffer.

tools/RAiDER/models/hres.py
jlmaurer: Just a note that the float number is the value I get as the native grid resolution, which is very slightly different than the division result (0.0810810810...).

jlmaurer:
@dbekaert this should be good to go now!

dbekaert (approved):
LGTM @jlmaurer  thnx

PR #307, opened 2021-05-27
Output/plot variogram rmse

sssangha:
## Description
Save and plot rmse for variogram fit. Before this commit, this information was only passed around in memory but not saved. Other than passing the `-variogramplot` argument, the user doesn't have to specify anything else to generate this.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)


jlmaurer (approved):
Seems fine to me!

PR #312, opened 2021-07-08
Fix losreader

jlmaurer:
This update ensures that the statevector code for generating unit look vectors is working properly. 

## Description

- Update losreader.py: clean up, verify outputs, add documentation
- Update the C-code of Geo2rdr: remove print statements, clean up, add documentation
- Update utilFcns.py: add functions, unit tests
- When using an orbit file, the code automatically passes the user input time through the code to trim the orbit file 
- Adds a flag for catching bad matches between the input time, orbit file, and lat/lon points requested

## Motivation and Context

- Wasn't sure before that the state_to_los code was working. We have checked and verified that it is
- Fixes #122
- Fixes #79 

## How Has This Been Tested?

- all unit tests pass locally
- added unit tests to cover new functionality

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have added an explanation of what your changes do and why you'd like us to include them.
- [x] I have written new tests for your core changes, as applicable.
- [x] I have successfully ran tests with your changes locally.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #314, opened 2021-07-20
Fix losreader new

bbuzz31:
This update makes [nearly] consistent the LOS calculation from statevector with the calculation from  heading/azimuth angle ISCE raster. 
See #312 for more details.


bbuzz31:
Here are comparisons of ENU unit vectors from heading/azimuth  calculation (left column) and state vector calculation (right)
image

And the residuals:
image

Ascending track 4.

bbuzz31:
moved to #316

PR #316, opened 2021-08-03
Fix los reader new 1

bbuzz31:
see #314


jlmaurer:
Closing this as it is replaced by #317

Issue #293, opened 2021-03-16
[BUG] Heights are not currently being consistently handled throughout the code


jlmaurer:
- There are ellipsoidal and geoidal heights currently being used without any conversions between
- Inputs are likely to always be ellipsoidal, while weather models are typically related to the geoid
- Will difference be significant enough to fix given the range of variations between the geoid and WGS-84 ellipsoid?

jlmaurer:
Need to check what the vertical datum is for each weather model
ECMWF - EGM96

dbekaert:
So two items here:

1) weather models come with its own native projection (horizontal proj). Both pressure and model levels compute geopotential which would be respective to the geoid. The geoid depends on what is assumed by the model.

"ECMWF data is referenced in the horizontal with respect to the WGS84 ellipse (which defines the major/minor axes) but in the vertical it is referenced to the Geoid (EGM96)." see [here](https://confluence.ecmwf.int/display/CKB/ERA-Interim%3A+documentation)

We should add in the vertical datum as well using the proj conventions that are compatible with GDAL.

gdalwarp -s_srs "+proj=longlat +datum=WGS84 +no_defs +geoidgrids=egm96_15.gtx" -t_srs "+proj=longlat +datum=WGS84 +no_def" dem_in_egm96_geoid.tif dem_in_wgs84_ellipsoid.tif

2) we convert the weather model vertical sampling into the same coordinate system as the DEM. We need to make sure that the DEM projection and vertical datum are propagated 

3) Example on converting the EGM96 to WGS84 vertical datum:
- raters with corrections [here](https://www.agisoft.com/downloads/geoids/)
- gdalwarp -s_srs "+proj=longlat +datum=WGS84 +no_defs +geoidgrids=egm96_15.gtx" -t_srs "+proj=longlat +datum=WGS84 +no_def" dem_in_egm96_geoid.tif dem_in_wgs84_ellipsoid.tif

dbekaert:
@cmarshak could you cover this once back? This is synergistic with the ARIA work.

cmarshak:
Ok - we can figure out what's best for this repo.

I worked with David on this for ARIA/ACCESS and it is still very much a work in progress (or maybe it will just be cut/paste for these types of issues): https://github.com/aria-jpl/dem_stitcher

There are things that can be adapted here - likely these two pieces:

[...]

PR #318, opened 2021-08-12
Use DEM-stitcher

cmarshak:
## Description
See: https://github.com/dbekaert/RAiDER/issues/293

1. Use [GLO-30 DEM](https://registry.opendata.aws/copernicus-dem/) from open s3 bucket. 
2. Ensure that DEM is in area centered-coordinates. Note that GLO-30 is distributed with "pixel as point" (see this [discussion](https://github.com/OSGeo/gdal/issues/1505#issuecomment-489469904)) - can verify inspecting original tile gdal tags.
3. Convert to Ellipsoidal height

Some cons:

1. Basically, utilize ARIA's/my [repo](https://github.com/aria-jpl/dem_stitcher/). It's an unstable repo in the sense that I am the only maintainer. But this is still more stable than copy/pasting code and maintaining it in two places. Will be maintained for ARIA pipeline.
2. Requires `geopandas` (which is awesome) - maybe not a con?
3. VRT warping not implemented currently

## Motivation and Context
Make sure that DEM is in area coordinates.

## How Has This Been Tested?

This is imported from [repo](https://github.com/aria-jpl/dem_stitcher/) that is distributed on pypi and ensures basic tests are performed.

Wrapped in RAiDER and can be used as below:

```
from RAiDER.dem import getDEM

# Bay area
# xmin, ymin, xmax, ymax
bounds = [-121.5, 34.95, -120.2, 36.25]
# ymin, ymax, xmin, xmax
extent = [bounds[1], bounds[3], bounds[0], bounds[2]]
getDEM(extent)
```

The resulting data is called `glo_30.dem.wgs84` in the current working directory.

N/A

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- I have added an explanation of what your changes do and why you'd like us to include them.


cmarshak:
I will not do virtual warping because other DEMs are not in gdal readable format on the servers they are stored e.g. 3dep. This would require more work and logic. Moreover, tiles are removed after downloaded. Very useful to have feature to download original tiles (and save if necessary) for debugging/inspecting resampling.

cmarshak:
As an FYI, the builds are taking 20 minutes now. That's a really long time. One of the builds exited with code [137](https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory#:~:text=This%20is%20a%20Linux%20error,memory%2C%20your%20build%20will%20fail.) - and thats before pytest was run. I tested things locally fine.

```...
============================ 253 passed, 5 skipped, 64813 warnings in 60.00s (0:01:00) ============================
```

This build issue maybe an issue I will have to address another time.

tools/RAiDER/dem.py
```diff
@@ -252,72 +254,49 @@ def isInside(extent1, extent2):
...
+    """Downloads tiles, merges, and gets ellipsoidal height of DEM. Output
+    format is ISCE gdal-readable raster.
 
     Parameters
-    __________
-    extent      - A list containing [lat_min, lat_max, lon_min, lon_max]
-    out_dir     - Directory to write the full-resolution DEM
-    num_threads - Number of threads to use when warping
+    ----------
+    extent : list
```
jlmaurer, line 266: Should include the ordering of the bounding box elements

tools/RAiDER/dem.py
```diff
@@ -252,72 +254,49 @@ def isInside(extent1, extent2):
...
-        logger.warning(
-            'Query area encompasses {} km^2, supersedes DEM maximum download'
-            'area of 225000km, so I will download the DEM in chunks'.format(query_area)
-        )
-
-    Nchunks = max(int(np.ceil(query_area / 225000)) + 1, 2)
-    chunk_size = (lon_max - lon_min) / Nchunks
-    lon_starts = np.arange(lon_min, lon_max, chunk_size)
+    out_dir : str, optional
+        A list containing [lat_min, lat_max, lon_min, lon_max]
```
jlmaurer, line 269: Looks like this line should up a line?

tools/RAiDER/dem.py
jlmaurer: Saw the list below, just seems to be in the wrong place?

jlmaurer (changes requested):
Looks like there is a typo in the arguments, otherwise looks good! 

A question on this: is the DEM that ends up getting downloaded already ellipsoidal-corrected? If so we should probably wait to merge until we can apply the same correction to the weather model grid node elevations.

cmarshak:
Yea - I flipped the descriptions in the docstring.

I updated that and removed some functions that were no longer being used.

cmarshak:
> question on this: is the DEM that ends up getting downloaded already ellipsoidal-corrected? If so we should probably wait to merge until we can apply the same correction to the weather model grid node elevations.

Not quiet. Each DEM is referenced to a geoid. GLO-30 is EGM-08 and SRTMv3 is EGM-96. In this update, we convert these heights into the ellipsoidal heights.  We can convert to specific geoid as well depending on what is needed. We should coordinate more about what is needed for RAiDER.

jlmaurer:
@cmarshak looks good to me.

jlmaurer (approved):

Issue #308, opened 2021-05-28
proj6 and gdal3


dbekaert:
@bbuzz31 @jlmaurer could we check how raider is impacted for this?
 https://github.com/isce-framework/isce2/pull/279#event-4808095048

jlmaurer:
@dbekaert RAiDER definitely assumes GDAL3 at the moment and all the syntax assumes this.

dbekaert:
@jlmaurer will add documentation before this can be clsoed

jlmaurer:
Our current situation with PROJ, pyproj, and GDAL:
- Currently GDAL3 is required by the environment file. However, as best I can tell there is no GDAL3-specific code used in the repository. I believe we imposed GDAL3 for the purpose of being able to read and write HDF5/NETCDF files; however, we are not currently using GDAL to read or access those files in the code itself (not including dependencies) 
- I checked the dependencies as well as possible. Rasterio requires GDAL 2.4 or later; geopandas requires Fiona (link to GDAL) and Pyproj but does not specify version dependency for GDAL, pyproj must be >=2.2.
- Pyproj may be a problem for some users down the line because the latest version requires Python>3.7 and PROJ 7. As long as users are using a conda environment there should be no issue, but other users may have problems. 
- Our current pyproj requirement in the environment is 2.1.0, however this is not consistent with geopandas. We should require >=2.2 so we are consistent across the board and the "always_xy" argument should always get handled correctly.

jlmaurer:
#319 addresses the pyproj dependency issue.

jlmaurer:
This was actually addressed by #323

PR #319, opened 2021-09-07
Update pyproj requirement

jlmaurer:
## Description
See #308 for rationale. 

## Motivation and Context

Our current version of pyproj required is >=2.1.0, however this is inconsistent with the "always_xy" argument and with geopandas, which is now a requirement.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -27,7 +27,7 @@ dependencies:
  - pandas
  - progressbar
  - pygrib
- - pyproj>=2.1.0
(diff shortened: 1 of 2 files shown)


jlmaurer:
This is breaking with an out-of-memory code. I ran it on locally and it took a *long* time to resolve all the dependencies. Not sure what is the problem, but perhaps we need to re-visit all the dependencies and trim unneeded ones.

jlmaurer:
This has been taken care of by #323

PR #325, opened 2021-09-21
Initial fix for converting weather model heights to ellipsoidal

### LINKED ISSUE #293: [BUG] Heights are not currently being consistently handled throughout the code

jlmaurer:
Problem: the DEM is ellipsoidal heights, while the weather model heights are geoidal
Potential fix: Convert the weather model heights to ellipsoidal before saving the NETCDF file

## Description

- This solution is based on  @cmarshak 's [dem_stitcher](https://github.com/aria-jpl/dem_stitcher) program, but implements a separate version of the correction that works with NDarrays instead of rasters. 
- The idea is simple: download a geoid correction from AGISOFT, interpolate to the weather model grid nodes, and then add the correction to the z-values of the weather model grid
- Uses the efficient [RegularGridInterpolator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html) in scipy, which is also used elsewhere in the code. 
- No new dependencies introduced
- Main functionality is in the `_convert_z_to_ellipsoid` method of the base weather model class. 
- This PR also contains a bug fix for the _get_ll_bounds method

## Motivation and Context

Fixes #293 

## Caveats / Remaining Issues
- It's not currently clear to me what geoid correction to use. In the linked issue (#293) I reference the ECMWF page that states geoidal heights are in EGM96. However, I don't see how EGM08 would be incorrect to use (everything gets calculated from the geopotential), and as it's a better geoid maybe we ought to use that one instead of EGM96. 
- The same issue holds for the other weather models that are not ECMWF - is it correct to use a generic geoid for them? May need to reach out to ECMWF folks to get an idea of what is correct on this front. 
- Along with these issues we should probably document the code a little bit better about the issues with vertical projections

## How Has This Been Tested?

- Methodology has been tested for consistency
- Need to add unit tests (can be done before merging)
- Need to format to match autopep8 (can be done before merging)

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/models/weatherModel.py
+++ b/tools/RAiDER/models/weatherModel.py
@@ -202,6 +202,8 @@ class WeatherModel(ABC):
 
             # Process the weather model data
             self._find_e()
+            # can add the allipsoidal correction here, get the projection info from the original netcdf files
+            self._convert_z_to_ellipsoid() # this is going to primarily come from remove_geoid in geoid.py in dem_stitcher
             self._uniform_in_z(_zlevels=_zlevels)
             self._checkForNans()
             self._get_wet_refractivity()
@@ -211,6 +213,11 @@ class WeatherModel(ABC):
             # Compute Zenith delays at the weather model grid nodes
             self._getZTD(zref)
             return None
+
+
+    def _convert_z_to_ellipsoid(self): 
+        '''Convert original geoidal heights to ellipsoidal'''
+        raise NotImplementedError
 
     @abstractmethod
     def load_weather(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


tools/RAiDER/models/weatherModel.py
```diff
@@ -212,6 +218,26 @@ def load(
...
+        Ny = geoid_profile['height']
+        geoid_x = [gt[2] + gt[0]*k for k in range(Nx)]
+        geoid_y = [gt[5] + gt[4]*k for k in range(Ny)]
+        geoid_y.reverse()
+        new_x, new_y, new_arr = geoid_transformer.transform(geoid_x, geoid_y, np.flipud(geoid_arr))
+        F = rgi((new_y, new_x), new_arr, method='linear', bounds_error=False)
+        ellip_correction = F((self._ys, self._xs))
+        self._zs = self._zs + ellip_correction
+
+
```
cmarshak, line 240: This seems to be the place where I can comment. Overall looks great.

So, I assume that the geo-metadata was not possible to retrieve from the original weather model files [1]. That said, my only suggestion is to distill the method that is being used for reprojection and encapsulate that in a function. For example, `reproject(src_arr, src_transform, dst_transform)`. Would be easier to test and debug down the line. I am not familiar with the array operations and assume they are code you have used elsewhere. My experience with pyproj is that it is meant for vector data with `(x, y, z)` - so interesting to see this. Not sure of the speed differences with gdal/rasterio.

Looks like the broadcasting is correct and so that's great.

[1] https://rasterio.readthedocs.io/en/latest/topics/datasets.html (e.g.`ds = rasterio('NETCDF:<path_to_file>:/path/to/group')`.

cmarshak: The other parts of the code look to be cleanup and I trust that you are doing this to improve the overall performance and may have been integrated from other pending pull requests.

cmarshak (commented):
If there are other parts of the pull request you want me to take a look at - let me know; otherwise, take my comments as you wish. Thanks for integrating these Jeremy!

jlmaurer:
Closing this as it is not needed anymore per [Discussion 501](https://github.com/dbekaert/RAiDER/discussions/501)

PR #326, opened 2021-09-28
Pass closest localtime as opposed to mean within range

sssangha:
Using the `--localtime` option from `raiderCombine.py` to access a specific set of GNSS-weather model residuals, the user must specify the local time of interest (first argument) and the search interval in hours (second argument). So that is to say, if your specified search interval is sufficiently large, you are more likely to pass through more observations throughout the day.

E.g. if the defined search interval is 3 hours for a given local time 6 am, then the program will search for and pass observations spanning anytime between 3 am and 9 am. The problem with the code was that this behavior was intentional in that all of these observations are passed and averaged. I instead propose it is more appropriate to only take the set of observations closest to the defined localtime and pass it. In this way, we may be more precise in reporting only observations closest to the defined local time.

To demonstrate issues related to this bug, compare the residual standard deviation outputs below generated with +/-2 hr and +/-3 hour thresholds, respectively. Note how the error over Europe — for instance — increases. This is because multiple observations spanning 3-9 am are being averaged, instead of taking the observation closest to the specified local time 6 am:
ZTD_minus_RAiDER_grid_delay_stdev

ZTD_minus_RAiDER_grid_delay_stdev

Here is the output generated with the +/- 3 hour threshold after the proposed update. Note how the residual across Western Europe is consistent vis-a-vis that derived with the +/- 2 hour threshold above:

ZTD_minus_RAiDER_grid_delay_stdev

On a related side-note, I should stress that being able to loosen the search interval is especially crucial for global-scale studies, as you are unable to achieve global coverage with localtime search intervals that are too conservative.

On an unrelated note, I’ve also added a dummy-check to filter out columns from the input files that aren’t used for plotting. If these are left in, it will lead to issues with CSV concatenation.

### DIFF
--- a/tools/RAiDER/gnss/processDelayFiles.py
+++ b/tools/RAiDER/gnss/processDelayFiles.py
@@ -189,13 +189,19 @@ def local_time_filter(raiderFile, ztdFile, dfr, dfz, localTime):
     dfr['Localtime_l'] = dfr['Localtime'] - datetime.timedelta(hours=localTime_hrthreshold)
     OG_total = dfr.shape[0]
     dfr = dfr[(dfr['Datetime'] >= dfr['Localtime_l']) & (dfr['Datetime'] <= dfr['Localtime_u'])]
-    print('Total number of datapoints dropped in {} for not being within {} hrs of specified local-time {}: {} out of {}'.format(
+    # only keep observation closest to Localtime
+    dfr['Localtimediff'] = abs((dfr['Datetime'] - dfr['Localtime']).dt.total_seconds() / 3600)
+    dfr = dfr.loc[dfr.groupby(['ID','Localtime']).Localtimediff.idxmin()].reset_index(drop=True)
+    print('Total number of datapoints dropped in {} for not being closest time within {} hrs of specified local-time {}: {} out of {}'.format(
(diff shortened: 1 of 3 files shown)


sssangha:
I pushed a series of new fixes to appropriately parse and plot data by local-time, if specified. The problem in particular involved mishandling of the date time crossing.

The artifacts associated with this bug were apparent as abrupt transitions along certain longitude bands (here a local time of 6 am was specified):
image

Here is the map after the fix (*Note refer to item 2 at the end of this message for significance of vertical blue lines):
Screen Shot 2021-10-03 at 11 12 28 PM

Additionally, I have added to additional features to support this PR:
1. Support to plot datetimes (i.e. hours, and days, years, minutes if the user so wishes). This helped confirm the local-times were appropriately being parsed for each zone via the `—localtime` option through `raiderCombine.py`, for instance here where we had specified a 6pm localtime with a 3 hour search radius:
Screen Shot 2021-10-03 at 10 57 35 PM


2. Estimate central longitudinal lines corresponding to every datetime interval (by hour) and highlight it on the map by specifying the new `—time_lines` option. This information is inherited into the output grids, making repotting of these lines seamless without redoing the analysis.

jlmaurer (commented):
Looks nice to me @sssangha !

PR #331, opened 2021-12-02
fix some comments in LOS and ZREF

jlmaurer:
## Description

- There is a typo in the losreader.py that states heading angle is degrees CW from north, when ISCE in reality uses CCW from east. The comment has been changed to reflect this convention. 
- Also fixed the default value of zref to reflect what is actually used. This should be addressed later as zref is not currently being used in ZTD calculation

### DIFF
--- a/tools/RAiDER/constants.py
+++ b/tools/RAiDER/constants.py
@@ -9,7 +9,7 @@
 import numpy as np
 
 _ZMIN = np.float64(-100)   # minimum required height
-_ZREF = np.float64(15000)  # maximum requierd height
+_ZREF = np.float64(80000)  # default maximum height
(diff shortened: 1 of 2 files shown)


jlmaurer:
@sssangha can you verify that the comment change I made in the losreader file is correct? I.e. that the direction is ccw, not cw.

jlmaurer:
Pinging @sssangha on this. Thanks!

dbekaert:
@jlmaurer can this be closed?

jlmaurer:
@dbekaert I'm still not sure which convention for heading is actually correct. The dev branch still says cw from north, but not sure if that is actually the right one. If so then yes can be closed, if not we should just update the comment.

dbekaert:
Can you not check isce3 documentation?

PR #340, opened 2022-06-09
Fix broken link to HRRR archive

the-garlic-os:
## Description

In [WeatherModels.md](https://github.com/dbekaert/RAiDER/blob/dev/WeatherModels.md), the link to the HRRR archive is missing the protocol bit, causing GitHub web to think it goes to `https://github.com/dbekaert/RAiDER/blob/dev/home.chpc.utah.edu/~u0553130/Brian_Blaylock/hrrr_FAQ.html` instead of `https://home.chpc.utah.edu/~u0553130/Brian_Blaylock/hrrr_FAQ.html`. Adding `https://` to the beginning of the URL fixes this.

## Type of change

## Checklist:


jlmaurer:
@the-garlic-os we probably should add to this PR to fix the problem with the HRRR API as well. @piyushrpt may also look at this, but for now you could take a look at implementing a solution. Did some quick searching and [this page](https://mesowest.utah.edu/html/hrrr/) looks like a good place to start, and also found [this package](https://blaylockbk.github.io/Herbie/_build/html/reference_guide/herbie_api.html) that may help.

dbekaert:
@jlmaurer I dont think this is needed anymore. We can close PR as its addressed in PR#379

jlmaurer:
Yep agree!

PR #346, opened 2022-06-14
flake8

the-garlic-os:
## Description

Resolves problems found with flake8. If a problem could not be immediately fixed without clarification or breaking changes, a "TODO" comment was left near the problem.

## Type of change

## Checklist:

- [X] I have successfully ran tests with your changes locally.

### DIFF
--- a/test/test_scenario_1.py
+++ b/test/test_scenario_1.py
@@ -1,14 +1,10 @@
 import datetime
 import os
 import pytest
-import urllib.error
 
 import numpy as np
 
 from test import TEST_DIR, pushd
-
-from pathlib import Path
-from test import DATA_DIR, TEST_DIR, pushd
 
 from RAiDER.losreader import Zenith
 from RAiDER.delay import tropo_delay
@@ -47,15 +43,6 @@ def test_tropo_delay_ERA5T(tmp_path):
 
 
 @pytest.mark.long
-def test_tropo_delay_GMAO(tmp_path):
-    '''
-    Scenario:
-    1: Small area, GMAO, Zenith delay
-    '''
-    core_test_tropo_delay(tmp_path, modelName="GMAO")
-
-
-@pytest.mark.long
 def test_tropo_delay_MERRA2(tmp_path):
     '''
     Scenario:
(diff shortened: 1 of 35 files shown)


test/test_scenario_1.py
```diff
@@ -46,15 +42,6 @@ def test_tropo_delay_ERA5T(tmp_path):
     core_test_tropo_delay(tmp_path, modelName="ERA5T")
 
 
-@pytest.mark.long
```
jlmaurer, line 49: @the-garlic-os is there a reason for deleting this particular test and not others?

tools/RAiDER/utilFcns.py
```diff
@@ -520,36 +517,36 @@ def writePnts2HDF5(lats, lons, hgts, los, lengths, outName='testx.h5', chunkSize
         f['Rays_len'].attrs['MaxLen'] = np.nanmax(lengths)
 
 
-def writeWeatherVars2HDF5(lat, lon, x, y, z, q, p, t, proj, outName=None):
```
jlmaurer, line 523: I don't think this function is used or needed anymore, so if nothing is calling it it can be deleted.

jlmaurer (changes requested):

test/test_scenario_1.py
the-garlic-os: This test was actually a duplicate of the one that was on line 77. `test_tropo_delay_GMAO` still exists and still runs in the test suite.

tools/RAiDER/utilFcns.py
the-garlic-os: I had a feeling. I didn't see anything in the project calling it, but because it was such a big function I was reluctant to remove it entirely in case it was still being worked on. If it's just not used anymore, I'll delete it.

jlmaurer (approved):

jlmaurer:
@the-garlic-os do you need to do any more work on this PR or is it ready to go? changes look good!

the-garlic-os:
With that last commit there were some files I realized I didn't go over the first time, but I'm all done now!

PR #356, opened 2022-08-11
Update ERA-5 API

jlmaurer:
## Description
ECMWF-ERA5 must have updated at some point; as a result the pressure levels were getting downloaded instead of model levels. I updated the API call and now the data gets downloaded and processed correctly. Also cleaned up a few dependencies that are not getting used. 

## How Has This Been Tested?

- unit tests pass
- autopep8

## Type of change

## Checklist:

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -12,6 +12,7 @@ dependencies:
 # Regular dependencies. Keep these in sync with environment.yml
  - python>=3
  - autopep8
+ - boto3
  - cartopy
  - cdsapi
  - cfgrib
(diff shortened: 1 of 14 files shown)


jlmaurer:
@dbekaert this should be able to be merged. We'll address the coverage in a future PR.

dbekaert (approved):
@jlmaurer LGTM.
My understanding is that this would directly have resulted in an error as it would have been looking for variables that does not exist. Please confirm.

jlmaurer:
@dbekaert that's correct, although the behavior was a little strange. Updating the data URL to use era5-reanalysis-complete resulted in errors; before that it was defaulting to pressure levels without erroring out, but would then raise an error later on in the code when trying to access missing variables.

PR #362, opened 2022-10-10
Use isce3 for LOS - incidence angle computation

piyushrpt:
Use isce3 for geometry calculations instead of previously used python extensions / existing orbit interpolation. 

## Description
- Use `isce3` for look angle computation
- Clean up workflow to have all 3 currently supported options work with golden datasets - zenith, LOS-files, LOS-state vectors.

## Motivation and Context
- Use `isce3` as geometry engine to be consistent with processing workflows.

## How Has This Been Tested?
- Tested with golden datasets provided by @jlmaurer and @bbuzz31 
- Current LOS-state vectors golden dataset seems to have an issue. Hence, compared outputs against the LOS-files option which was also generated for the same geometry with the same inputs by `isce2`. The only difference appears to be due to `float32` files being compared against `float64` calculations from `isce3`
- Testing done using `gdalinfo -stats` of golden dataset and new results. 
- Unit test for checkArgs updated according to changes made

## Questions?
- I have a `compare_stats.py` file that I have been using for testing? This is part of testing infra only and does not need to live in repo. However, if you wanted it here - where would you put it.


jlmaurer (approved):
@piyushrpt this all looks good to me!

PR #363, opened 2022-10-13
Ray Tracing - Computing Ray End points

piyushrpt:
## Description
Determine the end points of the line-of-sight ray for RayTracing option
- `Raytracing._xyz` will represent ECEF positions of targets on the ground
- `Raytracing._topxyz` will represent ECEF positions of ray starting point at the top of the atmosphere
- Both LOS angles as well as state vectors are supported. They are both internally transformed to unit vectors in LOS direction before ray tracing starts.

This PR is more to start discussion of the implementation approach @jlmaurer 

## Motivation and Context
- Slant range for satellite-based SAR images can be on order of 500-1000km
- Troposphere is only significant in the lower 15-30km
- Determining the start / end points through the troposphere only will significantly reduce computation burden for ray tracing

## How Has This Been Tested?
- No unit tests exist yet as we are starting from scratch and this approach is up for discussion
- All previous unit tests work as expected

### DIFF
--- a/test/test_llreader.py
+++ b/test/test_llreader.py
@@ -132,8 +132,8 @@ def test_readLL_files(parser, llfiles):
     assert proj == ''
 
     lats, lons, llproj, bounds, flag, pnts_file_name = readLL(args.query_area)
-    assert np.allclose(lats, lat_true)
-    assert np.allclose(lons, lon_true)
+    assert lats == latfile
+    assert lons == lonfile
     assert llproj == ''
 
     # Hard code the lat/lon bounds to test against changing the files
--- a/test/test_raytracing.py
+++ b/test/test_raytracing.py
@@ -44,6 +44,6 @@ def test_toa():
                        [0.0, 6388137.0, 0.0],
                        [0.0, -6388137.0, 0.0]])
 
-    z.getTopOfAtmosphere(toaheight=10000.)
+    topxyz = z.getIntersectionWithHeight(10000.)
 
-    assert np.allclose(z._topxyz, toppts)
+    assert np.allclose(topxyz, toppts)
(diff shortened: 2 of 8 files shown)


cmarshak:
In the event that this is helpful data point for vectorizing integration along a ray in a 3d cube: https://github.com/cmarshak/Integration-along-ray-in-3d-cube/blob/main/Vectorized%20Ray%20Integration.ipynb

I tried to provide some sanity checks, but maybe I am still off with what is needed. Happy to demonstrate it in real time. It works decently fast on medium sized datacubes (1,000 x 1,000 x 10) and rays with 20 points.

piyushrpt:
I have updated this PR to support both options:

1. `getIntersectionWithHeight` - that lets you get top of atmosphere or any height directly
2. `getIntersectionWithLevels` - that lets you intersection points with multiple levels at same time
   - This probably needs to be used with caution, as just using intersection with levels can miss traversal through multiple voxels in between
   - Best approach might still be to use a finely spaced regular grid (100m along the ray?). 10m seems like an overkill.

piyushrpt:
This PR has now been updated with the logic to only use bounds for fetching weather models.
1. For `Conventional/Zenith` - lat/lon files are only read just before they are absolutely needed
2. Logic for this has also been isolated from the path of `Raytracing`
3. Ideally, we would only load parts of lat/lon file as needed for processing a chunk. Currently, there is no chunking for `Conventional/Zenith` so this has not been implemented yet.
4. This should not be affected by any changes to Raytracing chunking logic. 

All old unit tests work with the changes and numerical results are not impacted.

piyushrpt:
I think we will revisit the `TODO` on a later date. I'm just tagging them to make sure that we do not forget about them

jlmaurer (approved):
LGTM

PR #364, opened 2022-10-17
First crack at a config file inputs

jlmaurer:
## Type of change

## Checklist:


tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+########## 5. Line-of-sight or zenith calculations
+##  RAiDER currently supports four options: 
+##  1. Calculate Zenith delays (ZTD) only (Default)
+##  2. Specify a line-of-sight raster file, either in radar coordinates 
+##     (requires lat/lon files in radar coordinates as query points)
+##     or a geocoded raster (from which the lat/lon locations will be extracted).
+##  2. Specify a georeferenced line-of-sight cube. This should be an 
+##     xarray-readable NETCDF4 file containing the field "los."
+##  3. Specify an orbit file containing state vector files
+los_file:
```
piyushrpt, line 49: Do we want to add a `los_convention` option to specify ISCE or HYP3 etc

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+##  1. Calculate Zenith delays (ZTD) only (Default)
+##  2. Specify a line-of-sight raster file, either in radar coordinates 
+##     (requires lat/lon files in radar coordinates as query points)
+##     or a geocoded raster (from which the lat/lon locations will be extracted).
+##  2. Specify a georeferenced line-of-sight cube. This should be an 
+##     xarray-readable NETCDF4 file containing the field "los."
+##  3. Specify an orbit file containing state vector files
+los_file: 
+los_cube: 
+orbit_file:  # currently ESA-formatted files are accepted
```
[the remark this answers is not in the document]
jlmaurer: Yep, definitely! I'll add that in.

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
+# vim: set filetype=yaml:
+##------------------------ raiderDelay.cfg ------------------------##
+########## 1. Date and time
+## Dates can be a single date, two dates that define a set of dates every day, or two dates with an interval
```
dbekaert, line 4: 1) in this comment: the date can also be the date_list 
2) Would be good per category to clarify what is minum to be specified.
So here i am unclear what if i give a date_list then i assume that date_start is not required.

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+## Standard formats are accepted. 
+date_start:   # first date to download. Specified as YYYYMMDD
+date_end:   # last date to download, auto for None (single date)
+date_step:   # date interval to download; e.g. 12 for every 12 days. Auto for none (downloads every day)
+date_list:   # Pass in a list of dates directly to process
+time:   # time to download in UTC time. Specified as HH:MM:SS or HH:MM
+
+########## 2. Weather model
+## Currently implemented weather models include: 
+## ERA-5 (ECMWF), ERA-5T (ECMWF), HRES (ECMWF), ERA-I (ECMWF), NCMR (ISRO), HRRR (NOAA), GMAO (NASA), MERRA-2 (NASA)
```
dbekaert, line 14: I think here we should clarify when not providing what the default is

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+weather_model:   ## don't include the text in parentheses in the models above
+
+########## 3. Query points
+## There are four options for specifying query points: 
+## 1. lat/lon raster files (such as produced by the ISCE software)
```
dbekaert, line 19: I think now this gets changed with @piyushrpt thoughts. In the end we are just looking for a GDAL compatible file no, instead of ISCE compatible?

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+## 2. A comma-delimited file containing at least the columns Lat and Lon, and optionally Hgt_m
```
dbekaert, line 20: should clarify this is called station_file?

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+## 3. A bounding box in lat/lon specified as a space-separated list: South North West East
```
dbekaert, line 21: with or without []

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+##    If a geocoded file is passed the relevant points will be automatically extracted.
+lat_file: 
+lon_file: 
+station_file: 
+bounding_box: 
+
+########## 4. Height info
+##  RAiDER will automatically download a DEM if needed; otherwise you may pass
+##  a DEM file, which should be a raster containing georeferencing information. 
+##  If "use_dem_xy" is set to true, then delays will be calculated at the input
```
dbekaert, line 32: DEM is needs to be geocoded. XY might be misleading?

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+station_file: 
+bounding_box: 
+
+########## 4. Height info
+##  RAiDER will automatically download a DEM if needed; otherwise you may pass
+##  a DEM file, which should be a raster containing georeferencing information. 
+##  If "use_dem_xy" is set to true, then delays will be calculated at the input 
+##  DEM pixels.
+##  Alternatively, specify a space-delimited list of height levels on which delays
+##  will be calculated at lat/lon values of your query points.
```
dbekaert, line 35: we should discuss what our entry is to do 3D delays and 2D upon request?

tools/RAiDER/cli/raiderDelay.yaml
jlmaurer: @dbekaert good point. I'll add those comments!

tools/RAiDER/cli/raiderDelay.yaml
jlmaurer: @dbekaert yep I'm not sure best either, some fields are required and some have defaults. We can link back to [this page](https://github.com/dbekaert/RAiDER/blob/dev/WeatherModels.md) for documentation.

tools/RAiDER/cli/raiderDelay.yaml
jlmaurer: These isce files are mentioned only as an example of radar-coordinate files, doesn't have to be isce-compatible.

tools/RAiDER/cli/raiderDelay.yaml
jlmaurer: Not sure yet what formats are allowed in yaml, I'll check.

tools/RAiDER/cli/raiderDelay.yaml
```diff
@@ -0,0 +1,60 @@
...
+## Currently implemented weather models include: 
+## ERA-5 (ECMWF), ERA-5T (ECMWF), HRES (ECMWF), ERA-I (ECMWF), NCMR (ISRO), HRRR (NOAA), GMAO (NASA), MERRA-2 (NASA)
+weather_model:   ## don't include the text in parentheses in the models above
+
+########## 3. Query points
+## There are four options for specifying query points: 
+## 1. lat/lon raster files (such as produced by the ISCE software)
+## 2. A comma-delimited file containing at least the columns Lat and Lon, and optionally Hgt_m
+## 3. A bounding box in lat/lon specified as a space-separated list: South North West East
+## 4. Pass a geocoded file for the line-of-sight option (see section 4 below).
```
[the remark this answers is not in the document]
jlmaurer: yep typo here

piyushrpt:
Sorry for bringing this up a little late but do you want to group the arguments as groups:

`inputs`, `processing` and `outputs`

Might help make things very clear as to impact of changing parameters.

jlmaurer:
> Sorry for bringing this up a little late but do you want to group the arguments as groups:
> 
> `inputs`, `processing` and `outputs`
> 
> Might help make things very clear as to impact of changing parameters. Something like this - https://github.com/opera-adt/COMPASS/blob/main/src/compass/defaults/s1_cslc_geo.yaml

Not a problem. This is what I was trying to do with the existing numbering; i.e. everything under 1. is in the "date" group, but I do like making it more explicit. Let me give it a shot and see how it goes.

piyushrpt:
Just documenting here what we discussed in #368 . The two additional parameters that are needed for generating cubes are:
1. Horizontal spacing - called `spacinginm` in the cube prototype
2. Output projection system of code - called `epsg` in the cube prototype

The heights are already captured by `heightlvs` in the main code.

jlmaurer:
@dbekaert @piyushrpt I've got the ZTD working for a bounding box input. It's been quite a bit of changes because of a couple goals: 
1) Don't read the lat/lon points until really needed
2) Along with that, handle the different query point options with a consistent API using a new set of AOI classes. This should be able to abstract quite a bit of several different places where we were doing argument checking, different things for the DEM depending on inputs, etc. In this scenario the geocube should be handled as simply another type of AOI (although I haven't get that far yet). Also I'm thinking it could help with chunking, but haven't looked into that yet. 
3) configuration file for parameters instead of in the CLI. I like this, and it would allow us to run every program from a single config file if desired. I changed the inputs into groups as suggested by @piyushrpt. 
4) Not required and not currently implemented except in the CLI, but we could go to a step-based approach as is used with MintPy where each step is handled as such, although we probably only have two or three steps.

jlmaurer:
Next few steps for this are to update the unit tests and check that a at least the basic examples run, then could be merged.

jlmaurer:
@dbekaert @piyushrpt I renamed some of the test suites that deal with the input argument IO so that they wont run. These will need to be updated to reflect the fact that we aren't using the CLI interface anymore.

jlmaurer:
@dbekaert @piyushrpt if one of you want to give a review and/or approve for merging, the tests are passing now!

piyushrpt (approved):
LGTM

PR #365, opened 2022-10-21
Replace gdal +osr  with rasterio + pyproj

piyushrpt:
## Description
As per discussion
- Replace `gdal` with `rasterio`
- Replace `osr` with `pyproj`
- Hopefully, final set of general changes before details of `Raytracing` are flushed out

## Motivation and Context
- Use one set of tools consistently
- We had a mix of these tools in the repo before

## How Has This Been Tested?
- All unit tests have been appropriately updated
- Golden datasets were tested against and nothing changed numerically
- Added a couple of TODOs to highlight spots to revisit. Some external scripts like `prepFromAria.py` might need more testing.


test/test_util.py
```diff
@@ -227,24 +229,22 @@ def test_least_nonzero_2():
...
-    band = ds.GetRasterBand(1)
-    band.WriteArray(np.random.randn(11, 11))
-    srs = osr.SpatialReference()
-    srs.ImportFromEPSG(4326)
-    ds.SetProjection(srs.ExportToWkt())
-    ds = None
-    band = None
+    with rasterio.open("test.tif", mode="w",
+                       width=11, height=11, count=1,
+                       dtype=np.float64, crs="EPSG:4326",
```
cmarshak, line 236: I think you might get warnings about using a crs as a string. I use

```
from rasterio.crs import CRS

crs = CRS.from_epsg(4326)
```

tools/RAiDER/models/weatherModel.py
```diff
@@ -566,11 +566,10 @@ def getXY_gdal(self, filename):
...
         '''
-        from osgeo import gdal
-        ds = gdal.Open(filename, gdal.GA_ReadOnly)
-        xSize, ySize = ds.RasterXSize, ds.RasterYSize
-        trans = ds.GetGeoTransform()
-        del ds
+        ds = rasterio.open(filename)
+        xSize, ySize = ds.width, ds.height
+        trans = ds.transform.to_gdal()
+        ds.close()
```
cmarshak, line 572: I think this should use context manager:

```
with rasterio.open(filename) as ds:
    xSize, ySize = ds.width, ds.height
    trans = ds.transform.to_gdal()
```

tools/RAiDER/utilFcns.py
```diff
@@ -100,51 +98,47 @@ def ecef2enu(xyz, lat, lon, height):
...
-    proj = ds.GetProjection()
-    gt = ds.GetGeoTransform()
+    with rasterio.open(fname) as src:
+        gt = src.transform.to_gdal()
+        proj = src.crs
+        xSize = src.width
+        ySize = src.height
+
     if not proj or not gt:
         raise AttributeError('File {} does not contain geotransform information'.format(fname))
```
cmarshak, line 112: This can be simplified to be:

```
with rasterio.open(fname) as ds:
    bounds = ds.bounds

return ds.bounds
```

tools/RAiDER/utilFcns.py
```diff
@@ -100,51 +98,47 @@ def ecef2enu(xyz, lat, lon, height):
...
+                logger.debug('Using user-supplied NoDataValue')
+                data[data == userNDV] = np.nan
+            else:
+                try:
+                    ndv = src.nodatavals[bnd]
+                    data[data == ndv] = np.nan
+                except BaseException:  # TODO: Which error(s)?
+                    logger.debug('NoDataValue attempt failed*******')
+            val.append(data)
+    src.close()
```
cmarshak, line 141: `src.close()` is not needed.

The val list is not needed either

```
with rasterio.open(fname) as ds:
   data = ds.read().transpose(1, 2, 0)
```

tools/RAiDER/utilFcns.py
```diff
@@ -157,25 +151,16 @@ def gdal_open(fname, returnProj=False, userNDV=None):
         return data, proj, gt
```
cmarshak, line 151: It may be advantageous to instead just return the rasterio profile here when `returnProj == True`. That could potentially make code more general and subsequent writing of rasters more streamlined. The projection and geotransform can be accessed as keys in this profile dictionary e.g.

```
with rasterio.open(filename) as ds:
    profile = ds.profile
proj = profile['crs']
gt = profile['transform']
```

This is obviously potentially annoying due to change in return types. It will simplify code when *writing* after reading a similar raster e.g.

```
with rasterio.open(filename_0) as ds:
    X = ds.read()
    p = ds.profile

with rasterio.open(filename_1, 'w', **p) as ds:
   ds.write(X + 1)
```

Anyways, just an idea.

tools/RAiDER/utilFcns.py
```diff
@@ -814,17 +796,21 @@ def write2NETCDF4core(nc_outfile, dimension_dict, dataset_dict, tran, mapping_na
...
+        # var.setncattr('false_easting', srs.GetProjParm('false_easting'))
+        # var.setncattr('false_northing', srs.GetProjParm('false_northing'))
+        # var.setncattr('latitude_of_projection_origin', np.sign(srs.GetProjParm('latitude_of_origin')) * 90.0)
+        # var.setncattr('latitude_of_origin', srs.GetProjParm('latitude_of_origin'))
+        # var.setncattr('semi_major_axis', float(srs.GetAttrValue('GEOGCS|SPHEROID', 1)))
+        # var.setncattr('scale_factor_at_projection_origin', 1)
+        # var.setncattr('inverse_flattening', float(srs.GetAttrValue('GEOGCS|SPHEROID', 2)))
+        var.setncattr('spatial_ref', crs.to_wkt())
+        var.setncattr('spatial_proj4', crs.to_proj4())
         var.setncattr('spatial_epsg', epsg)
```
cmarshak, line 814: This type of georeferenced writing to a netcdf can be greatly simplified with https://corteva.github.io/rioxarray/stable/

tools/RAiDER/utilFcns.py
```diff
@@ -100,51 +98,47 @@ def ecef2enu(xyz, lat, lon, height):
     return np.stack((e, n, u), axis=-1)
 
 
-def gdal_extents(fname):
+def rio_extents(fname):
     if os.path.exists(fname + '.vrt'):
         fname = fname + '.vrt'
```
jlmaurer, line 103: This may not be needed anymore. We were facing an issue where gdal wouldn't read certain attributes from the .dat file but would from the .vrt file. I think it was mainly the NoDataValue. I don't think rasterio has this limitation; @cmarshak do you happen to know?

tools/RAiDER/utilFcns.py
jlmaurer: +1 on this

tools/RAiDER/utilFcns.py
jlmaurer: +1 on this idea if it doesn't introduce too much complexity / other issues. I think this could actually simplify the code since all the attributes would get dragged along instead of just one or two .

tools/RAiDER/utilFcns.py
```diff
@@ -223,26 +208,24 @@ def writeArrayToRaster(array, filename, noDataValue=0., fmt='ENVI', proj=None, g
     array_shp = np.shape(array)
```
jlmaurer, line 208: I think this section (here and the next several lines) are where bringing along the attributes noted above would help a lot, since then everything is in there and we don't have to name it separately. We simply have one input dictionary that gets assigned when writing the file (new line 224 below).

test/test_util.py
piyushrpt: Done.

tools/RAiDER/models/weatherModel.py
piyushrpt: Done.

tools/RAiDER/utilFcns.py
piyushrpt: `bounds` appears to use a different convention from `SNWE` . Will keep it as is for now - unless we want to go in and update all usages of bbox.

tools/RAiDER/utilFcns.py
piyushrpt: We seem to modify data type.  I will keep this as is for now - if we strongly feel about this; we can go in and update everything to use profile instead.

tools/RAiDER/utilFcns.py
piyushrpt: I'm not sold on rioxarray. Will let team decide on it. `pyproj.to_cf` works well as it is the intended usage. I will leave changes to netcdf creation to folks who are owners of this capability.

tools/RAiDER/utilFcns.py
piyushrpt: Done.

tools/RAiDER/utilFcns.py
piyushrpt: Done.

jlmaurer (approved):

PR #366, opened 2022-10-25
update packaging and swtich to entrypoints

### LINKED ISSUE #358: Empty `tools` python module artifact in installation
The conda-forge reviewer had [this suggestion](https://github.com/conda-forge/staged-recipes/pull/20100#pullrequestreview-1090694502) which is worth addressing here:

> The installed artifacts contain[ an empty python module called "tools".](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=552697&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1814) which is installed directly to site-packages. It seems intentional, but I'm not sure what the purpose is because it is empty when installed to site-packages.

This appears to be an artifact of how the package and setuptools is setup (not intentional) and likely driven by how the python package is structured and this line: 
https://github.com/dbekaert/RAiDER/blob/dev/setup.py#L99

### LINKED ISSUE #359: Transition "scripts" to entry points/console scripts
The conda-forge reviewer had [this suggestion](https://github.com/conda-forge/staged-recipes/pull/20100#pullrequestreview-1090694502) which is worth addressing here:

> I also encourage you to transition your "scripts" to [entrypoints/console scripts](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#console-scripts) by listing function names in the setup.cfg/pyproject.toml. Instead of maintaining a bunch of files in bin which essentially are just redirecting to functions in Raider.

I would also highly encourage this as it makes RAiDER more portable and easier to maintain. For reference, I discussed a lot of these issues the effort to get MintPy on conda-forge here: https://github.com/insarlab/MintPy/issues/652 , includign an example of how console_scripts entry_points work.

jhkennedy:
Overall, this PR should be mostly transparent to users and is focused on packaging. It:
1. updates `setup.py` to include better metadata 
1. update `setup.py` to fix #358
1. Switches `seup.cfg` to `pyproject.toml` to be more inline with current python packaging recommendations
1. updates the `pybind11` extension to follow [current recommendations](https://github.com/pybind/python_example)
1. drops `environment-dev.yml` now that RAiDER is on conda-forge, the repo environment is really only needed for setting up a development environment
1. convert the old and [depreciated](https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts) `distutils` [`scripts`](https://docs.python.org/3/distutils/setupscript.html#installing-scripts) argument in `setup.py` to use  [`console_scripts` `entry_points`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#console-scripts) which are platform agnostic and significantly more portable (more below). 
   Fixes #359
   * Added `pytest` tests to ensure `entry_points` are installed correctly
1. drop all `!#` bangs as pip should be used for develop/source installs and `PYTHONPATH` should *never* be manipulated anymore 
1. updates CI scripts accordingly

## Type of change

## Checklist:

---

### `console_scripts` `entry_points` example

Entrypoints are definitely weird, and I think it typically works best to explain them with an example of what all happens with them. I'll use `tools/bin/prepARIA.py`  as the example here. 

If I wanted to make a `console_scripts` `entry_points`  for `prepARIA.py` with the same interface (meaning a `prepARIA.py` script available on `PATH` I can call), I'd need to:

#### 1. Define the entry point in `setup.py`

This would look like
 ```python
from setuptools import setup

setup(
    ...
    entry_points={'console_scripts': [
       'prepARIA.py = RAiDER.prepFromAria:prepFromAria',
        ]
    },
    ...
)
```
then, when `pip install ...` is called, `pip` will create a wrapper script called `prepARIA.py` which calls the `main` function (*with no arguments!*) in the ` RAiDER.prepFromAria` module. 
  
In a conda environment, that script will be located at `${CONDA_PREFIX}/bin/prepARIA.py` and will have two functional lines:
```python
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(load_entry_point('RAiDER', 'console_scripts', 'prepARIA.py')())
```
which is "set `sys.argv[0]` to my entrypoint name, load the entrypoint function, and execute it".


piyushrpt (approved):
LGTM

jlmaurer (approved):
LGTM

PR #370, opened 2022-11-01
Ray tracing cube prototype

piyushrpt:
## Description
Includes a prototype for ray tracing with output on a cube. This is meant to be prototype that will be integrated with rest of workflow once finalized.

## Motivation and Context
- Ray tracing output on a cube that can then be used for high res delay computation
- Breaks, integration along LOS ray as integration over steps between different height levels
- There is more room for vectorization at the expense of memory usage
- Eliminates need for custom cython/ pybind11 bindings
- Additional checks will be needed on parsing inputs to increase padding for ray tracing operations - depending on max altitude of model being used

## How Has This Been Tested?
- Golden tests still pass
- Numerical sniff test for values - seems reasonable
- Needs more elaborate testing


jlmaurer (approved):
@piyushrpt LGTM.

jlmaurer:
@piyushrpt Reviewed and approved! I'll let you merge in case you need to make any changes.

PR #371, opened 2022-11-05
Docker

jhkennedy:
This sets up the mechanics for building a docker container for any PR (to ensure it builds) and and push to main or dev, and then push it to this repositories container registry. See: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

container images will always be tagged with the current version number, and then:
*  for pushes to develop also tagged with `test`
* for pushes to main also tagged with `latest`

Note: I'm using actions I've developed and maintained here: 
https://github.com/ASFHyP3/actions
Specifically:
* [`reusable-git-object-name.yml`](https://github.com/ASFHyP3/actions/blob/develop/README.md#reusable-git-object-nameyml)
* [`reusable-docker-ghcr.yml`](https://github.com/ASFHyP3/actions/blob/develop/README.md#reusable-docker-ghcryml)

The version number generated for the docker container will be the same as reported by:
```
git describe --dirty --tags --long --match "*[0-9]*"
```
which for the current state of this PR, would look like:
```
v0.1.0-73-g78a27fc
```
which breaks down to `<latest-version-tag>-<commits-since-tag>-g<commit-short-hash>`. This will tie any particular docker image directly to the associated commit used to build it, and the "package version" (notably, RAiDER reports a static version based on the `version.txt` file, which can be out of sync with the git tags; I'll tackle that in a subsequent PR).

If you're *on* a version tag (or the commit associated with the version tag) the version will be reported just as:
```
v0.1.0
```
as expected.

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,49 @@
+FROM condaforge/mambaforge:latest
+
+# For opencontainers label definitions, see:
+#    https://github.com/opencontainers/image-spec/blob/master/annotations.md
+LABEL org.opencontainers.image.title="RAiDER"
+LABEL org.opencontainers.image.description="Raytracing Atmospheric Delay Estimation for RADAR"
+#LABEL org.opencontainers.image.vendor=""
+#LABEL org.opencontainers.image.authors=""
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+LABEL org.opencontainers.image.url="https://github.com/dbekaert/RAiDER"
+LABEL org.opencontainers.image.source="https://github.com/dbekaert/RAiDER"
+LABEL org.opencontainers.image.documentation="https://github.com/dbekaert/RAiDER"
+
+# Dynamic lables to define at build time via `docker build --label`
+# LABEL org.opencontainers.image.created=""
+# LABEL org.opencontainers.image.version=""
+# LABEL org.opencontainers.image.revision=""
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV PYTHONDONTWRITEBYTECODE=true
+
+RUN apt-get update && apt-get install -y --no-install-recommends git unzip vim && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Create non-root user/group with default inputs
+ARG CONDA_UID=1000
+ARG CONDA_GID=1000
+
+RUN groupadd -g "${CONDA_GID}" --system raider && \
+    useradd -l -u "${CONDA_UID}" -g "${CONDA_GID}" --system -d /home/raider -m  -s /bin/bash raider && \
+    chown -R raider:raider /opt
+
+USER ${CONDA_UID}
+WORKDIR /home/raider
+SHELL ["/bin/bash", "-l", "-c"]
+
+COPY --chown=${CONDA_UID}:${CONDA_GID} . /RAiDER/
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM condaforge/mambaforge:latest
+
+# For opencontainers label definitions, see:
+#    https://github.com/opencontainers/image-spec/blob/master/annotations.md
+LABEL org.opencontainers.image.title="RAiDER"
+LABEL org.opencontainers.image.description="Raytracing Atmospheric Delay Estimation for RADAR"
+#LABEL org.opencontainers.image.vendor=""
+#LABEL org.opencontainers.image.authors=""
```
jhkennedy, line 8: @dbekaert anything you'd like placed here?

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    conda clean -afy
+
+# Ensure that environment is activated on startup
+RUN echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.profile && \
+    echo "conda activate RAiDER" >> ~/.profile
+
+RUN python -m pip install --no-cache-dir /RAiDER/
+
+ENTRYPOINT ["/usr/bin/bash"]
+CMD ["-l"]
```
jhkennedy, line 49: This right now just gives you a bash shell in the container. I suspect we'll update it to `RAiDER.__main__` if/when that's available.

environment.yml
```diff
@@ -36,6 +36,7 @@ dependencies:
  - sysroot_linux-64
  - xarray
  # For packaging and testing
+ - pybind11>=2.5.0
```
jhkennedy, line 39: needed for `python -m setup.py --version` to work. Could go in the "for running" section, which really should be "for developing" I suppose (or those broken out).

Dockerfile
dbekaert: What would you suggest in terms of authors for open-source packages. Main contributors; but how would it change over time?

jhkennedy:
Note: This action needs to be available in the default branch (`dev`) before it will run.

Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM condaforge/mambaforge:latest
+
+# For opencontainers label definitions, see:
+#    https://github.com/opencontainers/image-spec/blob/master/annotations.md
+LABEL org.opencontainers.image.title="RAiDER"
+LABEL org.opencontainers.image.description="Raytracing Atmospheric Delay Estimation for RADAR"
+LABEL org.opencontainers.image.authors="David Bekaert<david.bekaert@jpl.nasa.gov>, Jeremy Maurer <maurer.jeremy@gmail.com>, Piyush Agram <piyush@descarteslabs.com>"
```
jlmaurer, line 7: @jhkennedy I suggest to add @sssangha to this list as the primary developer of the GNSS interface and stats modules

dbekaert, line 7:
```suggestion
LABEL org.opencontainers.image.authors="David Bekaert<david.bekaert@jpl.nasa.gov>, Jeremy Maurer <maurer.jeremy@gmail.com>, Simran Sangha <simran.s.sangha@jpl.nasa.gov>, Piyush Agram <piyush@descarteslabs.com>"
```

jlmaurer (approved):
LGTM

PR #375, opened 2022-11-07
Fixes for cubes in non lat/lon systems

piyushrpt:
## Description
Projection systems and axis orders are a pain. This fixes cube generation in other coordinate systems

## Motivation and Context
NISAR products will use UTM system. 

## How Has This Been Tested?
Template input file used for testing zenith delays on a UTM grid:

```
# vim: set filetype=yaml:

look_dir: right

date_group:
  date_start:
  date_end:
  date_step:
  date_list:  [20200101]

time_group:
  time: "23:00:00"
  end_time:

weather_model: HRES

aoi_group:
  lat_file:
  lon_file:
  height_file_rdr:
  station_file:
  bounding_box: 53.0 56.0 -4.0 -1.0
  utm_zone:
  grid_x:
  grid_y:

height_group:
  dem:
  use_dem_latlon: False
  height_file_rdr:

los_group:
  ray_trace: False  # Use projected slant delay by default
  zref: # Integration height. Only used in raytracing.
  los_file:
  los_cube:
  orbit_file:

runtime_group:
  verbose: True
  output_directory: current/HR_HRES
  weather_model_directory: current/HR_HRES/weather_files
  cube_spacing_in_m: 2000.
  output_projection:  32630
```

### DIFF
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -179,7 +179,7 @@ def tropo_delay(dt, wf, hf, args):
         wetDelay = ifWet(pnts)
         hydroDelay = ifHydro(pnts)
 
-        # TODO - handle this better during parsing
+        # TODO - handle this better during set up of LOS object
         los._pad = 600
         los._time = dt
 
@@ -321,13 +321,20 @@ def tropo_delay_cube(dt, wf, args, model_file=None):
         out_type = "slant range"
         out_filename = wf.replace("_ztd", "_std").replace("wet", "tropo")
 
+        if args["look_dir"].lower() not in ["right", "left"]:
+            raise ValueError(
+                f'Unknown look direction: {args["look_dir"]}'
+            )
+
         # Get pointwise interpolators
         ifWet, ifHydro = getInterpolators(weather_model_file, "pointwise")
+
+
 
         # Build cube
         wetDelay, hydroDelay = build_cube_ray(
             xpts, ypts, zpts,
-            dt, args["los"]._file,
+            dt, args["los"]._file, args["look_dir"],
             wm_proj, crs,
             [ifWet, ifHydro])
 
@@ -478,7 +485,7 @@ def build_cube(xpts, ypts, zpts, model_crs, pts_crs, interpolators):
     return outputArrs
 
 
-def build_cube_ray(xpts, ypts, zpts, ref_time, orbit_file, model_crs, pts_crs, interpolators):
+def build_cube_ray(xpts, ypts, zpts, ref_time, orbit_file, look_dir, model_crs, pts_crs, interpolators):
     """
     Iterate over interpolators and build a cube
     """
@@ -498,10 +505,15 @@ def build_cube_ray(xpts, ypts, zpts, ref_time, orbit_file, model_crs, pts_crs, i
     ])
 
     # ISCE3 data structures
-    # TODO - Assuming right-looking for now
     elp = isce.core.Ellipsoid()
     dop = isce.core.LUT2d()
-    look = isce.core.LookSide.Right
+    if look_dir.lower() == "right":
+        look = isce.core.LookSide.Right
+    elif look_dir.lower() == "left":
+        look = isce.core.LookSide.Left
+    else:
+        raise RuntimeError(f"Unknown look direction: {look_dir}")
+    logger.debug(f"Look direction: {look_dir}")
 
     # Get model heights in an array
     # Assumption: All interpolators here are on the same grid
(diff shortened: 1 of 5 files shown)


tools/RAiDER/delay.py
```diff
@@ -167,7 +167,8 @@ def tropo_delay(dt, wf, hf, args):
                 lons,
                 hgts,
                 pnt_proj,
-                wm_proj
+                wm_proj,
+                flip_xy=pnt_proj.axis_info[0].abbrev in ["X","E","Lon"]
```
dbekaert, line 171: This is always embedded in geo datasets, so no need to rely on software version?

dbekaert (approved):
This looks good to me.

tools/RAiDER/delay.py
piyushrpt: Yeah. This is a function of the datasets / grids. The min version of pyproj that `requirements` is pinned to - this should not be an issue.

dbekaert:
I was initially a bit confused with the URM zone in the AOI group. The outputs as specified in the runTime group. i.e. UTM is embedded into projection 32630.

jlmaurer:
I think those grid options will only be needed if NISAR will have a pre-determined grid to provide delays. If they only need the delays on *any* grid in the appropriate coordinate system then the basic output projection argument is sufficient. @piyushrpt @dbekaert thoughts?

piyushrpt:
My understanding is it would be a regular grid in UTM. I still don't fully understand uniform xyz space. Could you write this down with an example. Were you think `grid_x`, `grid_y` be arrays or files like lat and lon files?

cmarshak:
Is the plan for this Yaml file to run "as is"?  I install raider and get the following error:

```
(raider) --- bekaert-team/raider-example » raiderDelay.py test.yaml
Run routine processing with raiderDelay.py on steps: ['load_weather_model', 'calculate_delays']
Remaining steps: ['calculate_delays']
--------------------------------------------------
  File "/Users/cmarshak/opt/anaconda3/envs/raider/bin/raiderDelay.py", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/cmarshak/bekaert-team/RAiDER/tools/RAiDER/runProgram.py", line 18, in main
    params = read_template_file(inps.customTemplateFile)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cmarshak/bekaert-team/RAiDER/tools/RAiDER/cli/raiderDelay.py", line 247, in read_template_file
    get_heights(
  File "/Users/cmarshak/bekaert-team/RAiDER/tools/RAiDER/cli/validators.py", line 61, in get_heights
    os.mkdir(dem_path)
FileNotFoundError: [Errno 2] No such file or directory: 'current/HR_HRES/geom'
```

piyushrpt:
You will need to create the directory tree as needed for those directories listed in the input file before attempting to run the script.  Default behavior should be something this team should decide and implement - whether to automatically create those folders if they exist or not/ raise an exception.

cmarshak:
> You will need to create the directory tree as needed for those directories listed in the input file before attempting to run the script. Default behavior should be something this team should decide and implement - whether to automatically create those folders if they exist or not/ raise an exception like the code does now

Got it - thanks.

tools/RAiDER/losreader.py
```diff
@@ -625,22 +625,28 @@ def get_radar_pos(llh, orb, out="lookangle"):
...
         pos_llh = ecef2lla(pos[..., 0], pos[..., 1], pos[..., 2])
-        pos = pos + lookvecs * (toaheight - pos_llh[2])[..., None]
+        pos = pos + lookvecs * ((toaheight - pos_llh[2])/factor)[..., None]
 
     # This is for debugging the approach
     # print("Stats for TOA computation: ", toaheight,
-    #      toaheight - np.nanmin(pos_llh[2]),
-    #      toaheight - np.nanmax(pos_llh[2]),
-    #     )
+    #       toaheight - np.nanmin(pos_llh[2]),
```
cmarshak, line 647: Could you add a comment as to what method you are using for this approximation?

Why don't you compute the intersection exactly? Can't you?

piyushrpt: This is simple Newton-Raphson. Exact computation is not possible since closed form expression for xyz to llh transformation is super complicated.

https://github.com/isce-framework/isce3/blob/04b17713adec76348ab63718895cb8eca4912d25/cxx/isce3/core/Ellipsoid.h#L197

piyushrpt: Documentation has been added now

jlmaurer:
> You will need to create the directory tree as needed for those directories listed in the input file before attempting to run the script. Default behavior should be something this team should decide and implement - whether to automatically create those folders if they exist or not/ raise an exception like the code does now

Just a note that we currently do (*if everything is working) create directories as needed for the weather model files and geometry files but not the output directory. If the output directory is specified it should exist beforehand. This makes sense to me but we should at minimum handle the directory existence more gracefully than is being done.

jlmaurer:
> My understanding is it would be a regular grid in UTM. I still don't fully understand uniform xyz space. Could you write this down with an example. Were you think `grid_x`, `grid_y` be arrays or files like lat and lon files?

My thinking was that these would be whatever is needed to define the NISAR grid. You might be able to do simply by defining the zone and scalar grid spacing, since you would already know the UTM origin and the InSAR frame size. Whatever is provided, you would  do the projection of the weather model data then interpolate horizontally to the NISAR grid. I'm not sure how they are defining their grids but that would be the info we need for that part. 

This was my thinking on the issue; but that doesn't mean it's the right way to go.

piyushrpt:
The current cube implementation uses the `output_projection` which is same as UTM system kwarg and `cube_spacing_in_m` which is the grid spacing.  The extent of the grid comes straight out of the `bounding_box` in the current implementation. 

Sounds like what you are describing is consistent with the current implementation? Do we need these specific kwargs or do we remove the other ones currently in use to use these instead?

jlmaurer:
I think we can remove those other kwargs and keep your current implementation. Sounds like its doing what we need!

Issue #376, opened 2022-11-08
[BUG] GMAO weather model issue with YML config file


cmarshak:
**Describe the bug**
There seems to be an issue formatting the GMAO weather model.

**To Reproduce**

Navigate to an empty directory. Create a subdirectory called `weather_files`.

```
raiderDelay.py test.yml
```
with `test.yml` as:

```
look_dir: right

date_group:
  date_start:
  date_end:
  date_step:
  date_list:  [20200101]

time_group:
  time: "22:00:00"
  end_time:

weather_model: GMAO

aoi_group:
  lat_file:
  lon_file:
  height_file_rdr:
  station_file:
  bounding_box: 53.0 56.0 -4.0 -1.0
  utm_zone:
  grid_x:
  grid_y:

height_group:
  dem:
  use_dem_latlon: False
  height_file_rdr:
  height_levels: "-500 0 500 1000 1500 2000 2500 3000 3500 4000 4500 5000 5500 6000 6500 7000 7500 8000 8500 9000"

los_group:
  ray_trace: False  # Use projected slant delay by default
  zref: # Integration height. Only used in raytracing.
  los_file:
  los_convention: isce  # can be "isce" or "hyp3", see *** for details
  los_cube:
  orbit_file:

runtime_group:
  verbose: True
  raster_format: nc # Can be any rasterio-compatible format
  output_directory: '.'
  weather_model_directory: weather_files
  cube_spacing_in_m: 2000.
```

The error is:
```
(RAiDER) --- bekaert-team/raider-example » raiderDelay.py test.yml
Run routine processing with raiderDelay.py on steps: ['load_weather_model', 'calculate_delays']
Remaining steps: ['calculate_delays']
--------------------------------------------------
Weather model GMAO is available from 2014-02-20 00:00:00-Present
WARNING: Rounded given hour from  22 to 21
/Users/cmarshak/opt/anaconda3/envs/RAiDER/lib/python3.11/site-packages/pyproj/crs/crs.py:1286: UserWarning: You will likely lose important projection information when converting to a PROJ string from another format. See: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
  proj = self._crs.to_proj4(version=version)
ERROR: Unable to save weathermodel to file
Traceback (most recent call last):
  File "/Users/cmarshak/bekaert-team/RAiDER/tools/RAiDER/models/gmao.py", line 145, in _fetch
[...]

Issue #336, opened 2022-02-16
[BUG] GMAO dates earlier than 12/2017 fail with Pydap>=3.3.0


scottstanie:
**Describe the bug**
I was creating a bunch of Line of sight-projected maps with GMAO using `raiderDelay.py`. The code works for Pydap version 3.2.1. 
They seem to have changed something so that it fails to grab all the data we need, and it leads to this traceback:

```
(mapping) [scott@grace igrams_looked]$ raiderDelay.py --model GMAO  --time 00:59:21 --lineofsight los_az_inc.tif -d ./elevation_looked.dem --latlon lat.geo.vrt lon.geo.vrt --out ./raider --date 20180827
Weather model GMAO is available from 2014-02-20 00:00:00-Present
WARNING: Rounded given hour from  0 to 0
ERROR: Unable to save weathermodel to file
Traceback (most recent call last):
  File "/home/scott/repos/RAiDER/tools/RAiDER/models/gmao.py", line 148, in _fetch
    writeWeatherVars2NETCDF4(self, lats, lons, h, q, p, t, outName=out)
  File "/home/scott/repos/RAiDER/tools/RAiDER/utilFcns.py", line 772, in writeWeatherVars2NETCDF4
    nc_outfile = write2NETCDF4core(nc_outfile, dimension_dict, dataset_dict, tran, mapping_name='WGS84')
  File "/home/scott/repos/RAiDER/tools/RAiDER/utilFcns.py", line 858, in write2NETCDF4core
    dataset_dict[data]['dataset'][ndmask] = FillValue
TypeError: only integer scalar arrays can be converted to a scalar index
Traceback (most recent call last):
  File "/home/scott/miniconda3/envs/mapping/lib/python3.8/pdb.py", line 1704, in main
    pdb._runscript(mainpyfile)
  File "/home/scott/miniconda3/envs/mapping/lib/python3.8/pdb.py", line 1573, in _runscript
    self.run(statement)
  File "/home/scott/miniconda3/envs/mapping/lib/python3.8/bdb.py", line 580, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/home/scott/miniconda3/envs/mapping/bin/raiderDelay.py", line 3, in <module>
    __requires__ = 'RAiDER==0.0.1'
  File "/home/scott/repos/RAiDER/tools/bin/raiderDelay.py", line 12, in <module>
    parseCMD()
[...]

PR #383, opened 2022-11-09
fix nan-dropping bug

jlmaurer:
Had another bug hatch from the last PR related to dropping nans in sub-dicts (the groups are nice for inputs but are a bit painful to parse accurately)

## How Has This Been Tested?

-added a new unit test to check that the None-catching is working.

## Type of change

## Checklist:

### DIFF
--- a/test/cli/test_raiderDelay.py
+++ b/test/cli/test_raiderDelay.py
@@ -1,0 +1,25 @@
+from argparse import ArgumentParser, ArgumentTypeError
+from datetime import datetime, time
+
+from RAiDER.cli.raiderDelay import drop_nans
+
+def test_drop_nans():
+    test_d = {
+        'key1': 1,
+        'key2': 'string',
+        'key3': None,
+        'key4': {'sub_key1': 1, 'sub_key2': None},
+        }
+    
+    out_d = drop_nans(test_d)
+
+    assert set(out_d.keys()) == set(['key1', 'key2', 'key4'])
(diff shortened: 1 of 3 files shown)


jlmaurer:
While I'm at it I also fixed the bug in accessing GMAO data, at least for the time-period post 12/2017 mentioned in #336. API change was causing the issue. Fixes #376

piyushrpt:
Looks like `output_directory` and related options like `weather_model_directory` etc also got masked out by the previous PR. They are no longer parsed.

jlmaurer:
@piyushrpt does the latest commit solve the missing parameters problem?

piyushrpt:
Yep - fixes it. Another minor bug from testing. `nc` is enfored in `checkArgs` - this used to be `args.raster_format`

jlmaurer:
@piyushrpt yes that's correct, that is just for the cube output which shouldn't take a raster format. Either we can rename that parameter 'format' or we could add another parameter cube_format or something

piyushrpt:
I'm seeing that impact regular tests with lat/lon/dem files as well. Maybe there needs to be some logic there to distinguish between the use cases. Another parameter also works for me. Right now everything is getting channeled through there

piyushrpt (approved):

jlmaurer:
I'll go ahead and merge this. Going to work on getting the raster format back in the right spot.

PR #384, opened 2022-11-09
Fix typo in docker build

jhkennedy:
Workflow didn't run because the `needs:` value was wrong.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     uses: ASFHyP3/actions/.github/workflows/reusable-git-object-name.yml@v0.6.0
 
   call-docker-ghcr-workflow:
-    needs: call-version-info-workflow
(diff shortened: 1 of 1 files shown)


jhkennedy:
We can see the docker build is working! And in the packages registry:
image

which brings you to:
image

@dbekaert, you'll have to make the container images public. You should be able to follow this link to do it (Scroll to the bottom):
https://github.com/dbekaert/packages/container/raider/settings

Of once you're viewing the packages, click the package setting in the bottom right.

dbekaert:
@jhkennedy i changed the settings. Can you confirm we are good to go?

dbekaert (approved):
Thanks for fixing, should be public visible now too

jhkennedy:
> @jhkennedy i changed the settings. Can you confirm we are good to go?

@dbekaert Confirmed!

PR #385, opened 2022-11-09
Update packaging to use pyproj and generate a dynamic version number

jhkennedy:
This updates the python packaging to use `pyproject.toml` where possible as [`setup.py` use is discoraged](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged).

This also changes RAiDER to produce a [PEP440](https://peps.python.org/pep-0440/) compliant version number dynamically based on the closest `git` tag using [`setuptools_scm`](https://github.com/pypa/setuptools_scm). 

I've updated the docker workflow to use a tag very similar to the python-reported one, so it's very easy to jump from a RAiDER version to a specific docker image, and likewise, from a commit to a specific python version and docker image.

## Details

On the backend, `setuptools_scm` uses a [`git describe --dirty --tags --long --match "*[0-9]*"`](https://git-scm.com/docs/git-describe), which is how our docker container is currently tagged, and transforms it to match PEP440. 

For example, `git describe --dirty --tags --long --match "*[0-9]*"` would return for the head of this PR:
```
v0.1.0-103-g8f53d97
```
and `setuptools_scm` will transform that to:
```
0.1.1.dev103+g8f53d97
```

That is, setuptools_scm "guesses" the next version number by updating the smallest part of the tag (`v0.1.0` --> `v0.1.1`), marks it a development release (`.dev103`) and changes the commit specifier to use a `+`. 

Importantly, if you're on a commit associated with a tag, `git describe` and `setuptools_scm` will only return that tag. 

Likewise, the docker image will now be tagged with:
```
0.1.1.dev103_g8f53d97
```
to closely match the python version (`+` is a disallowed character in docker tags).

### A note on development installs:

When you do a development/editable install:
```
python -m pip install -e .
```
the version number RAiDER will report like:
```python
import RAiDER
print(RAiDER.__version__)
```
will be *the version from when the install command was run* as RAiDER will look in the package metadata written during installation for the version number. You can always get the current computed version number like:
```
python -m setuptools_scm
```
or by simply re-running the `pip install` command (should be no harm in doing so). This should have no effect on actual working with RAiDER and is likely to be generally unnoticed.

## Type of change

## Checklist:


environment.yml
```diff
@@ -6,8 +6,7 @@
 name: RAiDER
 channels:
  - conda-forge
- - anaconda
- - defaults
+ - nodefaults
```
jhkennedy, line 9: Generally, it's best not to mix conda-forge and the defaults, so this enforces conda-forge only (which works!).

pyproject.toml
```diff
@@ -1,14 +1,60 @@
...
     "pybind11>=2.5.0"
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "RAiDER"
+authors = [
+    {name="David Bekaert", email="david.bekaert@jpl.nasa.gov"},
+    {name="Jeremy Maurer", email="maurer.jeremy@gmail.com"},
+    {name="Piyush Agram", email="piyush@descarteslabs.com"}
```
dbekaert, line 17:
```suggestion
    {name="Piyush Agram", email="piyush@descarteslabs.com"},
    {name="Simran Sangha", email="simran.s.sangha@jpl.nasa.gov"}    
```

jhkennedy:
Docker build is working but it doesn't have permission to push it the registry since I opened this PR from a fork.

jlmaurer:
@jhkennedy how do version increments work for major releases or when the counter reaches a threshold (e.g. 0.9 -> ?, 0.1.9 -> ?, etc)?

jhkennedy:
> @jhkennedy how do version increments work for major releases or when the counter reaches a threshold (e.g. 0.9 -> ?, 0.1.9 -> ?, etc)?

@jlmaurer I hope I'm answering your question correctly; let me know if not!

For *this* PR, `setuptools_scm` will always increment the smallest part of the version number. So, since RAiDER is doing [Semantic Versioning](https://semver.org/) (`v<MAJOR>.<MINOR>.<PATCH>`), it'll always increment the patch part of the version number for development versions (that is, for commits not tagged with a version number). There's no bound on any of the major/minor/patch numbers; it's perfectly valid to have a version like `v103.59.6754`. The change in the version number is supposed to communicate how *likely* it is to break your workflow -- major change == breaking, minor change == probably not breaking, patch change == shouldn't be breaking. 

Importantly, `setuptools_scm` doesn't do any version incrementing (except for the dev version), and you "release" a new version by tagging a commit with a new version:
```
git tag -a v0.2.0 -m "Release v0.2.0" b0a5c5c
git push --tags
```
If you're on `b0a5c5c` (that is `HEAD` points to `b0a5c5c`), `setuptools_scm` will report a version number of `v0.2.0` and then subsequent commits will follow this pattern: `v0.2.1.devN+gHASH`.

In a follow-on PR, I'll set up some CI/CD that will *do* the release process for any merge to `main` so that doesn't have to be done manually, including picking the correct next version number and creating the tag.

jlmaurer:
@jhkennedy thanks that's exactly what I was wondering! Sounds good.

jlmaurer (approved):
LGTM

jlmaurer:
Pinging @dbekaert if he wants to review.

dbekaert (approved):
LGTM

PR #386, opened 2022-11-10
Support for HRRR

piyushrpt:
## Description
- Support for HRRR model

## Motivation and Context
- Get HRRR working with ray tracing

## How Has This Been Tested?
- Download tested
- Subsetting tested
- Projection interpretation tested

### DIFF
--- a/tools/RAiDER/checkArgs.py
+++ b/tools/RAiDER/checkArgs.py
@@ -57,7 +57,7 @@ def checkArgs(args):
             wetFilename, hydroFilename = makeDelayFileNames(
                 d,
                 args.los,
-                'nc',
+                args.raster_format,
                 args.weather_model._dataset.upper(),
                 args.output_directory,
             )
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -219,6 +219,9 @@ def tropo_delay_cube(dt, wf, args, model_file=None):
     except CRSError:
         raise ValueError('output_projection argument is not a valid CRS specifier')
 
+    # For testing multiprocessing
+    # TODO - move this to configuration
+    nproc = 1
 
     # logging
     logger.debug('Starting to run the weather model cube calculation')
@@ -310,14 +313,26 @@ def tropo_delay_cube(dt, wf, args, model_file=None):
             )
 
         # Get pointwise interpolators
-        ifWet, ifHydro = getInterpolators(weather_model_file, "pointwise")
+        ifWet, ifHydro = getInterpolators(weather_model_file,
+                                          kind="pointwise",
+                                          shared=(nproc > 1))
 
         # Build cube
-        wetDelay, hydroDelay = build_cube_ray(
-            xpts, ypts, zpts,
-            dt, args["los"]._file, args["look_dir"],
-            wm_proj, crs,
-            [ifWet, ifHydro])
+        if nproc == 1:
+            wetDelay, hydroDelay = build_cube_ray(
+                xpts, ypts, zpts,
+                dt, args["los"]._file, args["look_dir"],
+                wm_proj, crs,
+                [ifWet, ifHydro])
+
+        ### Use multi-processing here
+        else:
+            # Pre-build output arrays
+
+            # Create worker pool
+
+            # Loop over heights
+            print("Testing")
 
     # Write output file
     # Modify this as needed for NISAR / other projects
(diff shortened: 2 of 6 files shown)


dbekaert:
@piyushrpt How are we currently validating ray-tracing. Comparing zenith projection to slant with slant directly? or providing zenith unit vector compared with zenith delay option?

piyushrpt:
Here is the example input file that I'm testing with right now:
```

look_dir: right

date_group:
  date_list:  [20181113]

time_group:
  time: "23:00:00"

weather_model: HRRR

aoi_group:
  bounding_box: 35.4 38.2 -76.6 -74.1

height_group:
  use_dem_latlon: False

los_group:
  ray_trace: False  # Use projected slant delay by default
  zref: 60000. # Integration height. Only used in raytracing.
    # orbit_file: geom/S1A_OPER_AUX_POEORB_OPOD_20181203T120749_V20181112T225942_20181114T005942.EOF

runtime_group:
  verbose: True
  output_directory: current/HR_HRRR
  weather_model_directory: current/HR_HRRR/weather_files
  cube_spacing_in_m: 2000.
  output_projection:  32611
```

Things look ok - we have reconciled reading HRRR with ERA5. But I seem to be encountering a whole bunch of nans here:
https://github.com/dbekaert/RAiDER/blob/1585569ac25f4002cee23ecdc0786d33e9f4beee/tools/RAiDER/models/weatherModel.py#L558

All interpolate calls in that function receive valid data but return only NaNs. If there are assumptions being made about arrays being contiguous etc this makes sense. Extra set of eyes familiar with code here would help.

piyushrpt:
Numerically, this should work. Things to check  include
1. Use of correct constants for HRRR
2. Correct interpretation of units for pressure, temperature and geopotential height etc

Hydrostatic  delay seems high but I suspect its because of units - possibly acceleration in geopotential height.

tools/RAiDER/delay.py
```diff
@@ -443,6 +463,17 @@ def build_cube(xpts, ypts, zpts, model_crs, pts_crs, interpolators):
     zmin = min(interpolators[0].grid[2])
     zmax = max(interpolators[0].grid[2])
 
+    # print("Output grid: ")
```
jlmaurer, line 466: Do these next few lines need to be deleted?

tools/RAiDER/utilFcns.py
```diff
@@ -844,33 +844,33 @@ def write2NETCDF4core(nc_outfile, dimension_dict, dataset_dict, tran, mapping_na
...
-        var.setncattr('spatial_proj4', crs.to_proj4())
-        var.setncattr('spatial_epsg', epsg)
-        var.setncattr('GeoTransform', ' '.join(str(x) for x in tran))  # note this has pixel size in it - set  explicitly above
+    # variable made, now add attributes
+    for k, v in crs.to_cf().items():
+        var.setncattr(k, v)
 
-    else:
-        raise Exception('Grid mapping name not supported; currently, only WGS84 (EPSG: 4326) is supported!')
+    # These might not be needed as to_cf adds a crs_wkt
```
jlmaurer, line 869: Does this need maintained or is it verified that we don't need the commented lines?

tools/RAiDER/checkArgs.py
```diff
@@ -57,7 +57,7 @@ def checkArgs(args):
             wetFilename, hydroFilename = makeDelayFileNames(
                 d,
                 args.los,
-                'nc',
+                args.raster_format,
```
jlmaurer, line 60: Maybe we should just call this 'format' and leave it at that. :) Can do in a different pr.

jlmaurer (commented):
Just a few comments, looks great!

jlmaurer (approved):
I'll go ahead and approve since they are pretty minor comments

tools/RAiDER/delay.py
piyushrpt: There is no debugging mechanism for now. Will leave this in there if someone wants to test this out when adding new projection systems for new models

tools/RAiDER/utilFcns.py
piyushrpt: These can go

tools/RAiDER/checkArgs.py
piyushrpt: Works for me. `output_format` just to be explicit. I'm still testing against the 2D golden datasets. So would be good to not have this hardcoded

jlmaurer (approved):

Issue #373, opened 2022-11-07
[BUG] Handle dateline crossing via `RAiDER.llreader.AOI`  and interpolators


piyushrpt:
**Describe the bug**
Providing bounding boxes or lat/lon files that cross the dateline can break model loading and other interpolation operations. 
Same can happen with interpolators if the underlying projection system of the weather models is `EPSG:4326` 

**Additional context**
It might be good to have logic to determine if AOI crosses dateline and carry it as a flag in the `AOI` object that is passed around in the workflow. 

Anything involving lat/lon should check for range in [-180,180) or [0, 360) depending on what the weather model uses.

PR #389, opened 2022-11-11
Fix buffer raytracing

### LINKED ISSUE #374: [BUG] Determine extent of padding based on workflow
**Describe the bug**
Raytracing approach might require significantly more padding that computing zenith delays or projecting zenith delays using LOS angles.  This padding extent depends on the highest level of weather model under consideration as well as orientation - looking from west / east etc. 

**Additional context**
1. It might be simplest to ignore east/west differences and pad based on highest altitude projected by inc angle
2. The amount of padding (back of envelope calculation):

`paddinginm = Hmax  / cos(incmax)`

Hmax is highest level of weather model and incmax is maximum incidence angle (say 50 degs for most sensors)

`paddinginlon = paddinginm / cos(maxlataoi)` 

Lon spacing changes as function of latitude and you will need to span more values in lon as you go further away from equator.

### LINKED ISSUE #387: [BUG] Confirming Edge Effects
**Describe the bug**
There are edge effects when using the ray-tracing. This may be correct but might be worth explaining here to non-expert. Or it might have to do with interpolating 0s outside of the area of interest. Not sure.

Here is a screenshot of hydro @ 5000 meters in QGIS. All levels have this effect (see product below).

image

image

Here is a [link](https://raider-development-tests.s3.us-west-2.amazonaws.com/ERA5_tropo_20211102T062159_std.nc) to the output cube.

**To Reproduce**
See #381 for details. Reproducing the yml file for clarity below:

```
look_dir: right

date_group:
  date_start:
  date_end:
  date_step:
  date_list:  [20211102]

time_group:
  time: "06:21:59"
  end_time:

weather_model: ERA-5

aoi_group:
  lat_file:
  lon_file:
  height_file_rdr:
  station_file:
  bounding_box: 53.0 56.0 -4.0 -1.0
  utm_zone:
  grid_x:
  grid_y:

height_group:
  dem:
  use_dem_latlon: False
  height_file_rdr:

los_group:
  ray_trace: True  # Use projected slant delay by default
  zref: # Integration height. Only used in raytracing.
  los_file:
  los_cube:
  orbit_file: S1B_OPER_AUX_POEORB_OPOD_20211122T112354_V20211101T225942_20211103T005942.EOF

runtime_group:
  verbose: True
  output_directory: .
  weather_model_directory: weather_files
  cube_spacing_in_m: 2000.
  output_projection:  4326
```

jlmaurer:
- Fixes #387 
- Fixes #374
- Adds a buffer to the weather model *only* when raytracing is requested. Uses 1 degree of buffering as it seemed like that would be simplest, always work, and not add too much overhead. Confirmed that it fixes the edge-effects problem seen by @cmarshak . 
- Also it looks like conventional slant delay is never used with a cube. I added a `NotImplementedError` when ray_trace is set to `False` and changed the file naming convention so that if you run all three you'll get unique names. @piyushrpt can the conventional STD be added easily or shall we drop it as an option on the cube? pinging @dbekaert on this as well. 
- Also fixes an unrelated bug in rio_open that returned a single array instead of list of arrays when opening multi-band rasters. 

## Description

- Confirmed ZTD and raytracing work for ERA-5 without edge artifacts.

Current: 
Screen Shot 2022-11-11 at 2 02 20 PM

After implementing raytracing buffer:
Screen Shot 2022-11-11 at 1 54 01 PM

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/llreader.py
+++ b/tools/RAiDER/llreader.py
@@ -32,19 +32,17 @@ class AOI(object):
     def projection(self):
         return self._proj
     
-    def update_buffer(self, los):
+    def add_buffer(self, buffer):
         '''
         Check whether an extra lat/lon buffer is needed for raytracing
         '''
-        breakpoint()
-        if los.ray_trace():
-            # if raytracing, add a 1-degree buffer all around
-            ll_bounds = self._bounding_box
-            ll_bounds[0] = np.max(ll_bounds[0] - 1, -90)
-            ll_bounds[1] = np.max(ll_bounds[1] + 1,  90)
-            ll_bounds[2] = np.max(ll_bounds[2] - 1,-180)
-            ll_bounds[3] = np.max(ll_bounds[3] + 1, 180)
-            self._bounding_box = ll_bounds
+        # if raytracing, add a 1-degree buffer all around
+        ll_bounds = self._bounding_box.copy()
+        ll_bounds[0] = np.max([ll_bounds[0] - buffer, -90])
+        ll_bounds[1] = np.min([ll_bounds[1] + buffer,  90])
+        ll_bounds[2] = np.max([ll_bounds[2] - buffer,-180])
+        ll_bounds[3] = np.min([ll_bounds[3] + buffer, 180])
+        return ll_bounds
 
 
 
--- a/tools/RAiDER/llreader.py
+++ b/tools/RAiDER/llreader.py
@@ -32,6 +32,23 @@ class AOI(object):
     def projection(self):
         return self._proj
     
+    def update_buffer(self, los):
+        '''
+        Check whether an extra lat/lon buffer is needed for raytracing
+        '''
+        breakpoint()
+        if los.ray_trace():
+            # if raytracing, add a 1-degree buffer all around
+            ll_bounds = self._bounding_box
+            ll_bounds[0] = np.max(ll_bounds[0] - 1, -90)
+            ll_bounds[1] = np.max(ll_bounds[1] + 1,  90)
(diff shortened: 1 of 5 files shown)


piyushrpt:
Reason we hadn't implemented conventional for cubes is that ARIA/NISAR products already have the LOS cube in there and that could be directly used with the zenith cubes. It can be easily added if needed with the `state_to_los` function:
https://github.com/dbekaert/RAiDER/blob/ce420af23f4bf67ab1043e7cea30ece4e3d03404/tools/RAiDER/losreader.py#L459

And just calling it with llh corresponding to targets in a loop for each height layer

jlmaurer:
@jhkennedy is the Docker build failure is just due to a permission issue that only certain people can build on that instance?

tools/RAiDER/llreader.py
```diff
@@ -32,6 +32,21 @@ def bounds(self):
     def projection(self):
         return self._proj
     
+    def add_buffer(self, buffer):
```
piyushrpt, line 35: One thing which is a pain, maybe for future is that adding a buffer might take polygons over the dateline even if they were not in the first place. In general, we need logic to handle the dateline

piyushrpt (approved):
Looks good to me. We still have #373 to deal with in general

jhkennedy:
@jlmaurer The docker image built fine, and just fails when pushing because forks don't have permission to push. 
https://github.com/dbekaert/RAiDER/actions/runs/3447662718/jobs/5753951334#step:7:842

I could break out the push part and have the push not run for forks so we get a green checkmark

PR #394, opened 2022-11-17
add jupyter to conda install

bbuzz31:
- add jupyter to install


jlmaurer (approved):
@bbuzz31 lgtm

bbuzz31:
Note that circleci fails for an unrelated reason.

It can't seem to get the statistics from either the lat.rdr or lon.rdr files ... this works locally so i suspect it's related to the paths in the GEOM_DIR variable

PR #401, opened 2022-11-19
begin support for cli interface to dlGNSS and specific steps for delay

bbuzz31:
start a branch for downloading GNSS with config interface
- also starts support for specific steps in delay.py


bbuzz31:
this is a bit hacky but works.

to download GPS stations in some region (given in a config file), run `raiderDelay.py CFG --dostep download_gnss`

I tested this over LA and the downloaded stations (gnssStationList_overbbox.csv) match the sample file in the notebook repo (sample_gnss_list.csv).

jlmaurer (approved):
@bbuzz31 looks good to me!

jlmaurer:
After merging we can make the further tweaks, like adding parameters to the config file etc.

PR #402, opened 2022-11-22
support reading the AOI options

bbuzz31:
start to enable raider to use the geocoded file option in the aoi group (i.e., get lat/lons from a geocoded file)
had to fix the order of the coordinates and specific handling of GUNWs (because of the nested group netcdf layers)

this also works for an SRTM DEM
this also works for reading heights now at DEM levels


dbekaert:
@bbuzz31 There is a mixture here from gnss and GUNW.

tools/RAiDER/cli/raiderDelay.py
```diff
@@ -244,7 +242,11 @@ def read_template_file(fname):
         if key == 'date_group':
             template['date_list'] = parse_dates(AttributeDict(value))
         if key == 'aoi_group':
-            template['aoi'] = get_query_region(AttributeDict(value))
+            ## in case a DEM is passed and should be used
```
jlmaurer, line 245: @bbuzz31 I'm wondering if we should flatten the parameter dict once it gets read. That way we don't have to worry about groups. It's complicated since some things depend on other things. Is that what you're doing here?

tools/RAiDER/cli/raiderDelay.py
```diff
@@ -190,11 +193,6 @@ def read_inps2run_steps(inps, step_list):
 
         run_steps = step_list[idx0:idx1]
 
-    # empty the step list if:
```
jlmaurer, line 193: Does this work ok? I.e. if you pass raiderDelay.py -g you should print the config then exit.

tools/RAiDER/cli/raiderGNSS_dl.py
```diff
@@ -0,0 +1,168 @@
+import argparse
```
jlmaurer, line 1: Here's where I think the unified CLI helps us a lot. We use the same CLI for everything, and then just call the step needed, similar to what you have working now.

tools/RAiDER/cli/validators.py
```diff
@@ -121,20 +126,18 @@ def get_query_region(args):
             raise ValueError('Lats are out of N/S bounds; are your lat/lon coordinates switched? Should be SNWE')
         query = BoundingBox(bbox)
 
-    #TODO the next two options won't be reached currently because they are not in the aoi_group
```
jlmaurer, line 124: Here's a difference to what I was thinking originally and the new. For "geocoded file" I was not htinking GUNW but a geocoded DEM or LOS file. You could pass a single geocoded DEM and get all the AOI information that is needed.

tools/RAiDER/utilFcns.py
```diff
@@ -136,27 +149,28 @@ def rio_open(fname, returnProj=False, userNDV=None, band=None):
         # If user requests a band
         if band is not None:
             ndv = nodata[band - 1]
-            data = src.read(band)
+            data = src.read(band).squeeze()
```
jlmaurer, line 152: Is this needed for getting rid of 1-D axes? E.g. [1,20,20] --> [20,20]?

tools/RAiDER/utilFcns.py
```diff
@@ -165,20 +179,24 @@ def nodataToNan(inarr, listofvals):
...
+
     if os.path.exists(fname + '.vrt'):
         fname = fname + '.vrt'
 
     # Turn off PAM to avoid creating .aux.xml files
     with rasterio.Env(GDAL_PAM_ENABLED="NO"):
         with rasterio.open(fname) as src:
-            gt = src.transform.to_gdal()
-            proj = src.crs
+            gt    = src.transform.to_gdal()
```
jlmaurer, line 198: Seems like uneeded extra space?

tools/RAiDER/cli/raiderDelay.py
bbuzz31: that's not a bad idea. I guess would just have to make sure there aren't duplicate parameters.

And yeah this case is an ugly hack around for the use case when someone specifies a bounding box and use dem lat lon. 

Another option would be to move the use_dem param to the AOI group (possibly should do this anyway)

tools/RAiDER/cli/raiderDelay.py
bbuzz31: Yes, this is redundant. we already check for this on lines 134

tools/RAiDER/cli/raiderGNSS_dl.py
bbuzz31: Ah, I did not mean to add this file into the push. It actually does nothing. I incorporated the dlGNSS step into the runProgram.py / raiderDelay step which are in effect the start of a unified CLI

tools/RAiDER/cli/validators.py
bbuzz31: now you can add a DEM or a GUNW :D 
I think it should work for an LOS file too

tools/RAiDER/utilFcns.py
bbuzz31: yes exactly. and it raises problems down the line without it because it thinks [1,20,20] is a 3D dataset when it's not

tools/RAiDER/utilFcns.py
bbuzz31: just a stylistic choice

I prefer (ok the idea is that tehy're all aligned in my script... markup is moving them here):
            gt     = src.transform.to_gdal()
            proj  = src.crs
            stats = src.statistics(band)
to 
            gt = src.transform.to_gdal()
            proj = src.crs
            stats = src.statistics(band)

jlmaurer (approved):

Issue #403, opened 2022-11-22
Errors using DEM


xhuang-jpl:
try to run the program using the DEM, there is an error coming out "No such driver registered: %s, b''nc". The output is the netCDF format. please see more in the attached figure. The yaml file is also attached.  Have tested all the tiff file, which are all readable by GDAL. I could share the DEM etc file if you need to test. Thank you

![image](https://user-images.githubusercontent.com/118782850/203190740-85eb893a-e42e-4cd1-ad40-ce495adfa324.png)
[raiderDelay.yaml.zip](https://github.com/dbekaert/RAiDER/files/10061787/raiderDelay.yaml.zip)

jlmaurer:
@xhuang-jpl thanks for posting the yaml file! The issue is that you have specified the output file format as NETCDF, but [writing netcdfs is presently banned](https://rasterio.readthedocs.io/en/latest/api/rasterio.drivers.html) in rasterio. Try changing `raster_format` to `GTiff` for example to see if that runs. We should probably raise an error when a user passes an invalid format. 

I noticed  that your yaml file mentions a "cube," but if you give raster inputs to RAiDER it will return a raster output, not a cube. To get a NETCDF cube, you can specify a bounding box and a grid spacing (see [this example](https://github.com/dbekaert/RAiDER-docs/blob/master/notebooks/RAiDER_tutorial/example_yamls/nisar_example.yml)).

xhuang-jpl:
Thank you @jlmaurer , I have revised the format as the GTiff, seems that no issues about the driver, but there is one more error comes out, please see attached. Thanks again
<img width="897" alt="image" src="https://user-images.githubusercontent.com/118782850/203392093-4cec9996-3060-4c7e-93d1-02526ab60e2c.png">

bbuzz31:
This is fixed in my new branch: https://github.com/dbekaert/RAiDER/pull/402

xhuang-jpl:
tested, seems working. Thank you @bbuzz31. in addition, is the **los** the incidence angle? Thanks

xhuang-jpl:
must the dem be in WGS84 projection? I tested a dem in UTM projection along with its lat and lon file. Here is the error:
![image](https://user-images.githubusercontent.com/118782850/203444117-d649b000-6495-4964-9c66-8f095fa22913.png)

I guess the bounding box is calculated by the dem file instead of using the lat/lon files?

bbuzz31:
[...]

PR #408, opened 2022-11-23
return delay files in geo coords if geocoded aoi

bbuzz31:
If a user passes in a geocoded file for the AOI, the resulting wet/hydro delays will be written in those coords

Note that the `proj` input argument is not currently supported and should be removed until it's implemented (which may take some doing)

### DIFF
--- a/tools/RAiDER/utilFcns.py
+++ b/tools/RAiDER/utilFcns.py
@@ -244,7 +244,7 @@ def writeArrayToRaster(array, filename, noDataValue=0., fmt='ENVI', proj=None, g
     '''
     array_shp = np.shape(array)
     if array.ndim != 2:
-        raise RuntimeError('writeArrayToRaster: cannot write an array of shape {} to a raster image'.format(array_shp))
+        raise RuntimeError(f'writeArrayToRaster: cannot write an array of shape {array_shp} to a raster image')
 
     # Data type
     if "complex" in str(array.dtype):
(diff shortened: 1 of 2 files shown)


bbuzz31:
addresses #403. note that outputs will still be saved in WGS84

bbuzz31:
This needs to be more carefully vetted before merging as it may have unintended consequences

dbekaert:
I dont think proj input is to be kept. We added proj as output not sure why
we need it as input if we enforce the coordinate to be gdal compatible.
Jeremy thoughts?

tools/RAiDER/cli/validators.py
```diff
@@ -72,7 +72,7 @@ def get_heights(args, out, station_file, bounding_box=None):
         elif os.path.exists(args.dem):
             out['dem'] = args['dem']
             if bounding_box is not None:
-                dem_bounds = rio_extents(rio_profile(args.dem))
```
jlmaurer, line 75: Should we add a try/except here so either will work? Or is it always [0]?

tools/RAiDER/delay.py
```diff
@@ -118,7 +118,7 @@ def tropo_delay(dt, wetFilename, hydroFilename, args):
 
     ###########################################################
     # Load the downloaded model file for CRS information
-    wm_proj = rio_profile(f"netcdf:{weather_model_file}:t")["crs"]
```
jlmaurer, line 121: Same thing here as above, and below as well.

tools/RAiDER/utilFcns.py
```diff
@@ -120,7 +123,13 @@ def rio_profile(fname):
             f"{fname} does not contain geotransform information"
         )
 
-    return profile
+    elif profile['crs'].to_epsg() != 4326:
+        fname_wgs84 = rio_reproject_geocoded(fname)
```
jlmaurer, line 127: Is this the part that wasn't working to correctly reproject the outputs?

PR #409, opened 2022-11-28
Projection bounds

bbuzz31:
This adds support for reusing a weather model file that is in projected coordinates. 
I.e., if HRRR is already downloaded, and you pass a bounding box in lat/lon, it will mismatch because HRRR is in meters.

### DIFF
--- a/tools/RAiDER/models/hrrr.py
+++ b/tools/RAiDER/models/hrrr.py
@@ -269,6 +269,8 @@ class HRRR(WeatherModel):
             save_dir=Path(os.path.dirname(out)),
         )
         pf = H.download(":(SPFH|PRES|TMP|HGT):", verbose=verbose)
+        # breakpoint()
+        # ds = H.xarray('q')
 
         self._makeDataCubes(pf, out)
 
--- a/tools/RAiDER/models/hrrr.py
+++ b/tools/RAiDER/models/hrrr.py
@@ -269,8 +269,6 @@ class HRRR(WeatherModel):
             save_dir=Path(os.path.dirname(out)),
         )
         pf = H.download(":(SPFH|PRES|TMP|HGT):", verbose=verbose)
-        # breakpoint()
-        # ds = H.xarray('q')
 
         self._makeDataCubes(pf, out)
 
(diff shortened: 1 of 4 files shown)


tools/RAiDER/models/weatherModel.py
```diff
@@ -439,10 +439,16 @@ def checkContainment(self: weatherModel,
            and False otherwise.
         """
         ymin_input, ymax_input, xmin_input, xmax_input = ll_bounds
-        input_box = box(xmin_input, ymin_input, xmax_input, ymax_input)
+        input_box   = box(xmin_input, ymin_input, xmax_input, ymax_input)
         xmin, ymin, xmax, ymax = self.bbox
         weather_model_box = box(xmin, ymin, xmax, ymax)
 
+        world_box  = box(-180, -90, 180, 90)
```
jlmaurer, line 446: I don't think you need the if-statement, just transform everytime. We should just make sure to add a small buffer to the download region

tools/RAiDER/models/hrrr.py
```diff
@@ -269,6 +269,8 @@ def _download_hrrr_file(self, DATE, out, model='hrrr', product='prs', fxx=0, ver
             save_dir=Path(os.path.dirname(out)),
         )
         pf = H.download(":(SPFH|PRES|TMP|HGT):", verbose=verbose)
+        # breakpoint()
```
jlmaurer, line 272: is this needed?

tools/RAiDER/models/weatherModel.py
bbuzz31: Do all the weather models have a projection though?

tools/RAiDER/models/hrrr.py
bbuzz31: nope forgot to remove

tools/RAiDER/models/weatherModel.py
dbekaert: All models should come with projections.

bbuzz31: I think we should still have the check that the user input is in geocoordinates before converting

jlmaurer: Makes sense.

jlmaurer (approved):

jlmaurer:
I'm going ahead and merging this so I can build on it.

PR #414, opened 2022-11-30
change CLI to raider.py and fns to main

bbuzz31:
Replaces #412 

Refactor to have CLI to raider be `raider.py` rather than `raiderDelay.py` as this will be the main entry point for other processing chains going forward.

There are a couple tweaks to improve reading of config file arguments


bbuzz31:
for some reason when i `git add test_scenario_1.py` it automatically stages (as deleted) test_chunking and test_delay. not sure what effect this will have

bbuzz31:
Note the last two commits can be removed if we want to keep the `--steps` option. I don't think it's necessary as there is really only one step and workflows will later be handled by the `++process` options

bbuzz31:
This should be good to go. I've tested on all the models (with a bounding box as AOI) as well as a number of options (date/time slicing, lat/lon files, DEM file) for LA (see the optionsTesting.ipynb in the docs)

tools/RAiDER/cli/raider.py
```diff
@@ -273,3 +229,32 @@ def drop_nans(d):
...
+    # parse
+    inps = parseCMD(iargs)
+
+    # Read the template file
+    params = read_template_file(inps.customTemplateFile)
+
+    # Argument checking
+    params = checkArgs(params)
+
+    params['download_only'] = inps.download_only
```
jlmaurer, line 245: This should already be handled by the `read_template_file` function, with the default of False

tools/RAiDER/delay.py
```diff
@@ -708,3 +552,158 @@ def build_cube_ray(xpts, ypts, zpts, ref_time, orbit_file, look_dir, model_crs,
 
     if output_created_here:
         return outputArrs
+
+
+###############################################################################
+def main(dt, wetFilename, hydroFilename, args):
```
jlmaurer, line 558: It would be ideal to call this with all the arguments instead of as a dict.

jlmaurer (approved):
Looks great @bbuzz31 ! I left a few comments, most of which were not anything you changed but some changes would be nice to incorporate prior to release. If you want to go ahead and merge we can address in another PR. I'll go ahead an approve for now.

PR #421, opened 2022-12-02
temporary gdal fix

jlmaurer:
- conda updated gdal dependency to 3.6 and might be breaking our other deps
- adds gdal back to environment.yml and pins to <3.6


jhkennedy:
Closing this in favor of #422 where I believe I've got the fix going.

PR #426, opened 2022-12-03
Pulling refs/tags/v0.2.0 into dev

dbekaert:
Fast-forward of main to dev failed!

PR #427, opened 2022-12-03
Release: v0.2.0

jlmaurer:
- New RAiDER release 0.2.0
- See [changelog](https://github.com/dbekaert/RAiDER/blob/dev/CHANGELOG.md) for updates and changes.


jlmaurer:
@dbekaert @jhkennedy I created a new branch ('main_new') from the prior commit to merging last night on main. @jhkennedy does it suffice to rename main to 'main_old' and rename this branch to 'main' and then merge without squashing?

dbekaert:
I am not following. I think the pr should generally from dev to main. And
it should be from the raider dev to raider main not from a fork.

We could update our readme landing page and explain the main and dev
branches too. I noticed we have a lot of legacy branches too to can be
cleaned up.

jhkennedy:
@dbekaert there's two things to consider:
1.) the automatic versioning scheme works by finding the "closest" tag (number of commits away it is)
2.) the release workflow is intended to ensure the `dev` branch can always find the right tag

Right now, the `v0.2.0` tag is on the `main` branch, which due to the squash, is entirely disconnected from develop, so the closest tag will be `v0.1.0` still. 


If we merge #426, which attempts to connect `main` to  `dev`, the `v0.2.0` tag *still* appears to be 890 commits away as it has to go all the way back to the common parent, and then up to the squash-merge commit (I don't actually understand why it doesn't just hop across the merge and instead goes all the way around). Since `v0.1.0` was tagged on the `dev` branch, it appears only 131 commits away, and so our version scheme will pick that up. 

@jlmaurer has set `merge_new` branch to that common parent, and:
1.  deleting the current `main` branch
2. deleting the `v0.2.0` release and tag
3. renaming `main_new` to `main`
4. then redoing the release 
would clean that all up.

I can also just do it manually right now with a couple of git commands and a force push to main, which would be faster and might be cleaner.

--- 

Also, it's probably worth disabling squash/rebase merging in the repository settings:
image

---

Note: A lot of projects prefer squash merges (I find them limiting and vain, personally), and if you wanted to use them, we'd just have to rework the pipelines. Basically, you wouldn't have a `main` and `develop`, just a `main`, and the release process would be pushing a tag instead of merging a PR (or adding some way to indicate this PR to `main` is a special release PR, or just releasing a new version for every single merge).

dbekaert:
Ok got it. Sounds good on the proposed approach.

dbekaert:
@jhkennedy  The settings of the repo have been update to remove rebase and squash options

PR #428, opened 2022-12-03
add missing deps

jhkennedy:
Adding a few missing dependencies picked up by the conda-forge package dependency analysis in conda-forge/raider-feedstock#2

These all get implicitly installed via other packages, but because they are also explicitly imported, should be in the environment.


dbekaert (approved):
Thnx for catching that!

PR #431, opened 2022-12-06
initial attempt at library interface

bbuzz31:
I've changed the function calls to `main` and `tropo_delay_cube` to expose all arguments and allowing calling as a library.

It's pretty ugly.

I wonder if we should still force the library version to take a config file, as there are a lot of options.

I pushed a script `./test/test_library.py` that can be run to showcase how to use it.

### DIFF
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -16,7 +16,7 @@ from pyproj import CRS, Transformer
 from pyproj.exceptions import CRSError
 
 import isce3.ext.isce3 as isce
-from RAiDER.constants import _STEP
+from RAiDER.constants import _STEP, _ZREF
 from RAiDER.delayFcns import (
     getInterpolators,
     calculate_start_points,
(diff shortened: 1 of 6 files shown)


tools/RAiDER/delay.py
```diff
@@ -555,7 +553,14 @@ def build_cube_ray(xpts, ypts, zpts, ref_time, orbit_file, look_dir, model_crs,
...
 ###############################################################################
-def main(dt, wetFilename, hydroFilename, args):
+# def main(dt, wetFilename, hydroFilename, args, library=True):
+def main(dt, wetFilename, hydroFilename,
+            los=None, aoi=None, ray_trace=False, zref=_ZREF, los_cube=None,
+            height_levels=None, dem_file=None, use_dem_latlon=False,
+            bounding_box=None, lat_file=None, lon_file=None, hgt_file_rdr=None,
+            station_file=None, geocoded_file=None,
+            wmLoc='.', weather_model=None, cube_spacing=2000, download_only=False,
+            outformat='Gtiff', crs='EPSG:4326', look_dir='right', verbose=False):
```
cmarshak, line 563: I think this is the primary (main?) deliverable.

- `main` is a pretty generic name for function - what about something more descriptive?
- for something that is exposed to the real world, specifying types would be useful
- also the return type using `->` would be helpful
- The docstring format is not something I am familiar with - but I recommend using a formatter in your IDE

bbuzz31:
Done in #437

PR #432, opened 2022-12-06
Setup initial structure for working hooks

bbuzz31:
I've implemented the ++proc functionality for the default delay calculations and downloadGNSS.

Nothing has changed for the user. They can still run `raider.py CONFIG` or `raiderDownloadGNSS.py` ...

But now with the hooks we can do other stuff with `raider.py`, like run `raiderDownloadGNSS`.

```
(RAiDER) buzzanga@MT-311096:~/Software_InSAR/RAiDER_git$ raider.py -h
usage: runDelays [-h] [-g] [--download-only] [customTemplateFile]

Command line options for RAiDER processing. Default options can be found by running
raider.py --generate_config

Download a weather model and calculate tropospheric delays

positional arguments:
  customTemplateFile    custom template with option settings. ignored if the default smallbaselineApp.cfg is input.

options:
  -h, --help            show this help message and exit
  -g, --generate_template
                        generate default template (if it does not exist) and exit.
  --download-only       only download a weather model.

Usage examples:
raider.py -g
raider.py customTemplatefile.cfg
```

```
(RAiDER) buzzanga@MT-311096:~/Software_InSAR/RAiDER_git$ raider.py ++h
usage: raider.py [+h] [++process {runDelays,raiderDownloadGNSS}]

options:
  +h, ++help            show this help message and exit
  ++process {runDelays,raiderDownloadGNSS}
                        Select the HyP3 entrypoint to use (default: runDelays)
```

```
(RAiDER) buzzanga@MT-311096:~/Software_InSAR/RAiDER_git$ raider.py ++proc runDelays -h
usage: runDelays [-h] [-g] [--download-only] [customTemplateFile]

Command line options for RAiDER processing. Default options can be found by running
raider.py --generate_config

Download a weather model and calculate tropospheric delays

positional arguments:
  customTemplateFile    custom template with option settings. ignored if the default smallbaselineApp.cfg is input.

options:
  -h, --help            show this help message and exit
  -g, --generate_template
                        generate default template (if it does not exist) and exit.
  --download-only       only download a weather model.

Usage examples:
raider.py -g
raider.py customTemplatefile.cfg
```

PR #439, opened 2022-12-12
Update CLI to use ++ args

bbuzz31:
I've implemented the ++proc functionality for the default delay calculations, downloadGNSS, and delays from a GUNW.

Nothing has changed for the user. They can still run `raider.py CONFIG` or `raiderDownloadGNSS.py` ...

But now with the hooks we can do other stuff with `raider.py`, like run `calcDelaysARIA`.
```
raider.py ++help
usage: raider.py [+h] [++process {calcDelays,downloadGNSS,calcDelaysARIA}]

options:
  +h, ++help            show this help message and exit
  ++process {calcDelays,downloadGNSS,calcDelaysARIA}
                        Select the HyP3 entrypoint to use (default: calcDelays)
```

```
raider.py -h
usage: calcDelays [-h] [-g] [--download_only] [customTemplateFile]

Command line interface for RAiDER processing with a configure file.Default options can be found by running: raider.py --generate_config

positional arguments:
  customTemplateFile    custom template with option settings. ignored if the default smallbaselineApp.cfg is input.

options:
  -h, --help            show this help message and exit
  -g, --generate_template
                        generate default template (if it does not exist) and exit.
  --download_only       only download a weather model.

Examples of use:
         raider.py customTemplatefile.cfg
         raider.py -g
```

```
raider.py  ++process calcDelaysARIA -h
usage: calcDelaysARIA [-h] [-m MODEL] [-d DEM] [-o OUTPUT_DIRECTORY] [-w WRITE] files

Prepare files from ARIA-tools output to use with RAiDER

positional arguments:
  files                 ARIA GUNW netcdf files (accepts single file and wildcard matching)

options:
  -h, --help            show this help message and exit
  -m MODEL, --model MODEL
                        Weather model (default=HRRR)
  -d DEM, --dem DEM     Custom DEM. Default: Copernicus GLO30 (downloaded on fly)
  -o OUTPUT_DIRECTORY, --output_directory OUTPUT_DIRECTORY
                        Directory to store results. Default: ./
  -w WRITE, --write WRITE
                        Optionally write the delays into the GUNW products
```

Replaces #432


bbuzz31:
Remaining ToDos:
- support downloading orbit on the fly
- support reading multiple orbit files for multiple dates / put in loop
- read in tropo delays, calculate interferometric delay, store in product
- check former

bbuzz31:
This also fixes a bugs exposed when running the notebook. Station file coordinates need to be transposed differently than other AOIs

CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) 
-and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.3.1]
+RAiDER package was refactored to use a __main__ file to allow calls to different functionality.
+The default is `calcDelays` which maintains the original functionality of calling `raider.py`.
+`raider.py ++process downloadGNSS ...` can now perform the functionality of `raiderDownloadGNSS.py ...`
+`raider.py ++calcDelaysARIA GUNW` is enabled but not fully functional
```
dbekaert, line 12: There might be many differnt ARIA products. Suggest to call it it "CalcDelaysGUNW" would also make GUNW more descriptiove you mean GUNW filename.

tools/RAiDER/aria/calcARIA.py
```diff
@@ -0,0 +1,29 @@
+"""
+Calculate the interferometric phase from the 4 delays files of a GUNW
```
dbekaert, line 2: Find the 4 delays confusing to a reader.

Calculate the interfer.... phase delay at acquisition time reported in the GUNW product.

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+    return dst
+
+
+## make not get correct S1A vs S1B
+def getOrbitFile(f:str, out_dir):
+    from eof.download import download_eofs
+    orbit_dir = os.path.join(out_dir, 'orbits')
+    os.makedirs(orbit_dir, exist_ok=True)
+    # import s1reader
+    group ='science/radarMetaData/inputSLC'
```
dbekaert, line 92: I suggest you do a check on the version of the product first; akak version in filename. In the to be new product we will have fixed the eta-data. You can check with @cmarshak which version that is. that version you can just read it and you will be good.

For the other versions you could do the fix for fetching right orbit.

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+    sats  = []
+    for key in 'reference secondary'.split():
+        ds   = xr.open_dataset(f, group=f'{group}/{key}')
+        slcs = ds['L1InputGranules']
+        ## sometimes there are more than one, but the second is blank
+        ## sometimes there are more than one, but the second is same except version
+        ## error out if its not blank;
+        # assert np.count_nonzero(slcs.data) == 1, 'Cannot handle more than 1 SLC/GUNW'
+        # slc = list(filter(None, slcs))[0].item()
+        slc = slcs.data[0] # temporary hack; not going to use this anyway
```
dbekaert, line 102: I asked some more clarification from @cmarshak to confirm that there i always 1 SLC at min correct. It would be good it he could confirm the logic (e.g. i assume that there filename data and UTC we are looking for an SLC that has a near match; that one will then tell you which sensor it is. This is of course assuming that there is always a match) ,if not; suggest to fall back to my approach for doing a simple geo2rdr check on both S1A/B orbit files to do the determination.

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+
+    dates = parse_dates_GUNW(f)
+    time  = parse_time_GUNW(f)
+    dts   = [datetime.strptime(f'{dt}T{time}', '%Y%m%dT%H:%M:%S') for dt in dates]
+    paths = download_eofs(dts, sats, save_dir=orbit_dir)
+    return paths
+
+
+## only this one opens the product; need to get lats/lons actually
+def get_bbox_GUNW(f:str, buff:float=1e-5):
```
dbekaert, line 114: @jlmaurer is raider internally also adding a buffer to give at least more grid point outside the user-requested BBOX? If that is the case we dont need to add twice a buffer and the number here could stay small.

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+        cfg  = {
+               'look_dir':  lookdir,
+               'weather_model': args.model,
+               'aoi_group' : {'lat_file': f_lats, 'lon_file': f_lons},
+               'date_group': {'date_list': str(dates)},
+               'height_group': {'dem': args.dem},
+               'time_group': {'time': time},
+               'los_group' : {'ray_trace': True,
+                              'orbit_file': orbits,
+                              'los_convention': 'isce',
```
dbekaert, line 202: Do you need LOS convention when you are using an orbit file?

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+    da_lon.to_netcdf(dst_lon)
+    logger.debug('Wrote: %s', dst_lat)
+    logger.debug('Wrote: %s', dst_lon)
+    return dst_lat, dst_lon
+
+
+def getHeights(f:str):
+    ds =  xr.open_dataset(f, group='science/grids/imagingGeometry')
+    hgts = ds.heightsMeta.data.tolist()
+    return hgts
```
dbekaert, line 43: Should be a configurable parameter in the future. 
The heights needs to span the earths topographic range [min and max topography]
The heights steps should be comparable to that of the model.

2 Caveats here:
- if we want to be able to stitch in ARIA-tools across products the heigh-levels needs to be consistent. Moreover the grids needs to be consistent and integer multiple. 
- If we now consider swapping between models e.g. HRES and HRRR we should likely also have the heights consistent (should be ok as they are similar models - TBC) and the lateral should be a multiple grid increment (TBC if that can be combined in ARIA-tools to). tagging @piyushrpt @sssangha

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+from RAiDER.utilFcns import rio_open, writeArrayToRaster
+from RAiDER.logger import logger
+
+""" Should be refactored into a class that takes filename as input """
+## ---------------------------------------------------- Prepare Input from GUNW
+def makeLatLonGrid(f:str, out_dir:str):
+# def makeLatLonGrid(f:str, reg, out_dir):
+    ds0  = xr.open_dataset(f, group='science/grids/data')
+
+    Lat, Lon  = np.meshgrid(ds0.latitude.data, ds0.longitude.data)
```
dbekaert, line 24: See my comment on the heights; related also to the lonlat. 
- Grids for the tropo needs to be fixed for all products around the world such we can build a vrt of multiple cubes (GUNW products)

tools/RAiDER/aria/prepFromARIA.py
```diff
@@ -0,0 +1,212 @@
...
+    hgts = ds.heightsMeta.data.tolist()
+    return hgts
+
+
+def makeLOSFile(f:str, filename:str):
+    """ Create line-of-sight file from ARIA azimuth and incidence layers """
+
+    group   = 'science/grids/imagingGeometry'
+    azFile  = os.path.join(f'NETCDF:"{f}":{group}/azimuthAngle')
+    incFile = os.path.join(f'NETCDF:"{f}":{group}/incidenceAngle')
```
dbekaert, line 51: Can we pass somewhere at the top the lcoation of the az, inc etc.
For now if you use the orbit this should not be needed.

CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) 
-and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
dbekaert, line 7: Was this deleted by accident?

tools/RAiDER/cli/__main__.py
```diff
@@ -0,0 +1,455 @@
...
+    return
+
+
+def main():
+    parser = argparse.ArgumentParser(prefix_chars='+',
+                        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '++process', choices=['calcDelays', 'downloadGNSS', 'calcDelaysARIA'],
+                     default='calcDelays',
+                     help='Select the HyP3 entrypoint to use'
```
dbekaert, line 442: No need for the HYP3 entry point statement; this is not HYP3.

tools/RAiDER/cli/__main__.py
```diff
@@ -0,0 +1,455 @@
...
+    # ARIA_calc(tropoDelayFile, args.write)
+
+    return
+
+
+def main():
+    parser = argparse.ArgumentParser(prefix_chars='+',
+                        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '++process', choices=['calcDelays', 'downloadGNSS', 'calcDelaysARIA'],
```
dbekaert, line 440: See comment on CalcDelelaysARIA. 
Would be good to split the CalcDelays also in a download and load component.
At min it should be easy to split the calcDelay from the pre-process model as its already exposed in the function.

tools/RAiDER/cli/__main__.py
```diff
@@ -0,0 +1,455 @@
...
+    p.add_argument(
+        '-m', '--model', default='ERA5', type=str,
+        help='Weather model (default=HRRR)'
+        )
+
+    p.add_argument(
+        '-d', '--dem', default='', type=str,
+        help='Custom DEM. Default: Copernicus GLO30 (downloaded on fly)'
+        )
+
```
dbekaert, line 375: technically no DEM is needed for the 3D cube to be made.

tools/RAiDER/cli/__main__.py
```diff
@@ -0,0 +1,455 @@
...
+    p.add_argument(
+        '-o', '--output_directory', default=os.getcwd(), type=str,
+        help='Directory to store results. Default: ./'
+        )
+
+    p.add_argument(
+        '-w', '--write', default=True,
+        help='Optionally write the delays into the GUNW products'
+        )
```
dbekaert, line 384: We should think about the location we write it into the file. Currently there is a mixture of hard-coded loctions in the code.

tools/RAiDER/aria/prepFromARIA.py
jlmaurer: Raider adds the needed buffer if using ray-tracing, although we need to check whether the buffer is sufficient at high latitudes.

tools/RAiDER/aria/prepFromARIA.py
jlmaurer: `los_convention` shouldn't be needed with orbit.

tools/RAiDER/aria/calcARIA.py
```diff
@@ -0,0 +1,29 @@
...
+    tot   = wet + hydro
+    return tot
+
+
+def main(delayFiles, update_nc):
+    """ Pull the wet and hydrostatic delays from the netcdf
+
+    Calculate the interferometric phase delay
+    Write to disk, and optionally update the netcdf
+    """
```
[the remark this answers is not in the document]
bbuzz31: these function should have just been labeled as placemarkers. They're built out in #441

tools/RAiDER/aria/calcARIA.py
bbuzz31, line 2: Will make sure its clear in #441

bbuzz31: Will make sure its clear in #441

tools/RAiDER/aria/prepFromARIA.py
jlmaurer: @bbuzz31 @dbekaert are we thinking to be able to add delay as a cube to the product or only ever the interferometric delays? If the latter I think the better way to get heights would to use the raider default in which the weather model height levels directly.

tools/RAiDER/aria/prepFromARIA.py
bbuzz31: I think we should move this discussion to #441 and just leave this branch as implementing the CLI changes

tools/RAiDER/aria/prepFromARIA.py
bbuzz31: I think we should move this discussion to #441 and just leave this branch as implementing the CLI changes

tools/RAiDER/aria/prepFromARIA.py
bbuzz31: We can but I think they should just be removed and we always use orbit file

tools/RAiDER/aria/prepFromARIA.py
jlmaurer: Looks like answer to my comment above is the cubes. We may then want to define the global grid similar to how height levels are defined for weather models.

tools/RAiDER/aria/prepFromARIA.py
jlmaurer: Yes this should not be needed with orbit

tools/RAiDER/aria/prepFromARIA.py
dbekaert: We should define our own grid and heights; per my other comment so we can ensure use of the models and mixing along product edges and boundaries.

tools/RAiDER/cli/__main__.py
jlmaurer: @bbuzz31 @dbekaert this is one of the complexities I encountered before when trying to implement the "steps" approach. If you want to break out the download/preprocess/calculate delays steps it's not very straightforward without using something like steps. Perhaps a compromise would be to allow each step as separate workflows.

tools/RAiDER/cli/__main__.py
```diff
@@ -0,0 +1,455 @@
...
+
+    p.add_argument(
+        'file', type=str,
+        help='1 ARIA GUNW netcdf file'
+        )
+
+    p.add_argument(
+        '-m', '--model', default='ERA5', type=str,
+        help='Weather model (default=HRRR)'
+        )
```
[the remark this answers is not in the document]
jlmaurer: I suggest maybe to use GMAO as default as it will run without account.

tools/RAiDER/cli/__main__.py
jlmaurer: This goes back to my question about primary product for GUNW.

CHANGELOG.md
bbuzz31: yes i put it back

tools/RAiDER/cli/__main__.py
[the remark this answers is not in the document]
bbuzz31: HRRR will eventually be the default for HYP3 so I thought that should stay. I'm just using a coarser one now to speed up testing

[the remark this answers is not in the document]
dbekaert: HRRR will be for part of the ARIA products; not all as its specific region.
I am ok to keep it at HRRR, but there needs to be a clear error our.
@jlmaurer can you confirm that its the case?

tools/RAiDER/cli/__main__.py
dbekaert: We make a 3D cube of the delays. A DEM is not needed for that; just height levels and lateral spacing.

CHANGELOG.md
bbuzz31: Done

tools/RAiDER/aria/prepFromARIA.py
bbuzz31: I think this has been clarified in the discussion #435

tools/RAiDER/aria/prepFromARIA.py
bbuzz31: I think this has been clarified in the discussion #435

bbuzz31:
Replaced with #443

tools/RAiDER/cli/__main__.py
bbuzz31: Actually with this CLI it's not too difficult to do. But I think it adds unnecessary complexity. There are still a bunch of parameters the user would need to pass, so a config file is still pretty useful.

I've set it up that calcDelay (the default) can take `--download_only` to just stop processing at the weather model. This can also be specified in the configure file.

In the future, we could easily add another ++process like 'processWM' that could just take a few optional parameter (like model and bbox) and then compute for a few default times etcc...

PR #446, opened 2022-12-15
Start Python library API documentation

jlmaurer:
- Provides a few new unit tests and an initial go at Python library API documentation
- Currently only documenting two functions, `tropo_delay` and `prepareWeatherModel`, which are the main access points in Python. 
- Adds some unit tests
- Fixed some bugs and made a few tweaks related to getting unit tests working
- Starting converting docstrings to the format needed for parsing by mkdocs

- **Introduced a breaking change in terms of the CLI:** changed "raster_format" to "file_format" since it was being used for all file types

### Screenshots
Screenshot 2022-12-15 at 4 08 26 PM

Screenshot 2022-12-15 at 4 08 36 PM

Screenshot 2022-12-15 at 4 08 45 PM


jlmaurer:
@dbekaert @jhkennedy I've got a start to documentation here using MkDocs. The easiest way to deploy is to `mkdocs gh-deploy` that will create a Github pages account directly tied to this project; not sure if I have the needed permissions from my account but after merging @dbekaert should be able to pull and then run that command after `conda install -c conda-forge mkdocs`, or I can try.

jlmaurer:
@dbekaert @jhkennedy I've got a start to documentation here using MkDocs. The easiest way to deploy is to `mkdocs gh-deploy` that will create a Github pages account directly tied to this project; not sure if I have the needed permissions from my account but after merging @dbekaert should be able to pull and then run that command after `conda install -c conda-forge mkdocs`, or I can try.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) 
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.x]
```
dbekaert, line 9: Can change this to 0.3.2

dbekaert:
@jhkennedy:
+  How much of this can be generated automatically versus us making manual pages. For example the welcome to raider would be duplication from our main README; that seems like we cannot get around that. We should probably include the sponsor information, license message as required by JPL etc.

+ the subpages of functions seems more automatic. But what should we fix to also show:
    -  Show how the functions would need to be called and what the order is of the variables.
    -  Examples calls 
   
- Also i was wondering how this could look like in future. If someone now clicks the "AOI" object would that then reference the next function with information in there. The index page to the left would have all the functions?

dbekaert (approved):
In general this looks good to me.

jhkennedy:
@jlmaurer @dbekaert 

---
> The easiest way to deploy is to mkdocs gh-deploy that will create a Github pages account directly tied to this project; not sure if I have the needed permissions from my account but after merging @dbekaert should be able to pull and then run that command after conda install -c conda-forge mkdocs, or I can try.

We'll add an action that will do this automagically.

---

> +  How much of this can be generated automatically versus us making manual pages. For example the welcome to raider would be duplication from our main README; that seems like we cannot get around that. We should probably include the sponsor information, license message as required by JPL etc.

Generally everything can be automatically generated, and if you need standard headers/footers/etc, you can extend the theme to provide them. (Easy with the material theme; likely easy with the readthedocs theme, but I'm less familiar). You can directly include the README and shouldn't need to duplicate it. 

> + the subpages of functions seems more automatic. But what should we fix to also show:
>     -  Show how the functions would need to be called and what the order is of the variables.
>     -  Examples calls 

That all can be specified in the docstring

> - Also i was wondering how this could look like in future. If someone now clicks the "AOI" object would that then reference the next function with information in there. The index page to the left would have all the functions?

Yes, you can reference other functions, and even parts of their documentation, directly: https://mkdocstrings.github.io/usage/#cross-references

---

I have a few questions overall:
1. since you have [RAiDER-docs](https://github.com/dbekaert/RAiDER-docs/) would you rather have a single website that has all that content and the API docs?
2. For the docs website, you'll typically want the main/default to be documentation for the latest release. Do you *also* need docs for develop (unreleased) things?

dbekaert:
wrt your questions:
- that would be nice, but i want to avoid the raider repo to get to large due to notebooks etc. 
 Perhaps we should update the documentation on the raider-docs page and directly impact. 
    - So have the notebooks + min description of it. We then provide the link to a readable notebook
    - We could at the same time make it also ore clear and rename raider-docs to a raider-notebook repo.
- I thought i might be nice because then you could pre-see items and enforce documentation to be added. But i am not sure how this would work?
-

PR #457, opened 2022-12-22
Support datelist

### LINKED ISSUE #456: [BUG] Multiple dates incorrectly read
To reproduce, make a config file with two dates: 
`date_group:
    date_list: (20200130, 20200124)
`

What happens is that the weather model objects gets instantiated with an empty `files` attribute so the first if block gets activated below and the weather model gets downloaded if it exists or reused.
Either way it then gets added to the `files` attribute. The delays are calculated, and the loop continues to the next date. Now the previous date is exists in `files` so the if block is skipped and the previous date is  reused to calculated delays at new date.

```
 download_flag = True
    if weather_model.files is None:
        weather_model.filename(time, wmLoc)
        if os.path.exists(weather_model.files[0]):
            if not force_download:
                logger.warning(
                    'Weather model already exists, please remove it ("%s") if you want '
                    'to download a new one.', weather_model.files
                )
                download_flag = False
    else:
        download_flag = False
```

`processWM.py`

bbuzz31:
Fixes #456 by creating a new filename from the datetime each new call to prepareWeatherModel. Implements a check to make sure that multiple dates are downloaded.

At some point, a more robust check of how the `filename` attribute is created and stored in the `WeatherModel` class should be undertaken.


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) 
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.2]
```
dbekaert, line 9: In dev we will not increment version anymore.

test/test_datelist.py
```diff
@@ -0,0 +1,47 @@
+import os
```
dbekaert, line 1: Do we have a test for all date parsing option? Date; date1-date2; date1-date2-sampling?

dbekaert:
@jlmaurer this used to be working prior to refactoring. Lets make sure we have a few test for it

jlmaurer (approved):
LGTM

dbekaert (approved):
LGTM

PR #460, opened 2022-12-22
GUNW to RAiDER to GUNW

bbuzz31:
Implement functionality to read an S1 GUNW, compute delays on a cube, write back into GUNW

Replaces #445


bbuzz31:
Here is a screen grab of the resulting layer in the GUNW:
image

bbuzz31:
Comparison of delays:

image

dbekaert:
@bbuzz31 few observations.

Assumptions on differences:
- Raider is only at nearest UTC, GACOS is temporally interpolated. GACOS is projected from zenith to slant. GACOS is HRES underneath, RAIDER is based off ERA5.

Suggestion if you would like to get it closer in terms of true comparison:
- Fake it to be at 00/06/12/18 UTC for both GACOS and RAIDER (aka no temporal interpolation).
- Use HRES for RAIDER such its the same as GACOS (aka same source model)


Observations.
1) Raider and ARIA-tools are off. Hence bug with interpolation somewhere.
    - The ARIA-tools version does not seem to have a topo dependancy
    - The ARIA-tools version is a lot smaller.
    @sssangha: 
    - Do we have same vertical units between the heights and the DEM?
    - Are the height or dimensions swapped (e.g. intersection on flipped dimension would show small magnitude without much topo dependancy)
2) The spatial extend is lacking in the raider cube but not in the aria-tools cube.
     @bbuzz31 raider cube should cover the full extend of the product. Please confirm?
3) Topo detail will be hard to see in Raider (0.1 degree grid) variant unless you intersect with a higher res DEM.

sssangha:
@bbuzz31 , I'll post some crude illustrations below to demonstrate what I observed on my end through ARIA-tools.

When taking the embedded layers at face-value (i.e. assuming their heights are in ascending order), I get values ~an order of magnitude off from the raider-derived interpolation:
Screen Shot 2022-12-23 at 2 34 02 AM

But when I flip the cube vertically (i.e. assuming instead the heights are listed in descending order), I get values that are ~the same magnitude as the raider-derived interpolation:
Screen Shot 2022-12-23 at 2 34 20 AM

bbuzz31:
Outstanding tasks to:

- Figure out spatial interpolation 
- Implement lat/lon/height levels per #444 
- Interpolate delays in time

These issues can probably be fixed in the coming days/weeks and shouldn't prevent the HYP3 team to begin testing with this PR.

CHANGELOG.md
```diff
@@ -3,18 +3,24 @@
...
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/) 
+and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0]
+Functionality to read a GUNW, write a config, calc delays, write interferometric delays
+Fixed lat/lons at 0.05º (HRRR), 0.1º (all others)
+Metadata match GUNW
```
dbekaert, line 12: Adding of new GUNW support to RAiDER. This is an interface delivery allowing for subsequent integration into HYP3 (input/output parsing is not expected to change; computed data is not yet verified). 

### New/Updated Features
+ Working GUNW entry point in workflow for raider.py
+ Ability to parse a GUNW to workflows from which all required RAiDER information is extracted (e.g. dates, UTC, orbit, bbox, look direction, wavelength) with an option to specify weather model (those already supported by RAiDER) and ability to squeeze in the derived output into the original GUNW product.
+ Delays for GUNW are calculated in RAiDER using the ray-tracing option specifying bbox (GUNW driven), a hardcoded lateral posting (0.05º for HRRR and 0.1º for others),  fixed vertical height levels, using an different orbit file for  secondary and master. 
     - The hard-coded heights and posting will be refined per model and to ensure stitching abilities in ARIA-tools.
     - The orbit should be refined to not change between secondary and reference to avoid issues. See https://github.com/dbekaert/RAiDER/discussions/435#discussioncomment-4392665 
+ Bug fix for raider.py "date" input argument when multiple dates are requested (i.e. support of requesting two dates or two dates with a sampling).  
+ Add unit test for date input argument checking (single day, two dates, two dates with samples)
+ Write the diagnostic weather model files to the 'output_directory' rather than PWD
+ Fix for incorrectly written hard-cored projection embedded in the computed output data

dbekaert: Added dependcies for:
- sentinelof: used to fetch the orbit for GUNW 
- rioxarray: xxxx

dbekaert: Should also mention that currently for development purposes we hardcode version inside gunw to be 1c.

dbekaert: ### Not implemented / supported in this release
+ no temporal interpolation 
+ no refined model specific hardcoded spacing and heights
+ no ability for single orbit Interferometric calculation 
+ no verification of results

dbekaert (approved):
@bbuzz31 It would be nice to get closer to a meaningful result, but for interfce delivery this would do.

tools/RAiDER/cli/validators.py
```diff
@@ -144,8 +148,8 @@ def get_query_region(args):
         query  = GeocodedFile(args.geocoded_file, is_dem=is_dem)
 
     ## untested
-    elif 'los_cube' in args.keys():
-        query = Geocube(args.los_cube)
+    elif 'geo_cube' in args.keys():
```
jlmaurer, line 151: Might want to make this "if args.get('geo_cube') to be consistent with the others.

tools/RAiDER/losreader.py
```diff
@@ -607,9 +607,9 @@ def get_radar_pos(llh, orb, out="lookangle"):
     num_iteration = 30
     residual_threshold = 1.0e-7
 
-    # Get xyz positions of targets here
+    # Get xyz positions of targets here from lat/lon/height
     targ_xyz = np.stack(
-        lla2ecef(llh[:, 1], llh[:, 0], llh[:, 2]), axis=-1
+        lla2ecef(llh[:, 0], llh[:, 1], llh[:, 2]), axis=-1
```
dbekaert, line 612: This flipped. Can you make sure you are not introducing a bug. @jlmaurer @piyushrpt can you review.

PR #465, opened 2023-01-10
Fix interpolation when passing lat/lon files and station file

bbuzz31:
Fix the interpolation in two places, where the points for interpolation are input (lats/lon were out of order) and use the exact DEM bounds for interpolating heights

UPDATE:
I've now implemented the exact same result using xarray, which is an order of magnitude faster (at least).


bbuzz31:
Here you can see the correct positions of the islands in the DEM vs the delays as well as the topography

image

bbuzz31:
This is marked as draft until tests are implemented for validity, a check with the station files, and existing tests are passed

test/test_intersect.py
```diff
@@ -0,0 +1,135 @@
...
+
+    with open(dst, 'w') as fh:
+        yaml.safe_dump(params, fh,  default_flow_style=False)
+
+    return dst
+
+
+def test_cube_intersect():
+    """ Test the intersection of lat/lon files with the DEM (model height levels?) """
+    TEST_DIR = '/Users/buzzanga/Software_InSAR/RAiDER_git/test'
```
jlmaurer, line 62: This won't work on circleci. :) You can import the TEST_DIR I think.

tools/RAiDER/checkArgs.py
```diff
@@ -48,7 +48,7 @@ def checkArgs(args):
                     args.output_directory,
                     '{}_Delay_{}.csv'
                     .format(
-                        args.weather_model.Model(),
```
jlmaurer, line 51: Is this the same thing?

tools/RAiDER/cli/__init__.py
```diff
@@ -36,7 +36,7 @@ class AttributeDict(dict):
             raster_format='GTiff',
             file_format='GTiff',
             download_only=False,
-            output_directory=os.getcwd(),
```
jlmaurer, line 39: Is this needed? os.getcwd() version is OS agnostic whereas I think '.' only works on *nix systems.

jlmaurer (approved):
@bbuzz31 this looks great! Great job adding all those tests. I made a few minor comments and otherwise looks good to me!

tools/RAiDER/dem.py
```diff
@@ -40,16 +40,12 @@ def getHeights(ll_bounds, dem_type, dem_file, lats=None, lons=None):
         hts = None
 
     elif (dem_type == 'download') or (dem_type == 'dem'):
-        if ~os.path.exists(dem_file):
-            download_dem(ll_bounds, writeDEM=True, outName=dem_file)
-            
-        #TODO: interpolate heights to query lats/lons
+        zvals, metadata = download_dem(ll_bounds, writeDEM=True, outName=dem_file)
```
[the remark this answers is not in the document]
bbuzz31: Possibly not, but I think at some point it may be needed (for example, if a dem_lat_lon is specified as AOI and no DEM is given?)

test/test_intersect.py
bbuzz31: You're right of course, but somehow the test passed?

tools/RAiDER/checkArgs.py
bbuzz31: For most weather models yes, but the former was 'ERA-5', the latter 'ERA5'

tools/RAiDER/cli/__init__.py
bbuzz31: I'm pretty sure pytest wouldn't allow it, but I will verify.

PR #466, opened 2023-01-11
bug fixes for latlon files

jlmaurer:
- Similar to #465 but different approach. Not 100% sure which one is better / works in all cases yet
- dem-stitcher seems to have changed how it returns the dem arrays; flipped compared to what RAiDER was expecting. Flipping the array seems to fix the problem.


bbuzz31:
I think we should stick with the [improved version that i will push soon] of my PR.
This introduces slight errors as the linspace we make will be slightly off from the actual pixel coordinates in the DEM. 
Better to use rasterio/gdal to read the coordinates directly, and then interpolate using those.

PR #467, opened 2023-01-11
Better handle package data for conda build

jhkennedy:
The current conda-forge build is missing a critical "package data" file. This PR explicitly calls out `*.yml` and `*.yaml` files inside `src/RAiDER` as package data so they are included in the archive.

NOTE: while a `MANIFEST.in` works for `pip` and `setuptools` everywhere outside of conda, conda-build would still not include the package data in the conda build. This is the only method that I could get to work. 

---

Explicitely calling out the package data, or including a `MANIFEST.in`, isn't supposed to be needed because everything is included via `setuptools_scm`. Importantly, when building the RAiDER distribution `setuptools_scm`  uses the `git` history (tags, tracked files) to define the package version and included files and produces a built src or binary distribution, which pip uses to install the package. 

when building the conda package, we're not actually cloning the repo, but instead downloading a git archive with **does not** include the `git` history. For the package version, we explicitly tell `setuptools_scm` which version to use:
https://github.com/conda-forge/raider-feedstock/blob/main/recipe/meta.yaml#L14
But we also need to explicitly tell it what files to include, which is what this PR does.

--- 

**Alternatively** we could switch to building the conda package from distribution published to PyPI, which *will* include metadata describing the version package files to include, but `pip` installs from PyPI won't work (easily at least) as we have dependencies that are conda-forge only (e.g., ISCE3) and compiled extensions (we'd need to build binary distributions for a variety of architectures). This is *not* a good user experience so we shouldn't do this unless we do the work to make installing from PyPI relatively easy.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,6 @@ jobs:
   call-version-info-workflow:
     uses: ASFHyP3/actions/.github/workflows/reusable-version-info.yml@v0.7.0
     with:
-      conda_env_name: RAiDER
       python_version: '3.10'
 
   call-docker-ghcr-workflow:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,14 +12,14 @@ on:
 
 jobs:
   call-version-info-workflow:
-    uses: ASFHyP3/actions/.github/workflows/reusable-version-info.yml@v0.6.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-version-info.yml@v0.7.0
     with:
       conda_env_name: RAiDER
       python_version: '3.10'
 
   call-docker-ghcr-workflow:
     needs: call-version-info-workflow
-    uses: ASFHyP3/actions/.github/workflows/reusable-docker-ghcr.yml@v0.6.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-docker-ghcr.yml@v0.7.0
     with:
       version_tag: ${{ needs.call-version-info-workflow.outputs.version_tag }}
       release_branch: main
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,6 +13,6 @@ on:
 
 jobs:
   call-changelog-check-workflow:
-    uses: ASFHyP3/actions/.github/workflows/reusable-changelog-check.yml@v0.6.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-changelog-check.yml@v0.7.0
     secrets:
       USER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -14,12 +14,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: mamba-org/provision-with-micromamba@v14
         with:
-          mamba-version: "*"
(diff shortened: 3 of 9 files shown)


jlmaurer:
@jhkennedy @dbekaert this seems fine to me. Will this manifest catch all package files? I didn't see a "*.yaml" or anything like so wasn't sure. 

> This is not a good user experience so we shouldn't do this unless we do the work to make installing from PyPI relatively easy

It is a bit annoying that it would be so much work to make a PyPI package, but probably I would vote against it for the moment, seems like a lot of work to be added when we are trying to get up to a stable version.

jhkennedy:
> Will this manifest catch all package files? I didn't see a "*.yaml" or anything like so wasn't sure.

@jlmaurer I'm still working out if this actually fixes it or not, but yes, the intent is to include all of it, which is what the `graft` keyword is supposed to do: https://packaging.python.org/en/latest/guides/using-manifest-in/

jhkennedy:
Okay, so, if I do this:
```shell
conda create -n tmp python=3.10 pip python-build
conda activate tmp

wget https://github.com/dbekaert/RAiDER/archive/refs/tags/v0.4.0.tar.gz
tar -zxvf v0.4.0.tar.gz

wget https://patch-diff.githubusercontent.com/raw/dbekaert/RAiDER/pull/467.patch

cd RAiDER-0.4.0
patch  < ../467.patch

export SETUPTOOLS_SCM_PRETEND_VERSION=0.4.0
python -m build

tar -tf dist/RAiDER-0.4.0.tar.gz | grep yaml
unzip -l dist/RAiDER-0.4.0-cp310-cp310-linux_x86_64.whl | grep yaml
```

We do see the yaml file included in the source and wheel distribution correctly. 

This is what conda-forge *should* be doing, but the actual `.conda` package that comes out doesn't seem to have it, so I still have a bit of investigating to do.

jhkennedy:
As described above, I'm applying this change as a patch because I'm attempting to patch the current release on conda-forge as well; see [conda-forge/raider-feedstock#7](https://github.com/conda-forge/raider-feedstock/pull/7)

To check the conda package locally, I'm doing:
```shell
git clone -b conda-manifest git@github.com:jhkennedy/raider-feedstock.git
cd raider-feedstock

./build_locally.py linux_64_numpy1.21python3.10.____cpython

cd build_artifacts/linux-64
unzip raider-0.4.0-py310h9b08913_1.conda
tar --use-compress-program=unzstd -tf pkg-raider-0.4.0-py310h9b08913_1.tar.zst | sort
```

which unfortunately still doesn't show a `lib/python3.10/site-packages/RAiDER/cli/raider.yaml` file being added to the package :sob:

jhkennedy:
Okay, after *entirely too many* iterations, I've got a setup that works for:
* `pip` installed (editable and not)
* `sdist` and `.whl` builds of the package
* `conda-build` tested locally. 

So the `raider.yaml` package data file should now be included everywhere.

jhkennedy:
The miniconda GitHub action has been having a ton of intermittent environment creation errors, causing a lot of the checks to fail. I've also:
* dropped minconda for micromamba
* bumped the reusable actions versions

which should resolve those issues. More discussion here: https://github.com/ASFHyP3/actions/pull/49

jlmaurer:
@jhkennedy that's some great work! 

One of our unit tests is pinging a server that is evidently down a lot, which is why circleCI is failing. pinging @dbekaert and @bbuzz31 just to let them know. The problem test is in test_datelist and I believe calls the GMAO server. We don't need to worry about it for this PR.

jlmaurer (approved):

PR #473, opened 2023-01-18
Prepare for Hyp3 integration

forrestfwilliams:
To support RAiDER processing of GUNW's in HyP3, we need:
* to download the GUNW from S3
* Replace the GUNW in S3 with an updated version including the tropospheric corrections

which is all handled via the new `RAiDER.aws` module (adds `boto3` as a dependency)

This PR changes the GUNW interface to:
* make `file` argument optional `--file`
* if `--file` is not provided:
  * and a `--bucket` argument is provided: download GUNW from S3
  * and a `--bucket` argument is *not* provided: throw an error
* changes the `--model` argument to `--weather-model`
* replaces `_` with `-` in arguments. E.g., `--output-directory` instead of `--output_directory` (argparse automatically casts `-` to `_` in argument variable names)
  
This PR also
* changes the Docker entrypoint from `bash` to `raider.py`
* Fixes next version number in Changelog
* removes a bunch of unused imports

### DIFF
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -397,7 +397,7 @@ def calcDelaysGUNW():
         )
 
     p.add_argument(
-        '-m', '--model', default='HRRR', type=str,
+        '-m', '--weather_model', default='HRRR', type=str,
         help='Weather model.'
         )
 
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -6,9 +6,7 @@ import yaml
 
 from textwrap import dedent
 
-import RAiDER
 from RAiDER import aws
-from RAiDER.constants import _ZREF, _CUBE_SPACING_IN_M
 from RAiDER.logger import logger, logging
 from RAiDER.cli import DEFAULT_DICT, AttributeDict
 from RAiDER.cli.parser import add_out, add_cpus, add_verbose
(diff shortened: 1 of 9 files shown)


tools/RAiDER/cli/raider.py
```diff
@@ -377,37 +378,55 @@ def downloadGNSS():
...
         '-u', '--update_GUNW', default=True,
-        help='Optionally update the GUNW by writing the delays into the troposphere group (Default=True).'
+        help='Optionally update the GUNW by writing the delays into the troposphere group.'
         )
+    
+    args       = p.parse_args()
 
+    if args.weather_model == 'None':
+        print('Nothing to do!')
+        return
```
jhkennedy, line 421:
```suggestion
    if args.weather_model == 'None':
        # NOTE: HyP3's current step function implementation does not have a good way of conditionally
        #       running processing steps. This allows HyP3 to always run this step but exit immediately
        #       and do nothing if tropospheric correction via RAiDER is not selected. This should not cause
        #       any appreciable cost increase to GUNW product generation.
        print('Nothing to do!')
        return
```

tools/RAiDER/cli/raider.py
```diff
@@ -377,37 +378,55 @@ def downloadGNSS():
...
 
-    args       = p.parse_args(args=iargs)
-    args.argv  = iargs if iargs else os.sys.argv[1:]
     # args.files = glob.glob(args.files) # eventually support multiple files
+    if args.file:
+        pass
+    elif args.bucket:
+        args.file = aws.get_s3_file(args.bucket, args.bucket_prefix, '.nc')
+    else:
+        raise ValueError('Either argument --file or --bucket must be provided')
```
jhkennedy, line 429: There's gotta be a better way to do this... :thinking:

jhkennedy (changes requested):
Looking good! There's still some work around logging, and the `boto3` needs to be added to the `environment.yml` file

tools/RAiDER/cli/raider.py
jhkennedy: @jacquelynsmale figured it out!

jhkennedy (approved):
I am able to run the GUNW entrypoint against a file in S3 both directly and with the Docker container. I think this is ready at this point.

PR #475, opened 2023-01-19
Few small fixes and checks

jlmaurer:
Adds some basic warnings, tests, and checks

## Description

- Adds a warning when the user requests negative height levels 
- Checks for nans when creating the interpolator objects
- functionalizes parts of _tropo_delay_cube* to help with readability and testing
- Adds to and updates unit tests

## How Has This Been Tested?

- unit tests pass locally

## Type of change

## Checklist:


bbuzz31 (approved):
LGTM

README.md
```diff
@@ -103,7 +103,7 @@ For development, we recommend installing directly from source.
 ```
 git clone https://github.com/dbekaert/RAiDER.git
 cd RAiDER
-conda create -f environment.yml
+conda env create -f environment.yml
```
dbekaert, line 106: Should be removed... Thats only when you are updating your environment. this fails at first pass.

bbuzz31: I think the fix (`conda env create -f environment.yml`) is correct. See: https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file

jlmaurer: I think this might be a conda version thing...I'm looking at [this website](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file). I can't seem to find the earlier documentation, but I'm pretty sure that I used `conda create -f environment.yml` before. Not sure best way to proceed?

jhkennedy: @dbekaert, @bbuzz31 is correct for these instructions since this is a list of steps from a fresh clone to a development install. You could add a section about keeping the environment+install up to date and recommend users do
```
conda env update -f environment.yml
python -m pip install -e .
```
any time they switch branches or pull. The first command is only necessary if the dependencies have changed. The second is only necessary anytime the compiled extensions change, the package structure changes, or the entry points change. But it (should)  never hurt to do both at any point, so there's no harm in aggressively recommending it.

--

@jlmaurer conda has a messy API and pretty meh docs, so you might be confusing a couple of the commands:
* `conda create` accepts a `--file` argument (but doesn't have a `-f` short name for it!)  for `conda_spec.txt`/`requirements.txt` type files
  * can be `@EXPLICIT` links to package archive files
  * can be "normal" requirements specs (e.g., `isce3>=0.9`)
  * cannot be a mix of explicit and normal specs
* `conda env create` accepts a `-f`/`--file` argument for environment definition yaml files

Also, to make things even more fun, `micromamba` decided to "clean up" the conda API and does allow:
```
micromamba create -f environment.yml
```

Another fun thing: if you do `conda env update -f environemnt.yml` and the environment doesn't exist already, it will create it. So you can *always* do `conda env update -f environment.yml` and it's typically good practice to do it anytime you pull or switch branches in case there are dependency changes. 

---

tl;dr: `conda env create -f environment.yml` is correct. `conda env update -f environment.yml` also works.

dbekaert:
@jlmaurer negative height level warning or when the height level is below that of the model?

PR #477, opened 2023-01-23
Release v0.4.1

jhkennedy:
Adding RAiDER to the ARIA-HyP3 pipeline MVP is ready -- if we release this we can deploy the ARIA-HyP3 deployments (ASFHyP3/hyp3#1435) with a RAiDER step for further feedback.

---
### New/Updated Features
+ Reorder target points for intersection
+ Use exact coordinates of DEM to interpolate heights to target lat/lons
+ Support DEM interpolation to station file
+ Implement end-to-end test for intersection of cube with lat/lon files
+ Implement end-to-end test for calculation at stations delay
+ Update AOI to store the output directory so DEM is written to right place
+ `calcDelaysGUNW` will optionally download a GUNW product from AWS S3, process it, and upload a new version of the GUNW product in-place with the tropospheric correction layers added so that RAiDER can be used in ARIA GUNW production via HyP3 

### Fixed
+ Package data is more explicitly handled so that it is included in the conda-forge build; see [#467](https://github.com/dbekaert/RAiDER/pull/467)


CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [0.4.1]
```
dbekaert, line 10: @jhkennedy Can we add a note that the -gunw workflow we are still verifying the results?

dbekaert (approved):
LGTM @jhkennedy could we add a small update to the change log that we are still verifying correctness of the GUNW workflow

jhkennedy:
@dbekaert good idea! Will do!

CHANGELOG.md
jhkennedy: @dbekaert good idea! Will do!

PR #479, opened 2023-01-24
Interpolate weather model

bbuzz31:
Prototype for using a weighted mean to interpolate an user passed time between weather model interals (i.e., if you ask for 13:55 in GMAO, it will compute weighted mean of 12:00 and 15:00 of the GMAO delays.

Need to still implement a test (which is why this is draft status)

Only delays on a cube (not intersected delays) are currently supported.


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### New/Updated Features
++ Support temporal interpolation when user requested time is not exactly on a weather model time
```
dbekaert, line 10: Are you allowing for an input option where this is done? aka it should be configurable whether or not to do temporal interpolation or jumping to nearest time.

I did not see in this commit below where the function is called within raider itself.

bbuzz31: It was previously kept separate via a new entry point. It's now embedded directly in RAiDER.py:calcDelay

dbekaert:
@jlmaurer please review and verify

bbuzz31:
Replaced with #499

PR #480, opened 2023-01-24
add ERA5 as a possible weather model option

forrestfwilliams:
Change the `weather-model` option to accept `ERA5`

### DIFF
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -399,7 +399,7 @@ def calcDelaysGUNW():
 
     p.add_argument(
         '-m', '--weather-model', default='HRRR', type=str,
-        choices=['None', 'HRRR', 'HRES', 'GMAO', 'ERA5'], help='Weather model.'
+        choices=['None'] + ALLOWED_MODELS, help='Weather model.'
     )
 
     p.add_argument(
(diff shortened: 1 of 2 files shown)


tools/RAiDER/cli/raider.py
```diff
@@ -399,7 +399,7 @@ def calcDelaysGUNW():
 
     p.add_argument(
         '-m', '--weather-model', default='HRRR', type=str,
-        choices=['None', 'HRRR', 'HRES', 'GMAO'], help='Weather model.'
+        choices=['None', 'HRRR', 'HRES', 'GMAO', 'ERA5'], help='Weather model.'
```
jhkennedy, line 402:
```suggestion
        choices=['None'] + ALLOWED_MODELS, help='Weather model.'
```

Might be better to change this to the `ALLOWED_MODELS` constant from RAiDER:
https://github.com/dbekaert/RAiDER/blob/develop/tools/RAiDER/models/allowed.py#L1

**Note:** will also need to add 
```python
from RAiDER.models import ALLOWED_MODELS
``` 
to this file

cmarshak:
Very nice!

cmarshak (approved):

jhkennedy:
Docker failure is expected -- forks can't push.

CicleCI failure is `webob.exc.HTTPError: 503 Service Temporarily Unavailable` for one test; but the rest pass. So, :shipit:

Issue #483, opened 2023-02-01
[Feature/Enhancement] Use ENV Variables for HRES and ERA5 credentials in GUNW entrypoint


cmarshak:
The new secrets for the new weather models will need to be consumed in Raider this way.
 
@bbuzz31, this is too big a lift in the 24 hours before you go. I think this will go into the GUNW script here: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/aria/prepFromGUNW.py#L231. Please provide guidance (maybe with JMaurer) where the best place to include this logic would be.
 
For HRES and ERA5, we need a way to inject env variables into the appropriate hidden files.
 
Here is how it’s done in the GUNW plugin: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/__main__.py#L53-L82 (written by Joe – note the tests: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/tests/test_main.py#L6-L73)
 
The environment variables would then write to the hidden files for either ERA5 or HRES as documented nicely here: https://github.com/dbekaert/RAiDER/blob/dev/docs/WeatherModels.md

dbekaert:
This has been added in #484

Issue #481, opened 2023-01-31
Ensure that the correct number of files are written out by RAiDER for GUNW


bbuzz31:
If a user (or HYP3) requests a date/time for which the weather model doesn't exist, an error is raised and caught and the raider processing continues. We might want to put into raider:calcDelaysGUNW a catch to make sure that both delay files have been successfully calculated.

@sssangha good catch raising this!

bbuzz31:
@forrestfwilliams do you mind if I add a quick check on the number of files produced to your #480?

forrestfwilliams:
Yep, @bbuzz31 feel free to add it!

bbuzz31:
done in #482

PR #484, opened 2023-02-02
calcDelaysGUNW - added api credentials check and writing locally [Issue: #483]

mgovorcin:
Added option to add weather model API credentials and write it locally to a hidden file 
Issue: [#481](https://github.com/dbekaert/RAiDER/issues/483)

## Description
 added credentials.py to RAiDER.models with function check_api that check if env API credential file already exists, if not it looks for api: UID and KEY  in args and writes it locally. Added options to raider.py **process calcDelysGUNW to add api key and uid 

Added args parms to calcDelaysGUNW: --api_uid, --api_key
```
raider.py ++process calcDelaysGUNW -finput_file -m model -o output --api_uid UID --api_key KEY
```
NOTE: if API_RC hidden file exists such as ~/.cdsapirc, inserted UID and KEY will be ignored, There is an update flag that overwrites hidden file, but for calcGUNWDelays is set to be False. It can be added ``` if api_key is not None and api_uid is not None: update_flag=True``` . This is will give priority to user input?

Another option to add API credentials is through ENV variable, if API_UID and API_KEY are None. **Credentials.py** searches for the env and use them to populate API_RC hidden file: List of environmental API variables:

cdsapirc : models(ERA5, ERA5T):
    $RAIDER_ECMWF_ERA5_API_KEY
    $RAIDER_ECMWF_ERA5_UID

ecmwfapirc : models(HRES, ERAI):
    $RAIDER_HRES_URL
    $RAIDER_HRES_API_KEY
    $RAIDER_HRES_EMAIL

netrc : models (GMAO)
     $EARTHDATA_USERNAME
     $EARTHDATA_PASSWORD

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/aria/prepFromGUNW.py
+++ b/tools/RAiDER/aria/prepFromGUNW.py
@@ -15,6 +15,7 @@ import shapely.wkt
 import RAiDER
 from RAiDER.utilFcns import rio_open, writeArrayToRaster
 from RAiDER.logger import logger
+from RAiDER.models import credentials
 from eof.download import download_eofs
 
 ## cube spacing in degrees for each model
@@ -230,6 +231,10 @@ def update_yaml(dct_cfg:dict, dst:str='GUNW.yaml'):
 
 def main(args):
     """ Read parameters needed for RAiDER from ARIA Standard Products (GUNW) """
+
+    # Check if WEATHER MODEL API credentials exists, if not create them or raise ERROR
+    credentials.check_api(args.weather_model, args.api_uid, args.api_key,
+                          prompt_flag=False, update_flag=False)
 
     GUNWObj = GUNW(args.file, args.weather_model, args.output_directory)
     GUNWObj()
--- a/tools/RAiDER/aria/prepFromGUNW.py
+++ b/tools/RAiDER/aria/prepFromGUNW.py
@@ -232,7 +232,7 @@ def update_yaml(dct_cfg:dict, dst:str='GUNW.yaml'):
 def main(args):
     """ Read parameters needed for RAiDER from ARIA Standard Products (GUNW) """
 
-    # Check if WEATHER MODEL API credentials exists, if not create them or raise ERROR
+    # Check if WEATHER MODEL API credentials hidden file exists, if not create it or raise ERROR
     credentials.check_api(args.weather_model, args.api_uid, args.api_key,
                           prompt_flag=False, update_flag=False)
 
--- a/tools/RAiDER/aria/prepFromGUNW.py
+++ b/tools/RAiDER/aria/prepFromGUNW.py
@@ -233,8 +233,7 @@ def main(args):
     """ Read parameters needed for RAiDER from ARIA Standard Products (GUNW) """
 
     # Check if WEATHER MODEL API credentials hidden file exists, if not create it or raise ERROR
-    credentials.check_api(args.weather_model, args.api_uid, args.api_key,
-                          prompt_flag=False, update_flag=False)
+    credentials.check_api(args.weather_model, args.api_uid, args.api_key)
 
     GUNWObj = GUNW(args.file, args.weather_model, args.output_directory)
     GUNWObj()
--- a/tools/RAiDER/models/credentials.py
+++ b/tools/RAiDER/models/credentials.py
@@ -1,6 +1,6 @@
 '''
-dict that hold api information and help url for downloading weather model data
-   -  api information is stored in hidden file in home directory 
+dict that holds api information and help url for downloading weather model data
+   -  api information is stored in a hidden file in home directory 
 
         api filename      weather models          uid           key
         cdsapirc          ERA5, ERA5T             uid           key
--- a/tools/RAiDER/models/credentials.py
+++ b/tools/RAiDER/models/credentials.py
@@ -113,8 +113,8 @@ def check_api(model: str,
         api_filename_path = api_filename_path.expanduser()
 
         # if update flag is on, overwrite existing file 
-        if update_flag:
-            api_filename_path.unlink() if update_flag==True and api_filename_path.exists() else None
+        if update_flag is True:
+            api_filename_path.unlink(missing_ok=True)
         
         # Check if API_RC file already exists
         if not api_filename_path.exists() and UID and KEY:
--- a/tools/RAiDER/models/credentials.py
+++ b/tools/RAiDER/models/credentials.py
@@ -1,14 +1,40 @@
 '''
-dict that holds api information and help url for downloading weather model data
-   -  api information is stored in a hidden file in home directory 
+API credential information and help url for downloading weather model data 
+    saved in a hidden file in home directory 
 
-        api filename      weather models          uid           key
-        cdsapirc          ERA5, ERA5T             uid           key
-        ecmwfapirc        ERAI, HRES              email         key
-        netrc             GMAO, MERRA2            username      password
-        <NAN>             HRRR [public access]    <NAN>         <NAN> 
+api filename      weather models          uid           key           host
+_________________________________________________________________________________
+cdsapirc          ERA5, ERA5T             uid           key         https://cds.climate.copernicus.eu/api/v2
+ecmwfapirc        ERAI, HRES              email         key         https://api.ecmwf.int/v1
+netrc             GMAO, MERRA2            username      password    urs.earthdata.nasa.gov 
+<NAN>             HRRR [public access]    <NAN>         <NAN> 
 '''
 
+import os
+
+# system environmental variables
+API_ENVIRONMENTS= {
+    'cdsapirc'   : {'uid' : 'CDS_UID',
+                    'key' : 'CDS_KEY'},
+
+    'ecmwfapirc' : {'uid' : 'ECMWF_EMAIL',
+                    'key' : 'ECMWF_KEY'},
+                    
+    'netrc'      : {'uid' : 'EARTHDATA_USERNAME',
+                    'key' : 'EARTHDATA_PASSWORD'}
+                    }
+
+# filename for the hidden file per model
+API_FILENAME = {'ERA5'  : 'cdsapirc',
+                'ERA5T' : 'cdsapirc',
+                'ERAI'  : 'ecmwfapirc',
+                'HRES'  : 'ecmwfapirc',
+                'GMAO'  : 'netrc',
+                'MERRA2': 'netrc',
+                'HRRR'  :  None
+               }
+
+# api credentials dict
 MODEL_API_DICT = {
         'cdsapirc' :   {'api' : """\
                                 url: https://cds.climate.copernicus.eu/api/v2\
@@ -33,6 +59,15 @@ MODEL_API_DICT = {
                        }
                 }
 
+# Check if enviroments exists
+def _check_envs(api):
+    try:
+        uid = os.getenv(API_ENVIRONMENTS[api]['uid'])
+        key = os.getenv(API_ENVIRONMENTS[api]['key']) 
+        return uid, key
+    except:
+        return None, None
+
 # Function to check and write MODEL API credential for downloading weather model data
 def check_api(model: str,
               UID: str = None,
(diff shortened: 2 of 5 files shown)


tools/RAiDER/models/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+              UID: str = None,
+              KEY: str = None,
+              prompt_flag: bool = True,
+              update_flag: bool = False) -> None:
+    import os
+
+    # Weather model credential filename
+    # typically stored in home dir as hidden file
+    if model in ('ERA5', 'ERA5T'):
+        ap_filename = 'cdsapirc'
```
cmarshak: typo - if you have flake8 install - should highlight this for you because this variable is never used.

tools/RAiDER/models/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+    elif model in ('ERAI, HRES'):
+        api_filename = 'ecmwfapirc'
+    elif model in ('GMAO'):
+        api_filename = 'netrc'
+    else:
+        api_filename = None #for HRRR
+
+    if api_filename:    
+        # Check if the credential api file exists
+        api_filename_path = os.path.expanduser('~/.')+ f'{api_filename}'
```
cmarshak, line 57: pathlib?

You can then do things like `Path('~/.') / api_filename`

tools/RAiDER/models/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+            if UID is None or KEY is None:
+                url = MODEL_API_DICT[api_filename]['url']
+
+                #if prompt flag selected, ask user to input the API uid & key
+                if prompt_flag:
+                    print(f'GET MODEL API credentials, link: {url}')
+                    UID = input('Please type your UID [uid, email, username]:')
+                    KEY = input('Please type your KEY [key, password]')
+                    if UID == '' or KEY == '':
+                        raise ValueError('ERROR : UID and/or KEY are empty, define them !!')
```
cmarshak, line 76: I think this is confusing - more to maintain. I would remove the prompt option.

I like the error messages - people will likely just make the hidden files themselves. Less code to maintain.

tools/RAiDER/models/credentials.py
```diff
@@ -0,0 +1,90 @@
...
+                    msg = f'{api_filename_path} and weather model credential API UID and KEY,'
+                    msg += ' do not exist !!'
+                    msg += '\nGet API info from ' + '\033[1m' f'{url}' + '\033[0m, and added it!'
+                    raise ValueError(msg)
+                
+            # Create file with inputs, Needed only once
+            print(f'Writing {api_filename_path} locally!')
+            with open(api_filename_path, 'w') as f:
+                f.write(MODEL_API_DICT[api_filename]['api'].format(uid=UID,
+                                                                    key=KEY))
```
cmarshak, line 89: Nice

cmarshak (approved):
Looks good.

A few aesthetic things.

Look at Joe's tests and add some tests to the function to make sure you catch errors correctly.

It would also be helpful for ASF development to provide an example in this PR about how this is used. They will want some interface related to AWS:

```
raider ++process calcGUNW ....
```

Super helpful, Marin!

cmarshak:
For the changelog - look at the latest release and the current changelog version (dev branch).

When I am reading this, the release on conda is 0.4.1. The release you are adding to is 0.4.2 - you would add a brief description there as to what you are adding.

cmarshak:
One more piece that you need to talk to @asjohnston-asf and @jhkennedy about.

They typically write API credentials as environment variables.

Please coordinate with them so that if the weather model is HRES it reads something like `os.environ['HRES_API_KEY']`.

tools/RAiDER/models/credentials.py
dbekaert: Does hardcoding the path with forward slash make this only suitable for linux?

tools/RAiDER/aria/prepFromGUNW.py
```diff
@@ -231,6 +232,10 @@ def update_yaml(dct_cfg:dict, dst:str='GUNW.yaml'):
 def main(args):
     """ Read parameters needed for RAiDER from ARIA Standard Products (GUNW) """
 
+    # Check if WEATHER MODEL API credentials hidden file exists, if not create it or raise ERROR
+    credentials.check_api(args.weather_model, args.api_uid, args.api_key,
```
dbekaert, line 236: @mgovorcin @cmarshak  I wonder if we should add this add the main riader program. The checking of credentials is applicable to all workflows where a model with API is requested?

dbekaert: @jlmaurer I also wonder is there a way for us to check without doing a downlaod of data if credentials are correct? aka do the ecmwfapi and cdsa api allow for a credential check?

cmarshak: I think raider already checks if the credentials exist and has great documentation on how to setup the hidden credential files. Pretty sure errors are very helpful.

I honestly think the existing approach sufficient/preferred. It aligns well with the existing API libraries too. I think passing credentials each time you want to call raider is probably not preferred choice for a user because when they share commands/code, they are *not* going want to share the credentials (just instructions for their team to update their own hidden credential files).

The main thrust of this is being able to pass environment variables that will be defined when a container is created. I don't know the details of this approach except this is a secure, reliable way to pass credentials for this type of processing.

mgovorcin: I think this can be included in the main raider workflow. This is why I added the prompt option to remind a user about API credentials in order to use raider.py. The majority of people do not really read all instructions.

This function serves as a double check to see if the hidden file exists, if not it creates it, and it does it **only once** (the first time). Later on, a user does not need to pass the credentials or think about this anymore. In the case of calcDelayGUNW, prompt_flag will be hardcoded to False, as we do not want to get stuck by waiting for the input.

For the downloading error check, maybe we can add one test dataset to download per model, and try to download it. I do not know if this is a good option as this will run every time when we wanna run raider. I think it good approach to report error if files cannot be downloaded and point user to check his credentials.

tools/RAiDER/models/credentials.py
mgovorcin: already fixed

tools/RAiDER/aria/prepFromGUNW.py
dbekaert: Got it, perhaps we could add a seperate script stand-alone from the raider workflow where we do this.
e.g. script --model 
- the script checks if file exist, if not checks env variables, if nto pings users
- then script makes a dummy call to the API to check if credentials work and report out.
Half of the functionally exist with what you have already.
The dummy check i would like to see if there is a potentially handschake only option rather than fetching data.
 
Inclusion in the main raider code to see if the file exist would be good for that model being requested and error out if needed.

For example our HRES credentials cycle and you will want to catch that early.

tools/RAiDER/models/credentials.py
mgovorcin: I found here the same way of cross-platform dealing with homedir:
https://stackoverflow.com/questions/4028904/what-is-a-cross-platform-way-to-get-the-home-directory 
But open to suggestions.

tools/RAiDER/models/credentials.py
mgovorcin: @cmarshak I added prompt option if we wanna use credentials.py for the main raider workflow, so that we can remind user about API credentials, Also added the URL link where and how they can get it. 

By adding this, this function gets more general use and it is not just tailored for GUNWs.

If you still prefer to remove this for easier maintenance, could do that?

tools/RAiDER/models/credentials.py
mgovorcin: @cmarshak  I also prefer pathlib, but did not find it in environmental.yaml . So avoided introducing a dependency on 
 additional module and used the default one "os"

mgovorcin:
> 

@cmarshak  will add the tests as suggested. As for the passing os.environ['HRES_API_KEY'], maybe this can be done when passing API_UID or API_KEY as 

raider.py ++process calcDelaysGUNW - -m HRES  -uid $HRES_API_UID -key $HRES_API_KEY

just not sure how to pass env ${MODEL}_API_UID

cmarshak:
>just not sure how to pass env ${MODEL}_API_UID

Something like this: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/__main__.py#L64-L68

tools/RAiDER/models/credentials.py
mgovorcin: prompt option removed

mgovorcin:
> > just not sure how to pass env ${MODEL}_API_UID
> 
> Something like this: https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/__main__.py#L64-L68

added an option to search for the API env variables

cmarshak:
@mgovorcin - I am loosing you with how you are setting up your reading of env variables.

Please keep it simple here.

For HRES:

+ `RAIDER_HRES_URL`
+ `RAIDER_HRES_API_KEY`
+ `RAIDER_HRES_EMAIL`

Use these variables exactly as `os.environ['RAIDER_HRES_API_KEY']`.

For ERA5:

+ `RAIDER_ECMWF_ERA5_API_KEY`
+ `RAIDER_ECMWF_ERA5_UID`

These are how our secrets will be accessed.

tools/RAiDER/models/credentials.py
jhkennedy: @mgovorcin `pathlib` is in the python standard library, same as `os`:
https://docs.python.org/3/library/pathlib.html

mgovorcin: @jhkennedy thanks, then I will use pathlib instead os

cmarshak:
For the `netrc` (GMAO), use the environment variables exactly as they are in for the plugin. Please double check with @asjohnston-asf.

mgovorcin:
> @mgovorcin - I am loosing you with how you are setting up your reading of env variables - there seems to be nested dictionaries. I trust you are doing things correctly - just make sure they are consistent with below.
> 
> For HRES:
> 
> * `RAIDER_HRES_URL`
> * `RAIDER_HRES_API_KEY`
> * `RAIDER_HRES_EMAIL`
> 
> Use these variables exactly as `os.environ['RAIDER_HRES_API_KEY']`.
> 
> For ERA5:
> 
> * `RAIDER_ECMWF_ERA5_API_KEY`
> * `RAIDER_ECMWF_ERA5_UID`
> 
> These are how our secrets will be accessed.

@cmarshak  I checked [cdsapi](https://github.com/ecmwf/cdsapi) and [ecwmfi](https://github.com/ecmwf/ecmwf-api-client)  and use the ones API's are looking for. 

I guess this is to use take into account different credentials for HRES model. I will update it to the ones above

jlmaurer:
@mgovorcin @dbekaert  I think this looks good to me!

tools/RAiDER/models/credentials.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # skip below if model is HRRR as it does not need API
+    if api_filename:    
+        # Check if the credential api file exists
+        api_filename_path = Path(output_dir) / (hidden_ext + api_filename)
+        api_filename_path = api_filename_path.expanduser()
+
+        # if update flag is on, overwrite existing file 
+        if update_flag:
+            api_filename_path.unlink() if update_flag==True and api_filename_path.exists() else None
```
jhkennedy, line 117: I believe this can be simplified to:
```suggestion
        if update_flag is True:
            api_filename_path.unlink(missing_ok=True)
```
https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

jhkennedy (approved):
This looks great! One tiny suggestion that you can take or leave.

cmarshak:
@mgovorcin @jlmaurer - when this is merged, can Raider be released to include these updates? Would it be possible to have this released this week? I am hoping ASF has the release to include ERA5 and HRES via Hyp3.

jhkennedy:
@cmarshak for you to start using it via HyP3, it just needs to be merged to `dev` as Hyp3 is using the `test` tagged container for the `INSAR_ISCE_TEST` job:
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE_TEST.yml#L85-L86

cmarshak:
Cool - not worried about release then. Thanks for clarifying, Joe.

PR #495, opened 2023-02-22
fix bug if RC_file exists

mgovorcin:
## Description
@bbuzz31  found this bug. Correctly read credentials from hidden RC file if exists


bbuzz31:
Looks good @mgovorcin. Can you just update the changelog? Test weren't passing previously because i think GMAO was down... hopefully successful now.

PR #497, opened 2023-02-28
Store SLC delays in GUNW

bbuzz31:
Change the method of storing delays in the GUNW from interferometric (ref-sec) to single SLC (ref and sec).
See groups in image as discussed last week.

@sssangha I believe this will require you making some adjustments to AT and then I can double check the conversion from meters to radians.

@mgovorcin this should also make your life easier in prepping aria for MintPy.

Note this is built on top of #495 

image


cmarshak (approved):
From what I can see you do not affect raider core (only add logging statements) and only modify the GUNW script. Should be fine to merge. If there are bugs, will have to fix them!

PR #499, opened 2023-03-01
Temporal interpolation of delays

jlmaurer:
Provides temporal interpolation as the default option when requesting a datetime not exactly covered by the weather model 

## Description

- Based on and replaces #479

- provides a better estimate of the delay when the requested datetime is not close to the time of the weather model


bbuzz31:
@jlmaurer note that this was before current dev so will have to be rebased, in particular to support the new handling of credentials.

Also note that I did implement two unit tests

CHANGELOG.md
```diff
@@ -19,7 +19,11 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 + Only orbit file is used (even if multiple specified) to minimize errors and ensure consistency over region
 + GUNW packaging is restructed to store SLC (ref and sec) wet and tropo delays rather than the differential
 
-## [0.4.1]
+##[0.4.2]
```
dbekaert, line 22: This release already exist on dev, but is being captured as a change. Can you double check the rebasing is done correctly?

tools/RAiDER/utilFcns.py
```diff
@@ -1135,3 +1136,59 @@ def transform_coords(proj1, proj2, x, y):
...
+
+    # If you're within 5 minutes just take the closest time
+    if get_dt(tclose, t0) < _THRESHOLD_SECONDS:
+        return [tclose]
+    else:
+        if t2 > tclose:
+            return [tclose, t2]
+        else:
+            return [t2, tclose]
+
```
dbekaert, line 1176: such an assumption should be documented. Once might question by 5 min why not 30 min; why interpolate at all?

tools/RAiDER/models/gmao.py
```diff
@@ -37,7 +37,7 @@ def __init__(self):
         self._k2 = 0.233  # [K/Pa]
         self._k3 = 3.75e3  # [K^2/Pa]
 
-        self._time_res = 1
+        self._time_res = 3
```
dbekaert, line 40: i notice this change here. should be captured in change log.

tools/RAiDER/constants.py
```diff
@@ -18,3 +18,6 @@
 
 R_EARTH_MAX = 6378137
 R_EARTH_MIN = 6356752
+
+_THRESHOLD_SECONDS = 5 * 60 # Threshold delta_time in seconds
```
dbekaert, line 22: see my comment below

dbekaert:
one item that i had brought up before with @bbuzz31:

What is the default behaviour that is allowed by RaiDER?
Do we allow 1 orbit file, 1 region, with multiple dates and  single UTC?

bbuzz31:
> one item that i had brought up before with @bbuzz31:
> 
> What is the default behaviour that is allowed by RaiDER? Do we allow 1 orbit file, 1 region, with multiple dates and single UTC?

I suggest we start with restricting it to do as you suggest: 1 region; 1 orbit; 1 UTC; multiple dates. 
Possibly add a warning for multiple dates alerting that a single orbit file will be used.

tools/RAiDER/constants.py
jlmaurer: @dbekaert Note that this particular threshold is for whether or not to interpolate. If the threshold is close enough, the code will just take the closest date. Threshold could be anything, I just picked 5 minutes and set it as a constant so it can be changed if desired.

tools/RAiDER/utilFcns.py
jlmaurer: @dbekaert yes exactly. I should update the CHANGELOG. The idea is is that if you are within some small increment of a date, there is no reason to interpolate. The increment is set by `_THRESHOLD_SECONDS`. E.g. if you're within 10 seconds, you save time and numerical issues by simply taking the closest time.

tools/RAiDER/models/ecmwf.py
```diff
@@ -247,15 +250,17 @@ def _get_from_cds(
...
+        """ Used for HRES """
         from ecmwfapi import ECMWFService
 
         server = ECMWFService("mars")
 
         # round to the closest legal time
         corrected_date = util.round_date(time, datetime.timedelta(hours=self._time_res))
         if not corrected_date == time:
-            logger.warning('Rounded given datetime from  %s to %s', time, corrected_date)
+            logger.warning('Rounded given hour from  %d to %d', time.hour, corrected_date.hour)
```
bbuzz31, line 263: This file is rebased incorrectly; should be the older version

bbuzz31 (changes requested):
ecmwf.py  file is rebased incorrectly; should be the older version

bbuzz31:
I think this is good to go after you pull in my fixes on your local branch @jlmaurer

bbuzz31 (approved):

Issue #503, opened 2023-03-15
[BUG] - RAIDER + ERA5 Fails over Hawaii


cmarshak:
**Describe the bug**
Here is a list of failed jobs.

```
[{'job_id': 'baf54612-cc8c-4182-9fb0-f7e6f06ad490',
  'job_type': 'INSAR_ISCE_TEST',
  'request_time': '2023-03-14T19:23:25+00:00',
  'status_code': 'FAILED',
  'user_id': 'cmarshak',
  'name': 'Hawaii_124_ERA5_0314',
  'job_parameters': {'compute_solid_earth_tide': True,
   'estimate_ionosphere_delay': True,
   'frame_id': 19224,
   'granules': ['S1A_IW_SLC__1SDV_20230217T043052_20230217T043122_047271_05AC44_293F',
    'S1A_IW_SLC__1SDV_20230217T043119_20230217T043147_047271_05AC44_56B5'],
   'secondary_granules': ['S1A_IW_SLC__1SDV_20230205T043052_20230205T043122_047096_05A65E_2987',
    'S1A_IW_SLC__1SDV_20230205T043120_20230205T043148_047096_05A65E_5D6F'],
   'weather_model': 'ERA5'},
  'logs': ['https://hyp3-a19-jpl-contentbucket-1wfnatpznlg8b.s3.us-west-2.amazonaws.com/baf54612-cc8c-4182-9fb0-f7e6f06ad490/baf54612-cc8c-4182-9fb0-f7e6f06ad490.log'],
  'expiration_time': '2023-09-11T00:00:00+00:00',
  'processing_times': [7254.499, 7.754]},
 {'job_id': '3890a1c2-27c3-4d2e-83c6-854ab139bdbe',
  'job_type': 'INSAR_ISCE_TEST',
  'request_time': '2023-03-14T19:23:25+00:00',
  'status_code': 'FAILED',
  'user_id': 'cmarshak',
  'name': 'Hawaii_124_ERA5_0314',
  'job_parameters': {'compute_solid_earth_tide': True,
   'estimate_ionosphere_delay': True,
   'frame_id': 19223,
   'granules': ['S1A_IW_SLC__1SDV_20230301T043052_20230301T043122_047446_05B233_640C'],
   'secondary_granules': ['S1A_IW_SLC__1SDV_20230217T043052_20230217T043122_047271_05AC44_293F'],
   'weather_model': 'ERA5'},
[...]

PR #504, opened 2023-03-15
catch latency issue for era5

bbuzz31:
See #503 
ERA5T replaces ERA5 for data within three months of today.

["Initial release data, i.e. data no more than three months behind real time, is called ERA5T."](https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation)

@dbekaert @jlmaurer 

I've updated the 'valid_date' parameter to reflect this, such that the dates will be skipped and log reflect it when the requested date is within three months.

1) How do we want to handle GUNW/HYP3? 
- We could automatically adjust the weather model to ERA5T but these weather data are not final.

2) How do we want to handle temporal interpolation?
- For example, latest available date is currently Dec 15, 2022. If use requests 23:01:00, then Dec 16, 2022 weather model file will try to get used and fail. We could revert to regular interpolation or automatically use 5T in this case as well

### DIFF
--- a/tools/RAiDER/models/era5.py
+++ b/tools/RAiDER/models/era5.py
@@ -1,9 +1,11 @@
 import datetime
+from dateutil.relativedelta import relativedelta
 
 from pyproj import CRS
 
 from RAiDER.models.ecmwf import ECMWF
 from RAiDER.logger import logger
+
 
 
 class ERA5(ECMWF):
@@ -20,9 +22,11 @@ class ERA5(ECMWF):
         self._proj = CRS.from_epsg(4326)
 
         # Tuple of min/max years where data is available.
-        self._valid_range = (datetime.datetime(1950, 1, 1), "Present")
+        end_date = datetime.datetime.today() - relativedelta(months=3)
+        self._valid_range = (datetime.datetime(1950, 1, 1), end_date)
+
         # Availability lag time in days
-        self._lag_time = datetime.timedelta(days=30)
+        self._lag_time = relativedelta(months=3)
 
         # Default, need to change to ml
         self.setLevelType('pl')
--- a/tools/RAiDER/models/weatherModel.py
+++ b/tools/RAiDER/models/weatherModel.py
@@ -262,19 +262,25 @@ class WeatherModel(ABC):
         '''
         Checks the time against the lag time and valid date range for the given model type
         '''
+        end_time = self._valid_range[1]
+        end_time = end_time if isinstance(end_time, str) else end_time.date()
+
         logger.info(
-            'Weather model %s is available from %s-%s',
-            self.Model(), self._valid_range[0], self._valid_range[1]
+            'Weather model %s is available from %s to %s',
+            self.Model(), self._valid_range[0].date(), end_time
         )
+
         if time < self._valid_range[0]:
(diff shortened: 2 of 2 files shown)


tools/RAiDER/models/era5t.py
```diff
@@ -14,4 +14,5 @@ def __init__(self):
         self._Name = 'ERA-5T'
 
         self._valid_range = (datetime.datetime(1950, 1, 1), "Present")  # Tuple of min/max years where data is available.
-        self._lag_time = datetime.timedelta(days=5)  # Availability lag time in days
+        # Availability lag time in days; actually about 12 but unstable on ECMWF side
```
dbekaert, line 17: 12 hours, min, sec ;)

tools/RAiDER/models/era5t.py
```diff
@@ -14,4 +14,5 @@ def __init__(self):
...
+        self._lag_time = datetime.timedelta(days=1)
```
dbekaert, line 18: So you are making it 24 hours? Can you check what the documention states for ERA5T? and reference it in the text?

dbekaert:
> I've updated the 'valid_date' parameter to reflect this, such that the dates will be skipped and log reflect it when the requested date is within three months of today (= when raider is run).

I looked at the code and you are raising an error; i.e. not skipping. Error is the right behavior i believe

dbekaert:
> We could automatically adjust the weather model to ERA5T but these weather data are not final.

That would currently be a different model. If we need to resolve the latency issue my preference would be to use HRES. So perhaps we just skip for now how we want to handle it.

dbekaert:
> How do we want to handle temporal interpolation?

I though your error out in case you are missing one or two weather model corrections. 
Hence that would mean you would not even get to the temporal interpolation part.

I think for the production we should handle this in part from the enumeration stage.

bbuzz31:
> > I've updated the 'valid_date' parameter to reflect this, such that the dates will be skipped and log reflect it when the requested date is within three months of today (= when raider is run).
> 
> I looked at the code and you are raising an error; i.e. not skipping. Error is the right behavior i believe. If we would let is pass through then there is no way in production to make sure its included.

Yes and no. The error is raised but then caught in the main program (`raider.py`). So if there are multiple dates requested, some can work and others not

dbekaert:
Can you catch it and error out in your workflow?

bbuzz31:
@dbekaert we should be good to go now.

RAiDER itself can still skip over a date (and now it will also skip the date if temporal interpolation is requested and fails)
However, the GUNW workflow will only work if two delays cubes (one per SAR date) are successfully written.

bbuzz31:
I also fixed #505 by simply replacing `np.sum(wgts) == 1` with `np.isclose(np.sum(wgts), 1)` , as the sum of wgts was 0.9999999999999999 which is false in former and true in latter.

cmarshak:
@bbuzz31 - when there is a minor fix like the interpolation from #505 - I think it is preferable to make that fix and merge it in separate PR, particularly if the priority is high. The error catching is important but not a priority and should be done according to David's suggestion. Please merge the interpolation fix as soon as you have a chance into dev either in this PR or elsewhere. We need to keep catching errors by running via hyp3.

dbekaert (approved):
Lgtm @bbuzz31 thnx

PR #511, opened 2023-04-06
Update scipy requirements

sssangha:
## Description
Update scipy requirement to use recent revisions with bug-fix

## Motivation and Context
This scipy issue has been captured and accounted for in other repos (e.g. [mintpy](https://github.com/insarlab/MintPy/pull/968))

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 + Verify some constants / equations and remove the comments questioning them
 + Relocate the time resolution of wmodels to one spot
 + Skip test_scenario_3 until a new golden dataset is created
++ Update scipy requirement
 
 ## [0.4.2]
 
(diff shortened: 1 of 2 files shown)


dbekaert:
@sssangha mintpy actually does not have any requirements imposed on scipy.

jlmaurer (approved):
I think this is fine.

dbekaert:
Im still confused, is it an issue or has it been resolved?
I dont see that mintpy is skipping that specific version, its fully free to
resolve the version.

sssangha:
@dbekaert no, we must explicitly specify this condition, otherwise we run the risk of having this buggy `scipy 1.10.0` version persist in an environment.

E.g. let's say somebody has an existing build with `scipy 1.10.0`, then if they try to update with a YML with no condition imposed (i.e. just listing `scipy`), then nothing will happen as conda will move on after recognizing scipy already exists. In order to avoid this version, we must impose a restriction as I've proposed

PR #513, opened 2023-04-08
Allow logging path to be specified by user

### LINKED ISSUE #510: [BUG] Log file creation
**Describe the bug**
By default, RAiDER outputs a log file to the current working directory (see https://github.com/dbekaert/RAiDER/blob/d84593b23be6f4fd04e22c9a7ab0a33009c291a3/tools/RAiDER/logger.py#L17).

This causes issues when running the code in a working directory that does not have writing permissions. 

**To Reproduce**
Run the software in a directory that has no writing permission

**Expected behavior**
It is expected that the workflow allows the user to:
1. Specify the directory where the log file is created
2. Turn off the generation of a log file

**Additional context**
For further details, see the "Known issues" section here: https://github.com/nasa/opera-sds-pge/wiki/CSLC-S1-Gamma-Acceptance-Testing-Instructions#known-issues

jlmaurer:
## Description

- Fixes #510 

## How Has This Been Tested?

- `test_scenario_1.py` does not pass locally, but should not be related to this PR

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/cli/__init__.py
+++ b/tools/RAiDER/cli/__init__.py
@@ -40,5 +40,9 @@ DEFAULT_DICT = AttributeDict(
             weather_model_directory=None,
             output_projection='EPSG:4326',
             interpolate_time=True,
+            logger_path=os.getcwd(),
         )
     )
+
+# Can change the default log location
+LOGGER_PATH = DEFAULT_DICT.logger_path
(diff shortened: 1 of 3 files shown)


dbekaert:
@jlmaurer can the PR status be changed? LGTM is something still missing?

jlmaurer:
@dbekaert I realized that this is not quite sufficient after I started testing, which is why I changed to draft. The problem occurs because we import the logging module throughout the code, but in order to set the logger location based on the user configuration file it has to pass thru one of the main functions we have set up. This just takes a bit more thought about how to call the logging module, since it won't necessarily always be called as part of a main function call.

jlmaurer:
Closing as this is replaced by #514

PR #515, opened 2023-04-13
Implement tests of equations with synthetic weather model

bbuzz31:
- Unit tests for ray tracing equations  
- fix a tiny bug when a single point is requested by user


test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+            cos_factor = (high_ht - low_ht) / ray_length if cos_factor is None else cos_factor
+
+    return outputArrs
+
+
+@dataclass
+class StudyArea(object):
+    """ Object with shared parameters related to the study area """
+    reg:str
+    WM:str
```
cmarshak, line 156: Don't you use WM for classes sometimes? Please make the variable also more descriptive. Not upper case.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+    def setup_region(self):
+        if self.reg == 'LA':
+            self.SNWE  = 33, 34, -118.25, -117.25
+            self.dt    = datetime(2020, 1, 30, 13, 52, 45)
+            self.orbit = self.wd + \
+                '/S1B_OPER_AUX_POEORB_OPOD_20210317T025713_V20200129T225942_20200131T005942.EOF'
+
+        elif self.reg == 'Fort':
```
cmarshak, line 189: So for "region_short_name" or whatever you call it - you should have some documentation of what those regions are. Is it Los Angeles or Louisiana. Both are LA. Fort Lauderdale is under water right now, so that is the only Fort on my mind.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+            op.dirname(SAobj.path_wm_real)
+    dct_cfg['download_only'] = True
+
+    cfg = update_yaml(dct_cfg)
+
+    ## run raider to download the real weather model
+    cmd  = f'raider.py {cfg}'
+
+    proc = subprocess.run(cmd.split(), stdout=subprocess.PIPE, universal_newlines=True)
+    assert np.isclose(proc.returncode, 0)
```
cmarshak, line 228: This should be ==

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    # normalize with the theoretical data and compare difference with 0
+    resid     = (ray_data - raid_data) / ray_data
+    np.testing.assert_almost_equal(0, resid, decimal=7)
+
+    da.close()
+    del da
+    return
+
+
+def test_wet_eq1(reg='Fort', mod='GMAO'):
```
cmarshak, line 291: I believe you don't need to repeate your inputs if you use parametrization in pytest: https://www.tutorialspoint.com/pytest/pytest_parameterizing_tests.htm

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    # normalize with the theoretical data and compare difference with 0
+    resid     = (ray_data - raid_data) / ray_data
+    np.testing.assert_almost_equal(0, resid, decimal=7)
+
+    da.close()
+    del da
+    return
+
+
+def test_wet_eq2(reg='Fort', mod='GMAO'):
```
cmarshak, line 350: Combine these two tests.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+            'date_group': {'date_list': datetime.strftime(self.dt, '%Y%m%d')},
+            'cube_spacing_in_m': str(self.cube_spacing),
+            'los_group': {'ray_trace': True, 'orbit_file': self.orbit},
+            'weather_model': self.WM,
+            'runtime_group': {'output_directory': self.wd},
+        }
+        return dct
+
+
+def dl_real(reg='LA', mod='GMAO'):
```
cmarshak, line 210: All the things above are technically "fixtures". Would be nice to use the pytest decorator as such, but not required.

https://docs.pytest.org/en/6.2.x/fixture.html

test/test_synthetic.py
cmarshak: https://github.com/bbuzz31/RAiDER/blob/74535ed4baac1d960ec7f6f3431f9a202d16e6e0/test/__init__.py#L23

tools/RAiDER/models/weatherModel.py
```diff
@@ -847,34 +846,18 @@ def write(
...
+
+    n = np.ceil(ll_bounds[1])
+    N = f'{np.abs(n):.0f}S' if n <0 else f'{n:.0f}N'
+
+    w = np.floor(ll_bounds[2])
+    W = f'{np.abs(w):.0f}W' if w <0 else f'{w:.0f}E'
+
+    e = np.ceil(ll_bounds[3])
+    E = f'{np.abs(e):.0f}W' if e <0 else f'{e:.0f}E'
+    return f'{name}_{time.strftime("%Y_%m_%d_T%H_%M_%S")}_{S}_{N}_{W}_{E}.nc'
```
cmarshak, line 860: What are you doing here? As in I thought your work only was generating a test?

tools/RAiDER/losreader.py
```diff
@@ -228,6 +228,7 @@ def getLookVectors(self, ht, llh, xyz, yy):
         '''
         # TODO - Modify when isce3 vectorization is available
         los = np.full(yy.shape + (3,), np.nan)
+        llh = llh.copy()
```
cmarshak, line 231: Why?

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    cmd  = f'raider.py {cfg}'
+    proc = subprocess.run(cmd.split(), stdout=subprocess.PIPE, universal_newlines=True)
+    assert np.isclose(proc.returncode, 0)
+
+    # get the just created synthetic delays
+    ds = xr.open_dataset(
+        f'{SAobj.wd}/{SAobj.WM}_tropo_{SAobj.dts.replace("_", "")}_ray.nc')
+    da = ds['wet']
+    ds.close()
+    del ds
```
cmarshak, line 383: Why delete?

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    # normalize with the theoretical data and compare difference with 0
+    resid     = (ray_data - raid_data) / ray_data
+    np.testing.assert_almost_equal(0, resid, decimal=7)
+
+    da.close()
+    os.remove('./temp.yaml')
+    os.remove('./error.log')
+    os.remove('./debug.log')
+    del da
+    return
```
cmarshak, line 407: Remove return.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    raid_data = da.data * 1e6
+
+    assert np.all(np.abs(ray_data) > 1)
+    assert np.all(np.abs(raid_data) > 1)
+
+    # normalize with the theoretical data and compare difference with 0
+    resid     = (ray_data - raid_data) / ray_data
+    np.testing.assert_almost_equal(0, resid, decimal=7)
+
+    da.close()
```
cmarshak, line 402: You can close a dataset and still have an array from that dataset open? Didn't know that.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+def test_hydrostatic_eq(reg='Fort', mod='GMAO'):
+    """ Test the hydrostatic eqn by setting t=p and comparing to ray
+
+    The ray length is computed here (length_of_ray; m), and in raider (m * K/Pa)
+    ray length computed here is then scaled by constant K/Pa
+    ray length from raider is unscaled from parts per million
+    We check they are both large enough for meaningful numerical comparison (>1)
+    We then compute residual and normalize it we theoretical ray length
+    We ensure that normalized residual is not significantly different from 0
+        significantly different = 7 decimal places
```
cmarshak, line 241: Please mention something to the effect: The integration of the delay along a ray from the weather model node to the sensor is the length of the ray when the integrand is constant.

tools/RAiDER/delay.py
```diff
@@ -399,14 +402,17 @@ def _build_cube_ray(
                 else:
                     pts = np.stack(pts, axis=-1)
 
+                ## adjust the lowest level by ~ 1 mm to account for transformation error
+                if (pts[:, :, -1] < ht).all():
+                    pts[:, :, -1] = ht
+
```
cmarshak, line 408: What happens if this doesn't happen? It's not clear why you need this.

cmarshak: So there was a bug in raider previously?

cmarshak (changes requested):

test/test_synthetic.py
cmarshak: I mean the description "Test the hydrostatic eqn by setting t=p and comparing to ray" should be expanded.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.4.3]
++ add unit tests for the hydro and two pieces of wet equation
++ bump  bottom height of user requested levels by ~1mm during ray tracing to ensure interpolation works
```
dbekaert, line 11: What does this mean? Does this imply a bug in Raider? You should not need to modify code outside raider to get the tests to work.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
...
++ ensure directories for storage are written
++ don't allow use pass keys that dont exist (likely a typo on user end)
++ fix bug in writing delays for station files
```
dbekaert, line 14: are you doing more than what is described in the PR?

tools/RAiDER/delay.py
dbekaert: was it not working before?

tools/RAiDER/models/weatherModel.py
dbekaert: @jlmaurer Does this break things/ backward compatible?

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    params = {**params, **dct_cfg}
+
+    with open(dst, 'w') as fh:
+        yaml.safe_dump(params, fh,  default_flow_style=False)
+
+    return dst
+
+
+def update_model(wm_file:str, kind:str, wm_dir:str='weather_files_synth'):
+    """ Update the weater model file; keep kind to equation to test """
```
dbekaert, line 48: This description could be better. Like say what kind is i.e (hydro, wet1, wet2)

cmarshak: That is `kind` is `category_of_synth_weather_model` with a good description of what you are doing. Otherwise this is very unclear.

cmarshak: The doc-string should have all the categories that you are considering and why. The model equation and a reference should be somewhere too.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    p  = ds['p']
+    e  = ds['e']
+    if kind == 'hydro':
+        p = t
+    elif kind == 'wet1':
+        e = t
+        Obj._k3 = 0
+    elif kind == 'wet2':
+        e = t**2
+        Obj._k2 = 0
```
dbekaert, line 66: we should put these assumptions into the description of the function. 

i.e. when Hydo is kind we test the hydrostatic component of the equation P/T where P=T.

cmarshak: Again, this would be helped if there was some references to model integrals and assumptions.

test/test_synthetic.py
```diff
@@ -0,0 +1,407 @@
...
+    def __post_init__(self):
+        self.setup_region()
+
+        self.dts   = self.dt.strftime('%Y_%m_%d_T%H_%M_%S')
+        self.ttime = self.dt.strftime('%H:%M:%S')
+
+        self.wmObj = modelName2Module(self.WM.upper().replace("-", ""))[1]()
+
+        aoi = BoundingBox(self.SNWE)
+        aoi.add_buffer(buffer = 1.5 * self.wmObj.getLLRes())
```
dbekaert, line 168: Why not use internal functions. If this now changes internally then we dont catch that through the test?

test/test_synthetic.py
dbekaert: phrase the hypothesis of what you are testing and then explain how you test it.

The hydrostatic delay reduces to an integral along the ray path when P=T. However the constants k1 and scaling of 10-6 woill remain present leading to a result of ray-lenght *k1*10^-6. 

Below we do the following steps:

tools/RAiDER/checkArgs.py
```diff
@@ -58,7 +59,7 @@ def checkArgs(args):
                         d.strftime('%Y%m%dT%H%M%S'),
                     )
                 )
-                hydroFilename = None # only the 'wetFilename' is used for the station_file
+                hydroFilename = '' # only the 'wetFilename' is used for the station_file
```
dbekaert, line 62: Can you confirm/ justify the change?

test/test_synthetic.py
bbuzz31: I've noticed weird behavior (some of which you noticed below: a dataset can be closed but its data array stay open) when trying to close / flush datasets to disk. This was just to be extra safe

CHANGELOG.md
bbuzz31: I wouldn't call this a bug in raider so much as a manifestation of projection error. 
If user requests the bottom layer of weather model, which is -500 m, then due to the conversion of interpolation points from [WGS84 -> ECEF](https://github.com/dbekaert/RAiDER/blob/74535ed4baac1d960ec7f6f3431f9a202d16e6e0/tools/RAiDER/delay.py#L324)(for getting the ray vectors) and then from [ECEF to WGS84](https://github.com/dbekaert/RAiDER/blob/74535ed4baac1d960ec7f6f3431f9a202d16e6e0/tools/RAiDER/delay.py#L393) (for intersecting the ray vectors with delays) the bottom height layer becomes about -500.0002. (about 0.2 mm difference).

As this new value is outside of the actual weather model height we interpolate to (-500 m) it results in nan values.  For a real weather model, this results in the loss of 30 (260) nondimensional contributions to the integral for wet (hydrostatic) delays. Considering the total integral, this changes the ray length by about 100 m for hydrostatic equation (normalized difference of 0.001). 

I'm of the opinion we should include the bottom layer, which [my change](https://github.com/dbekaert/RAiDER/blob/74535ed4baac1d960ec7f6f3431f9a202d16e6e0/tools/RAiDER/delay.py#L406) does.

tools/RAiDER/delay.py
bbuzz31: Please see the comment on the change log which explains this fix.

CHANGELOG.md
cmarshak: 1. How would we know this without this comment above?
2. This discussion and improvement really merits its own PR
3. Can you develop a test to illustrate the improvement for illustration?

cmarshak: And...

You are assuming that pts z coordinate *never* exceeds 1 mm. What happens if the z coord exceeds 1 meter or even 1 km? Sounds like an unexpected side-effect. There is no reference to 1 mm in this update.

tools/RAiDER/checkArgs.py
bbuzz31: If a user asks for delays a point (s), there is no [hydroFilename (f)](https://github.com/dbekaert/RAiDER/blob/74535ed4baac1d960ec7f6f3431f9a202d16e6e0/tools/RAiDER/cli/raider.py#L222-L225). 

Then, if hydroFilename/f is None, the string replacement will fail for delays on slant  (L324, L326). hydroFilename/f isn't used in the subsequent call to writeDelays for this AOI type.

cmarshak: Is there an issue ticket? This should be a separate PR.

CHANGELOG.md
bbuzz31: hmm looks like i slightly misunderstood / misexplained. the z coord in `pts` variable which is being adjusted is the weather model z values. so that adjustment will always force it to the lowest z (-500; code needs to be updated to reflect this). There's a test in the last cells of [this notebook](https://github.com/dbekaert/RAiDER-docs/blob/66cd4e13dd55fd5ca3087b256abb1e4e7da780b8/notebooks/Synthetic_Test/SyntheticTest.ipynb).

test/test_synthetic.py
bbuzz31: Could you clarify - you want to hardcode the buffer we expect, so if the buffering changes within RAiDER, we'd catch that here?

tools/RAiDER/delay.py
```diff
@@ -402,7 +402,12 @@ def _build_cube_ray(
                 else:
                     pts = np.stack(pts, axis=-1)
 
-                ## adjust the lowest level by ~ 1 mm to account for transformation error
+                # ray points first exist in ECEF; they are then projected to WGS84
+                # this adds slight error (order 1 mm for 500 m)
+                # at the lowest weather model layer (-500 m) the error pushes the
+                # ray points slightly below (e.g., -500.0002 m)
+                # the subsequent interpolation then results in nans
+                # here we force the lowest layer up to -500 m if it exceeds it
```
bbuzz31, line 410: @piyushrpt @jlmaurer could you verify this is a satisfactory explanation and change?

bbuzz31:
@piyushrpt could you let us know why we stop integrating at 50 km rather than going all the way up to 80? Does this add bias wrt zenith projected to slant? 

https://github.com/dbekaert/RAiDER/blob/7957594dce848749c3252788617a2d4f349e240f/tools/RAiDER/delay.py#L353

test/test_synthetic.py
jlmaurer: Datasets should be deleted to flush

tools/RAiDER/delay.py
jlmaurer: This works for me.

tools/RAiDER/checkArgs.py
jlmaurer: @bbuzz31 this is not a big deal to me but I do tend to agree on a separate PR for this one. We could do some more thorough testing on filenames for various input options.

jlmaurer:
> @piyushrpt could you let us know why we stop integrating at 50 km rather than going all the way up to 80? Does this add bias wrt zenith projected to slant?
> 
> https://github.com/dbekaert/RAiDER/blob/7957594dce848749c3252788617a2d4f349e240f/tools/RAiDER/delay.py#L353

@piyushrpt can give his thoughts but based on looking at the code it appears to me that `MAX_TROPO_HEIGHT` as defined in the code was just a placeholder that could be modified by a user through the config file if desired. If this is the case, then we just need to ensure that `MAX_TROPO_HEIGHT` gets set to an appropriate default value and can be modified by the config file if desired. I suggest to deal with this in a separate PR.

CHANGELOG.md
jlmaurer: @bbuzz31 is correct in that this is I would say a very minor fix and doesn't have anything to do with the raytracing per se, has to do with the projections and numerical precision. I also think that this is needed to ensure the raytracing is tested correctly. The thing that needs to be done here is to ensure that the lowest z-level always gets set back to the input z-level rather than slightly beneath, as that results in nans.

tools/RAiDER/delay.py
```diff
@@ -399,14 +402,22 @@ def _build_cube_ray(
                 else:
                     pts = np.stack(pts, axis=-1)
 
+                # ray points first exist in ECEF; they are then projected to WGS84
+                # this adds slight error (order 1 mm for 500 m)
+                # at the lowest weather model layer (-500 m) the error pushes the
+                # ray points slightly below (e.g., -500.0002 m)
+                # the subsequent interpolation then results in nans
+                # here we force the lowest layer up to -500 m if it exceeds it
+                if (pts[:, :, -1] < model_zs[0]).all():
```
jlmaurer, line 411: @bbuzz31 I might change the implementation here slightly, using `model_zs.min()` instead of indexing, just in case there were ever to be a case where we would want to change the ordering.

jlmaurer: You may have to convert model_zs to a numpy array or use `np.min()` instead if it is a list.

tools/RAiDER/models/weatherModel.py
jlmaurer: @bbuzz31 I would tend to agree here in terms of not changing this since this PR needs to get through as quickly as possible, but if you have sufficient tests then its probably fine.

bbuzz31:
A few notes:

- I switched to ERA5 because of GMAO SSL issues. This necessitated changing the default type of ERA5 from pressure to model layers (as was supposed to be done anyway)
- I'm testing over many more height levels. In the upper levels there is a little more noise added so I decreased the required consistency between analytical/raider delays from 7 decimal place agreement to 6

test/test_synthetic.py
dbekaert: why re-code (hard-code) it here when its an internal function. If you change it internally is not consistent with the test

bbuzz31: Oh I see. I reproduced it exactly as in RAiDER. I think the hardcoding of buffering (1.5x) should be internal to the function. Any objections @jlmaurer?

This should be fixed in another PR probably? @cmarshak

cmarshak: Please do what David is saying here. Can you not import the routine?

cmarshak: If you are updating the function non-trivially, yes - make another PR (branched from dev). Then once that is merged, merged that here.

piyushrpt:
`MAX_TROPO_HEIGHT` was meant to be a placeholder as correctly interpreted. This is meant to be configurable / set appropriately as needed based on model inputs.

dbekaert:
@bbuzz31 if this PR is ready for review by @piyushrpt change it to ready. I would like to have the analytical unit test inplace to make an updated minor release this week.

dbekaert:
@piyushrpt ready for your review.

tools/RAiDER/delay.py
piyushrpt: Also works for me

piyushrpt (approved):
Tests look good to me. 6 digit match is really good.  Would have been happy with mm scale.

PR #522, opened 2023-04-20
Rewrite Ingest JSON for ASF DAAC and update GUNW entrypoint for testing

cmarshak:
## Description

Need to update the GUNW workflow to append a `weather_model` key to the Ingest json of the metadata when a bucket and its prefix are passed as arguments. In the metadata, we are now including a field `"weather_model": [<list_of_weather_models>]` . Here is the schema that will be used: https://github.com/asfadmin/grfn-ingest/blob/test/verify/src/metadata_schema.json (this is now included in the tests as well to validate the final json).

Although this is relatively straightforward modification (see these [lines](https://github.com/cmarshak/RAiDER/blob/update_gunw_metadata_json/tools/RAiDER/cli/raider.py#L505-L530) for the primary thrust), it was important to have some kind of testing. This required updates to the GUNW entrypoint (i.e. to ensure that we could test it without rerunning the computation) as well as cleaning up the previous test suite relevant to GUNW.

For improving coverage of GUNW testing, we parametrized the previous test so we could test HRRR and GMAO, both of which are used in production and have relatively low latency.

## How Has This Been Tested?
Monkeypatching the upload to and download from s3 as well as the computation.

1. The key `weather_model` has been successfully added to a json
2. The schema can be validated using the linked schema.


CHANGELOG.md
```diff
@@ -20,7 +20,13 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 + Verify some constants / equations and remove the comments questioning them
 + Relocate the time resolution of wmodels to one spot
 + Skip test_scenario_3 until a new golden dataset is created
-+ Update environment with scipy version minimum and requiring pybind11
++ Update environment with scipy version minimum and requiring pybind11 (the latter for Apple ARM64 with rosetta2)
```
cmarshak, line 23: In a previous PR - Jeremy wanted to specify why pybind11 was required. This minor comment in the changelog was added.

dbekaert (approved):
LGTM @cmarshak

tools/RAiDER/cli/raider.py
```diff
@@ -442,9 +445,7 @@ def downloadGNSS():
 
 
 ## ------------------------------------------------------------ prepFromGUNW.py
-def calcDelaysGUNW():
-    from RAiDER.aria.prepFromGUNW import main as GUNW_prep
-    from RAiDER.aria.calcGUNW import tropo_gunw_slc as GUNW_calc
+def calcDelaysGUNW(iargs: list[str] = None):
```
cmarshak, line 448: Following this: https://jugmac00.github.io/blog/testing-argparse-applications-the-better-way/

See earlier functions in this file - I see that Jeremy was already doing this.

cmarshak:
@jhkennedy @asjohnston-asf - I didn't test the upload / download of the json directly - should I? If I should, let me know roughly how. I already did the mostly hard part of mocking everything. Thanks for your help with this.

tools/RAiDER/cli/raider.py
```diff
@@ -17,6 +18,8 @@
 from RAiDER.cli.validators import DateListAction, date_type
 from RAiDER.models.allowed import ALLOWED_MODELS
 from RAiDER.utilFcns import get_dt
+from RAiDER.aria.prepFromGUNW import main as GUNW_prep
+from RAiDER.aria.calcGUNW import tropo_gunw_slc as GUNW_calc
```
cmarshak, line 22: I moved this outside the function Brett so monkey patching would work.

We have `calcGUNW` and `GUNW_calc` and ... - would be in our interest to have better names for these @bbuzz31

asjohnston-asf:
:+1: Adding the `weather_model` field to the json metadata file looks like I expected, no concerns there. The unit test does a good job of showing what the expected behavior is.

At a glance, unit testing for the top-level `calcDelaysGUNW` that does the download/upload looks like a bit of a challenge, so I doubt I'd bother if I were making the change myself.

If you did want to test `calcDelaysGUNW`, I'd recommend reaching for [pytest-mock](https://pypi.org/project/pytest-mock/) and/or [unitest.mock.MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) to mock `get_s3_file()` and `upload_file_to_s3` during your test, then assert that they were called with the expected parameters after the test.

asjohnston-asf:
@cmarshak Here's an example of what more thorough mocking using `pytest-mock` might look like. In addition to mocking the functions to yield specific return values, you can also assert that the mocked functions were called with the expected inputs.

Also leverages pytest's build-in [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) fixture to simplify handling of the temporary json file.

The patching got a little weird since `RAiDER.aria` isn't a package; I added an `__init__.py` and tweaked some imports for demonstration purposes, but I imagine it's risky to adopt those changes as-is without more vetting.

https://github.com/asjohnston-asf/RAiDER/commit/aa9cd8d367e32dbeea0a16f0fb0e0fc9af53265f

cmarshak:
@asjohnston-asf - thank you for your help with this. I just merged your fork here into mine.

Ok, learned a lot about pytest-mocker and mock. I see how side-effects can be used to simulate the return of multiple calls to the same function and pytest-mocker can check these were made with the right inputs and in a particular order.

Really, thank you so much - I really enjoyed learning about this. I reached for `monkeypatch` because I was using pytest - but I really see how this really helps with this type of work. And I didn't know about the `read_text` function in pathlib - very nice to know.

cmarshak:
@bbuzz31 - please review when convenient - I reorganized some of the GUNW code and I want you to be aware as you are the owner of this workflow. Once you review - it's ready to merge.

bbuzz31 (approved):
LGTM

PR #523, opened 2023-04-25
Add DEM docstring

MLicari1:
Made a more detailed doc string for the download DEM function.

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/dem.py
+++ b/tools/RAiDER/dem.py
@@ -57,7 +57,18 @@ def download_dem(
     buf=0.02,
     overwrite=False,
 ):
-    """  Download a DEM if one is not already present. """
+    """  
+    Download a DEM if one is not already present. 
+    Args:
+            llbounds: list/ndarry of floats   -lat/lon bounds of the area to download. Values should be ordered in the following way: [S, N, W, E]
+            writeDEM: boolean                 -write the DEM to file
+            outName: string                   -name of the DEM file   
+            buf: float                        -buffer to add to the bounds
+            overwrite: boolean                -overwrite existing DEM
+    Returns:
+            zvals: np.array         -DEM heights
+            metadata:               -metadata for the DEM
+    """
     if os.path.exists(outName) and not overwrite:
         logger.info('Using existing DEM: %s', outName)
         zvals, metadata = rio_open(outName, returnProj=True)


MLicari1:
@jlmaurer Here is the pr for the doc string we talked about at our meeting. Let me know if should make any changes to it. Thank you!

tools/RAiDER/dem.py
```diff
@@ -57,7 +57,18 @@ def download_dem(
...
+    Download a DEM if one is not already present. 
+    Args:
+            llbounds: list/ndarry of floats   -lat/lon bounds of the area to download. Values should be ordered in the following way: [S, N, W, E]
+            writeDEM: boolean                 -write the DEM to file
+            outName: string                   -name of the DEM file   
+            buf: float                        -buffer to add to the bounds
+            overwrite: boolean                -overwrite existing DEM
+    Returns:
+            zvals: np.array         -DEM heights
+            metadata:               -metadata for the DEM
```
jlmaurer, line 70: @MLicari1 could you add some details to what is in the metadata? You can check the rasterio documentation for examples.

tools/RAiDER/dem.py
```diff
@@ -57,7 +57,18 @@ def download_dem(
     buf=0.02,
     overwrite=False,
 ):
-    """  Download a DEM if one is not already present. """
+    """  
+    Download a DEM if one is not already present. 
+    Args:
+            llbounds: list/ndarry of floats   -lat/lon bounds of the area to download. Values should be ordered in the following way: [S, N, W, E]
+            writeDEM: boolean                 -write the DEM to file
+            outName: string                   -name of the DEM file
```
jlmaurer, line 65: Please add that if the DEM exists the default behavior is to use the existing.

jlmaurer (commented):
@MLicari1 few small suggestions. Also would you be able to add a unit test that would specifically test the existing DEM behavior? You can use the DEM in `test/scenario_0/geom/` as the example. Add the test to `test/test_dem.py` and use 
```
from test import TEST_DIR
SCENARIO_DIR = os.path.join(TEST_DIR, "scenario_0")
file = os.path.join(SCENARIO_DIR, 'geom', 'warpedDEM.dem')
```
to grab the file for the test.

tools/RAiDER/dem.py
MLicari1: Will do 👍. I should have some time tomorrow to work on these

jlmaurer (approved):

PR #524, opened 2023-04-25
Kyue add docstring

ngohikennyyue:
## Type of change

## Checklist:


tools/RAiDER/dem.py
```diff
@@ -22,9 +22,19 @@
 
 
 def getHeights(ll_bounds, dem_type, dem_file, lats=None, lons=None):
```
jlmaurer, line 24: We are actually not using this function at all now, so I suggest to remove it.

jlmaurer (commented):
@ngohikennyyue thanks for submitting! There is another PR that also covers the docstring on download_dem so I will wait for that one to merge and then compare this one.

jlmaurer:
Handled by #523

Issue #518, opened 2023-04-17
[BUG] weather model bounds too small for larger grid sizes


bbuzz31:
for a user input `cube_spacing_in_m` of 50000, the interpolators (weather model locations) do not fully cover the study area.
For the config file below,  buffer has to be increased in the far range

For HRRR, both the near range and far range (and possible y direction) need to be increased.

Note this should not affect GUNW workflow which uses a smaller (weather model dependent) `cube_spacing_in_m` 
```
aoi_group:
  bounding_box:
  - 33
  - 34
  - -118.25
  - -117.25
cube_spacing_in_m: '50000.0'
date_group:
  date_list: '20200130'
download_only: false
height_group:
  height_levels:
  - -500
  - 0
  - 500
  - 1000
look_dir: right
los_group:
  orbit_file: ./S1B_OPER_AUX_POEORB_OPOD_20210317T025713_V20200129T225942_20200131T005942.EOF
  ray_trace: true
time_group:
  interpolate_time: false
  time: '13:52:45'
weather_model: ERA-5
```
Download the orbit file with sentineleof `eof -d 20200130 -m S1B`

dbekaert:
@bbuzz31 few thoughts:
The bbox is provided already and as long as the spacing is constructing a grid within the BBOX then you should be ok.
So I assume that the grid is changing the bbox to be bigger than the provided BBox. If that is the case then that new bbox extent is what should be get fed into the buffer code.
Separately we make it at least a weather model node bigger so we can translate the model cube to the user defined cube.
I also suspect this issue should not be model specific, but more a function of more apparent issue depending on high model resolution.

PR #525, opened 2023-04-25
account for grid spacing before downloading wm

bbuzz31:
#518

### DIFF
--- a/test/test_synthetic.py
+++ b/test/test_synthetic.py
@@ -177,7 +177,7 @@ class StudyArea(object):
         self.wmObj = modelName2Module(self.wmName.upper().replace("-", ""))[1]()
 
         aoi = BoundingBox(self.SNWE)
-        aoi.add_buffer(buffer = 1.5 * self.wmObj.getLLRes())
+        aoi.add_buffer(self.wmObj.getLLRes())
 
         self.los  = Raytracing(self.orbit, time=self.dt)
 
--- a/test/test_synthetic.py
+++ b/test/test_synthetic.py
@@ -176,7 +176,11 @@ class StudyArea(object):
 
         self.wmObj = modelName2Module(self.wmName.upper().replace("-", ""))[1]()
 
+        self.hgt_lvls        = np.arange(-500, 9500, 500)
+        self._cube_spacing_m = 10000.
+
         aoi = BoundingBox(self.SNWE)
+        aoi._cube_spacing_m = self._cube_spacing_m
         aoi.add_buffer(self.wmObj.getLLRes())
 
         self.los  = Raytracing(self.orbit, time=self.dt)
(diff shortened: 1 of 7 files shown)


tools/RAiDER/cli/raider.py
```diff
@@ -209,8 +211,8 @@ def calcDelays(iargs=None):
     aoi = params['aoi']
     model = params['weather_model']
 
-    # add a small buffer
-    aoi.add_buffer(buffer = 1.5 * model.getLLRes())
+    # adjust user requested AOI by grid size and buffer slightly
+    aoi.add_buffer(model.getLLRes())
```
dbekaert, line 215: I am unclear why you modify this to be only 1* . Some of these resolutions are approximate and ensuring its >1 but technically <2 would make sure there is at least a grid cell added.

tools/RAiDER/cli/raider.py
```diff
@@ -97,7 +97,7 @@ def read_template_file(fname):
                 raise ValueError(f"Unknown look direction {value}")
             template['look_dir'] = value.lower()
         if key == 'cube_spacing_in_m':
-            template[key] = float(value)
+            template[key] = float(value) if isinstance(value, str) else value
```
jlmaurer, line 100: Is this needed? There is also the case of passing `int`s etc. If a general object is passed you don't necessarily want to allow the code to proceed.

tools/RAiDER/cli/raider.py
jlmaurer: I agree with @dbekaert  here. You probably want at least small amount >1.

tools/RAiDER/delay.py
```diff
@@ -119,24 +114,20 @@ def tropo_delay(
...
-
+    out_snwe    = transform_bbox(aoi.bounds(), src_crs=4326, dest_crs=crs)
     logger.debug(f"Output SNWE: {out_snwe}")
-    logger.debug(f"Output cube spacing: {out_spacing}")
 
     # Build the output grid
     zpts = np.array(heights)
-    xpts = np.arange(out_snwe[2], out_snwe[3] + out_spacing, out_spacing)
-    ypts = np.arange(out_snwe[1], out_snwe[0] - out_spacing, -out_spacing)
+    xpts = np.arange(out_snwe[2], out_snwe[3] + aoi._output_spacing, aoi._output_spacing)
```
jlmaurer, line 129: It's not generally good practice to have `*._*` objects used by code outside the object. What I normally do is define a method e.g. "spacing()" that returns the underscored value.

tools/RAiDER/cli/raider.py
bbuzz31: sorry this was a typo

jlmaurer (approved):
@bbuzz31 this seems fine to me now. Are you still waiting on anything?

jlmaurer:
@bbuzz31 @dbekaert looks like there is a small bug to fix, but otherwise I think this should be fine.

bbuzz31:
> @bbuzz31 this seems fine to me now. Are you still waiting on anything?

Nope i think we should be good after the tests run. Thanks @jlmaurer

test/test_synthetic.py
```diff
@@ -176,8 +176,12 @@ def __post_init__(self):
 
         self.wmObj = modelName2Module(self.wmName.upper().replace("-", ""))[1]()
 
+        self.hgt_lvls        = np.arange(-500, 9500, 500)
+        self._cube_spacing_m = 10000.
+
         aoi = BoundingBox(self.SNWE)
-        aoi.add_buffer(buffer = 1.5 * self.wmObj.getLLRes())
+        aoi._cube_spacing_m = self._cube_spacing_m
+        aoi.add_buffer(self.wmObj.getLLRes())
```
bbuzz31, line 184: Note that the 1.5 * is now done within the `add_buffer` method itself

PR #527, opened 2023-04-28
Create a function to build the ray

bbuzz31:
I've isolated the functionality to build the rays between weather model layers to its own function. I then use this function directly in the synthetic test. Feedback is welcome

### DIFF
--- a/test/test_synthetic.py
+++ b/test/test_synthetic.py
@@ -4,7 +4,7 @@ from datetime import datetime
 
 from RAiDER.llreader import BoundingBox
 from RAiDER.models.weatherModel import make_weather_model_filename
-from RAiDER.losreader import Raytracing, getTopOfAtmosphere
+from RAiDER.losreader import Raytracing, build_ray
 from RAiDER.utilFcns import lla2ecef, ecef2lla
 from RAiDER.cli.validators import modelName2Module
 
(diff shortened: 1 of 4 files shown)


test/test_synthetic.py
```diff
@@ -113,45 +113,11 @@ def length_of_ray(target_xyz:list, model_zs, los):
...
-            low_xyz = getTopOfAtmosphere(xyz, LOS, low_ht, factor=cos_factor)
-            high_xyz = getTopOfAtmosphere(xyz, LOS, high_ht, factor=cos_factor)
-
-            ray_length = np.linalg.norm(high_xyz - low_xyz, axis=-1)
-
-            outSubs   += ray_length
-
-            cos_factor = (high_ht - low_ht) / ray_length if cos_factor is None else cos_factor
-
+        ray_lengths = build_ray(model_zs, ht, xyz, LOS)[0]
```
dbekaert, line 119: My main comment here is that as part of testing you are ignoring the optional arguments to the function. You should test them...

dbekaert (dismissed):
Per comments above

tools/RAiDER/losreader.py
```diff
@@ -725,4 +725,65 @@ def get_orbit(orbit_file, ref_time, pad):
 
     orb = isce.core.Orbit(svs_i)
 
-    return orb
\ No newline at end of file
+    return orb
+
+
+def build_ray(model_zs, ht, xyz, LOS, MAX_TROPO_HEIGHT=50000):
```
jlmaurer, line 731: I would specify the default max_tropo_height to be the value we define in the constants module. I think it's called Zmax? You can import that constant and then use it as a default value.

jlmaurer: I would also rename `ht` and `xyz` to be more explicit if this is the model xyz, the points xyz, etc. At the moment I'm confused what `ht` is. In the same vein definitely suggest to add a docstring with the args and return values, dtypes, and descriptions.

bbuzz31: @jlmaurer I only see Z_REF, which is 15000 and much lower than correct. it looks like this value is assigned to self._zmax in the weather models, but probably it's overwritten at some point? any thoughts on how best to proceed?

test/test_synthetic.py
bbuzz31: Agreed, but suggest we do it in another PR

bbuzz31:
> Also some proof that the code produces the same results before and after changes would be good.

This is now handled via test_slant.py

tools/RAiDER/losreader.py
bbuzz31: This is accomplished in #560

bbuzz31:
> Overall this look ok. @piyushrpt any thoughts?
> 
> My remaining comment relates to the max tropo height:
> 
>     * @bbuzz31 This is ignored in the testing, you are running the defaults. You should try setting a value too and you should test extreme cases. What if the user goes to top of atmosphere do we get more error integrated etc.
> 
>     * @jlmaurer Where are we at with exposing the max tropo height? Its default coded at many location in different functions. Has this been exposed at the top-level yet and consistently defined between Zenith and ray?
> 
>     * What is currently the value set at in the code? Do we by default always go to top of the model or not?

This is now taken care of in #560 with the caveat that rigorous testing of the effects of changing has not been conducted. A preliminary test over LA (test_slant.py) show a change in total delay of .002 m for a change in integration height from 80km to 50km at a single location at 0 elevation

jlmaurer (approved):

jlmaurer:
LGTM!

dbekaert (approved):
LGTM 
@piyushrpt let us know if you spot anything

PR #529, opened 2023-05-01
For GUNW workflow with bucket argument, does nothing if SLCs are not within Weather Model's Valid Range

### LINKED ISSUE #526: Enhancement to not update GUNW if Weather Model is Not Available for SLC Dates
For the ARIA-GUNW workflow, we do not want the operator to necessarily have to worry about the temporal availability of a weather model because the new GUNW will have additional layers that we still want to deliver to the DAAC even if RAIDER cannot be applied.

Examples:

1. Processing S1 data within 3 months of current date (no HRRR data will be available).
2. Processing the Aleutian Volcano chain but we can only go back to 2018 with HRRR there.

Will access weather model's instance variable `_valid_range` here: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/cli/raider.py#L496

Will be very straightforward.

cmarshak:
## Description
Resolves #526

(I also remove trailing white space in some of the python files that were edited).

An error (`ValueError`) is still thrown if HRRR is selected, but this is not within the coverage area - we can easily "do nothing" if this is desired (say we want to process a track across a boundary of HRRR spatial availability). 

## Motivation and Context
Complete workflow for GUNWs if SLC acquisitions do not fall into the weather model's valid date range.

## Notes
Modified `hrrr.py` in models to allow for the importing of constants (specifically the coverage areas) without instantiating the weather model classes.


tools/RAiDER/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .era5 import ERA5
+from .era5t import ERA5T
+from .gmao import GMAO
+from .hres import HRES
+from .hrrr import HRRR, HRRRAK
+
+__all__ = ['HRRR', 'HRRRAK', 'GMAO', 'ERA5', 'ERA5T', 'HRES']
```
cmarshak, line 7: Allows me to import these classes from models instead of having to use each weather model file directly.

setup.py
```diff
@@ -44,4 +44,5 @@
 setup(
     ext_modules=cython_extensions + pybind_extensions,
     cmdclass={"build_ext": build_ext},
+    package_data={'tools': ['RAiDER/models/*.zip']}
```
cmarshak, line 47: @jhkennedy - can you confirm this is done correctly?

I am using https://docs.python.org/3.11/distutils/setupscript.html#installing-package-data

Want to include a zipped file as data.

tools/RAiDER/aria/prepFromGUNW.py
```diff
@@ -10,20 +10,102 @@
...
+        else:
+            raise ValueError('HRRR was requested but it is not available in this area')
+
+    # source: https://stackoverflow.com/a/7668273
+    # Allows us to get weather models as strings
+    # getattr(module, 'HRRR') will return HRRR class
+    module = sys.modules['RAiDER.models']
+    weather_model_names = module.__all__
+    if weather_model_name not in weather_model_names:
+        raise ValueError(f'The "weather_model_name" must be in {", ".join(weather_model_names)}')
```
bbuzz31, line 94: can this be combined with `cli.validators.modelName2Module`?

bbuzz31 (approved):

tools/RAiDER/models/__init__.py
dbekaert: I thought we made it agnostic for users if its HRRR it can do conus and alaska. Adding it under allowed models might break assumptions from @jlmaurer

cmarshak: > Can you combine this with models.allowed.ALLOWED_MODELS?

the `__all__` reserved keyword is for importing from a module. It allos me to write `from RAiDER.models import HRES` rather than `from RAiDER.models.hres import HRES`. It's not changing any functionality.

That said, I believe it is good practice to decouple the constant `ALLOWED_MODELS` and those objects listed in `__all__` (partly because flake8 says so: https://stackoverflow.com/a/49266468). I don't think we want `__all__ = ALLOWED_MODELS + ...`. Also, I am including `HRRRAK` and we could potentially add more constants or functions not all of which are models, so the reverse `ALLOWED_MODELS = RAiDER.models.__all__` would not work.

> I thought we made it agnostic for users if its HRRR it can do conus and alaska. Adding it under allowed models might break assumptions from @jlmaurer

From what I can see, there is an instance method for [`_fetch`](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/models/hrrr.py#L82-L102) and this does what you are discussing, i.e. fetching the weather model files for `CONUS` or `AK` depending on the bounds. I am using properties of the weather model instances. Unfortunately, this does not impact the instance properties. To update the classes to dynamically update the properties I am using (`_date_range`) could be a lot more work.

tools/RAiDER/aria/prepFromGUNW.py
cmarshak: These functions are built in so let's just leave it for now (until another PR). I would have to modify the said function because HRRRAK is in the file HRRR (`cli.validators.modelName2Module` assumes weather models are in the same file as their name). Let's keep the `cli` and the GUNW workflow separate for now.

The more I think about this, my opinion is that there should be some constants that can be shared and imported without requiring us to access module attributes or instantiating classes. We should just have a dictionary or variable for valid date range for each model. Initializing an instance of a model is not great particularly because a a given weather model doesn't get initialized based on the bounding box of the AOI so we have to write in control flow anyways based on constants regardless. I think this should be a small separate PR and we can ask Jeremy what he thinks.

PR #532, opened 2023-05-03
update GUNW test to be agnostic of data shape

bbuzz31:
the GUNW test is failing on dev because the new buffering from #525 changes the `golden` affine transform.
I've updated the test to calculate manually an affine transform and compare that to the one embedded in the GUNW.
This update to the test is backward compatible. Also note the original purpose of this test was merely to check whether the CRS transformation was correctly set in the GUNW such that it can be read by rasterio / xarray.


dbekaert (approved):
LGTM @bbuzz31

PR #538, opened 2023-05-20
Fix hrrr xy

jlmaurer:
- Provides for querying troposphere delays with HRRR-AK

## Description

- implements a standard API for HRRR and HRRR-AK
- Fixes problem with subsetting (keeps lons to be 0-360 when crossing dateline)
- added a `RuntimeError` when trying to subset longitudes -180-180 when crossing the dateline
- Few clean-ups

## Motivation and Context

HRRR-AK wasn't working before because the subsetting was incorrect when the bounds cross the dateline. Leaving the longitudes in 0-360 form allows the subsetting to work properly

## How Has This Been Tested?

-added a couple unit tests; tests pass locally

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have added an explanation of what your changes do and why you'd like us to include them.
- [x] I have written new tests for your core changes, as applicable.
- [x] I have successfully ran tests with your changes locally.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #541, opened 2023-05-25
Fix ecmwf

bbuzz31:
ERA5 failed during cropping of the weather model to the user specified AOI due to changes in how the coordinate arrays are set up (#538).

I've corrected the shape broadcasting for ERA5 to pass, and verified that delays produced on this PR for ERA5, HRES, GMAO and HRRR are equivalent to delays produced on dev prior to #538  (#525).

Tests pass locally.

### DIFF
--- a/tools/RAiDER/models/ecmwf.py
+++ b/tools/RAiDER/models/ecmwf.py
@@ -104,7 +104,7 @@ class ECMWF(WeatherModel):
         self._lons, self._lats = np.meshgrid(lons, lats)
 
         # ys is latitude
-        self._get_heights(self._lats, hgt)
+        self._get_heights(self._lats, hgt.transpose(1, 2, 0))
         h = self._zs.copy()
 
         # We want to support both pressure levels and true pressure grids.
@@ -317,8 +317,7 @@ class ECMWF(WeatherModel):
         geo_hgt = z / self._g0
 
         # re-assign lons, lats to match heights
-        _lons = lons
-        _lats = lats
+        self._lons, self._lats = np.meshgrid(lons, lats)
 
         # correct heights for latitude
         self._get_heights(_lats, geo_hgt)
--- a/tools/RAiDER/models/ecmwf.py
+++ b/tools/RAiDER/models/ecmwf.py
@@ -71,6 +71,7 @@ class ECMWF(WeatherModel):
         '''
         self._load_model_level(*self.files)
 
+
     def _load_model_level(self, fname):
         # read data from netcdf file
         lats, lons, xs, ys, t, q, lnsp, z = self._makeDataCubes(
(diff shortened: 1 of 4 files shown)


tools/RAiDER/models/ecmwf.py
```diff
@@ -101,11 +101,10 @@ def _load_model_level(self, fname):
         self._q = q
         geo_hgt, pres, hgt = self._calculategeoh(z, lnsp)
 
-        # re-assign lons, lats to match heights
-        _lons = np.broadcast_to(lons[np.newaxis, np.newaxis, :], hgt.shape)
-        _lats = np.broadcast_to(lats[np.newaxis, :, np.newaxis], hgt.shape)
+        self._lons, self._lats = np.meshgrid(lons, lats)
```
bbuzz31, line 104: this is needed for consistency with HRRR; a 2D array should be used

tools/RAiDER/models/ecmwf.py
```diff
@@ -101,11 +101,10 @@ def _load_model_level(self, fname):
...
+
         # ys is latitude
-        self._get_heights(_lats, hgt)
+        self._get_heights(self._lats, hgt.transpose(1, 2, 0))
```
bbuzz31, line 107: this, I thought, should make the function work as internally in _get_heights we add a dimension to the last axis and then stack the lat/lons in height

tools/RAiDER/models/weatherModel.py
```diff
@@ -324,7 +324,10 @@ def _get_heights(self, lats, geo_hgt, geo_ht_fill=np.nan):
         Transform geo heights to WGS84 ellipsoidal heights
         '''
         geo_ht_fix = np.where(geo_hgt != geo_ht_fill, geo_hgt, np.nan)
-        lats_full = np.broadcast_to(lats[...,np.newaxis], geo_ht_fix.shape)
+        print ('LATS SHAPE', lats.shape)
```
bbuzz31, line 327: some diagnostics to show the shapes; HRRR conus (which works) and ERA5 are aligned dimension wise but the latter doesn't work

tools/RAiDER/models/ecmwf.py
piyushrpt: You don't necessarily have to use 2D. If memory is limited or you want to keep footprint down, use numpy stride tricks:

https://towardsdatascience.com/advanced-numpy-master-stride-tricks-with-25-illustrated-exercises-923a9393ab20

tools/RAiDER/models/weatherModel.py
```diff
@@ -324,7 +324,10 @@ def _get_heights(self, lats, geo_hgt, geo_ht_fill=np.nan):
...
+        print (geo_ht_fix.shape)
+        lats_full  = np.broadcast_to(lats[...,np.newaxis], geo_ht_fix.shape)
```
bbuzz31, line 329: this line is needed for HRRR

jlmaurer (approved):
LGTM as long as relevant tests pass

jlmaurer:
@bbuzz31 do we have any tests that test with ECMWF models currently?

bbuzz31:
@jlmaurer some of the tests have a variable (WM) that can be changed to ERA-5. Sometimes ERA-5 takes many hours to download.

In #543 I'm reusing downloaded ERA5 models for `test_intersect.py` so that is a partial test.

jlmaurer:
Ok great. Do those tests pass? The problem here is one of ordering.

bbuzz31:
Yup they pass

PR #544, opened 2023-05-25
check spatial validity early on and set HRRR-AK

### LINKED ISSUE #542: [BUG] No weather model available over Anchorage (Track 131) - HyP3
**Describe the bug**

Here is the job submission data:

```
{'job_id': '0ea9da74-f20e-45ef-8602-c424f82b4988',
  'job_type': 'INSAR_ISCE_TEST',
  'request_time': '2023-05-24T23:49:30+00:00',
  'status_code': 'FAILED',
  'user_id': 'cmarshak',
  'name': 'Anchorage_131_HRRR_0524',
  'job_parameters': {'compute_solid_earth_tide': True,
   'estimate_ionosphere_delay': True,
   'frame_id': 20343,
   'granules': ['S1A_IW_SLC__1SDV_20220210T162936_20220210T163004_041853_04FB8A_4E1F',
    'S1A_IW_SLC__1SDV_20220210T163002_20220210T163029_041853_04FB8A_8ED8'],
   'secondary_granules': ['S1A_IW_SLC__1SDV_20220117T162937_20220117T163004_041503_04EF80_0BFA',
    'S1A_IW_SLC__1SDV_20220117T163002_20220117T163029_041503_04EF80_BC1F'],
   'weather_model': 'HRRR'}
```

This does not seem correct given the SLCs are in the HRRR-AK's valid range of 2018 - present.

(Note 15 jobs were submitted and all failed - I checked 3 and they all had the same error message as the one provided above).

**To Reproduce**

See #503 for details on how to get input GUNW.

bbuzz31:
HRRR-AK is only available at 3 hr sampling intervals, and Herbie won't automatically figure out the nearest time. 
Rearranged RAiDER a bit to check right away whether the user requested bounds (after buffering) are within the requested weather model bounds. If they're not, it checks whether HRRR-AK was specified and updates the model internally.

N.B. This is built directly on dev and is independent of my other PRs and test pass locally.
Closes #542


jlmaurer:
@bbuzz31 this functionality should already be present, i.e. RAiDER will already query HRRR-AK if the user specifies HRRR and the bbox is over Alaska. See [here](https://github.com/dbekaert/RAiDER/blob/eb5a478a121b77273633764d986b44bcd94a32e9/tools/RAiDER/models/hrrr.py#L96) and following.

bbuzz31:
@jlmaurer Yes, but this happens too late in the code for non-3hr interval datetimes. `model.dtime()` which gets the sampling availability, happens before the weather model is downloaded. Since we use HRRR for both conus and alaska, this will go to the conus sampling and then fail.

Try examples where model = HRRR, bounding box is in alaska, and datetime is NOT in (0, 3, 12, 15, 18, 21)

jlmaurer:
Ah I see...that is tricky! I wonder if a slightly cleaner solution would be to pass a box to the dtime function? To avoid having to name weather models in the run program

bbuzz31:
Yeah you could do it that way. I thought it was better to split off HRRR-AK. Not sure what you mean by name weather models?

tools/RAiDER/models/weatherModel.py
```diff
@@ -487,6 +487,10 @@ def checkValidBounds(
         Returns:
         bool    True if the ll_bounds represent a valid area for the weather model
         '''
+        if self._Name == 'HRRR-AK':
```
jlmaurer, line 490: This would be better implemented as a model-specific method

tools/RAiDER/cli/raider.py
```diff
@@ -219,10 +219,20 @@ def calcDelays(iargs=None):
...
     if los.ray_trace():
-        wm_bounds = aoi.calc_buffer_ray(los.getSensorDirection(), lookDir=los.getLookDirection(), incAngle=30)
+        wm_bounds = aoi.calc_buffer_ray(los.getSensorDirection(),
+                            lookDir=los.getLookDirection(), incAngle=30)
     else:
         wm_bounds = aoi.bounds()
 
+    ## check the bounds of the weather model are okay
+    if not model.checkValidBounds(wm_bounds):
+        if model._Name == 'HRRR':
```
jlmaurer, line 229: This is what we want to avoid if possible.

jlmaurer (commented):
Somehow we need to setup the check so we don't have to explicitly name weather models in the api. Perhaps an earlier check in the base hrrr model, or just require an explicit hrrr-ak call, since none of this is an issue if hrrr-ak is separately defined.

bbuzz31:
Closed in favor of #547

PR #547, opened 2023-06-05
HRRR-AK2

bbuzz31:
HRRR-AK is only available at 3 hr sampling intervals, and Herbie won't automatically figure out the nearest time.
Rearranged RAiDER a bit to check right away whether the user-requested bounds are within the requested weather model bounds. If they're not, it checks whether HRRR-AK was specified and updates the model internally.

Closes https://github.com/dbekaert/RAiDER/issues/542

### DIFF
--- a/tools/RAiDER/models/weatherModel.py
+++ b/tools/RAiDER/models/weatherModel.py
@@ -485,9 +485,21 @@ class WeatherModel(ABC):
         ll_bounds : np.ndarray
 
         Returns:
-        bool    True if the ll_bounds represent a valid area for the weather model
+        bool    The weather model object
         '''
-        return box(ll_bounds[2], ll_bounds[0], ll_bounds[3], ll_bounds[1]).intersects(self._valid_bounds)
+        S, N, W, E = ll_bounds
+        if box(W, S, E, N).intersects(self._valid_bounds):
+            Mod = self
+
+        elif self._Name == 'HRRR':
+            from RAiDER.models.hrrr import HRRRAK
+            Mod = HRRRAK()
+            if not box(W, S, E, N).intersects(Mod._valid_bounds):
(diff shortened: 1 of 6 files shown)


tools/RAiDER/models/weatherModel.py
```diff
@@ -485,9 +485,23 @@ def checkValidBounds(
         ll_bounds : np.ndarray
 
         Returns:
-        bool    True if the ll_bounds represent a valid area for the weather model
+        bool    The weather model object
         '''
-        return box(ll_bounds[2], ll_bounds[0], ll_bounds[3], ll_bounds[1]).intersects(self._valid_bounds)
+        S, N, W, E = ll_bounds
```
jlmaurer, line 490: This looks great; only I would override the function in the hrrr module. So this function stays the same, but you define a new method with the same name in the `HRRR` weather model class with everything under line 494.

jlmaurer (commented):
LGTM except one easy comment!

jlmaurer (commented):
LGTM except one easy comment!

tools/RAiDER/delay.py
```diff
@@ -229,9 +228,8 @@ def _build_cube_ray(
     epsg4326 = CRS.from_epsg(4326)
     cube_to_llh = Transformer.from_crs(pts_crs, epsg4326,
                                        always_xy=True)
-    ecef_to_model = Transformer.from_crs(CRS.from_epsg(4978), model_crs)
-    # For calling the interpolators
-    flip_xy = model_crs.axis_info[0].direction == "east"
```
bbuzz31, line 234: @piyushrpt @jlmaurer and I couldn't find any use for this... do you remember why it was here, perhaps for NCMR?

jlmaurer (approved):

Issue #535, opened 2023-05-11
Remove test dependency on GMAO


bbuzz31:
Downloading API should be tested in a single test; all others should not require downloading weather model.
Possible options are staging data in S3, making unit tests smaller with simple embedded array tests, leveraging https://git-lfs.com/

PR #548, opened 2023-06-07
Reinstate test3

bbuzz31:
This reinstates the ray tracing and slant projection test, comparing a single location to a golden value and using the already staged weather files.

It also exposes the maximum tropo height that is used for the ray tracing integral, and sets the new default to 80 km (used to be 50 km).

~The synthetic test pass for 50 km but fails for 80 km in Alaska because the buffering is not big enough. I've confirmed the buffering is big enough when you change 80 km to 100 km in the equation we're using (see [maxZ](https://github.com/dbekaert/RAiDER/blob/faf6a115c09a2a31b25e343ab3d7aa3a430ccea2/tools/RAiDER/llreader.py#LL159C2-L162C2)) but am unsure what the best way to address this is.~

The synthetic test failed because the ERA5 CDSAPI truncates our desired grid by one cell. This is now accounted for in the buffering.

Note this is built on #547 and #551 

Part of #535


bbuzz31:
Replaced by #560

PR #564, opened 2023-07-12
HRRR Extent

### LINKED ISSUE #562: HRRR bounds issue [BUG]
**Describe the bug**
The `raider.py` program fails when I pass a `bounding_box` extent encompassing CONUS. Looks like we will have to adjust the padding behavior [here](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/models/weatherModel.py#L206) but am not sure how best to proceed. Perhaps previous ticket #552 is of relevance?

**To Reproduce**
Using the attached `yaml` file, run raider:
raider.py raider.yaml

 - RAiDER git tag: 9ddd4abad8414b3670194b4c08220f150d7ba00d
 - OS

[raider.yaml.txt](https://github.com/dbekaert/RAiDER/files/12009207/raider.yaml.txt)

bbuzz31:
Because of a quirk in the HRRR projection, a user requested bounding box could appear to be valid but actually be outside of the model domain.
This fixes the initial check to make sure the requested aoi is contained in HRRR domain.

UPDATE:
The code will give a warning if HRRR doesn't cover the requested AOI but continue processing. 
It will also write delays (and continue processing if requested) even if they contain nans

Closes #562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Latest updates:
++ check that the aoi is contained within HRRR extent
 + move the ray building out of the _build_cube_ray and into its own function for cleaner testing
 + update the tests to use the new build_ray function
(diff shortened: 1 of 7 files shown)


jlmaurer (commented):
I think this looks good to me!

jlmaurer (approved):

jlmaurer (approved):
Looks good!

PR #565, opened 2023-07-13
Update unit tests to remove some downloading

jlmaurer:
This PR focuses on streamlining unit tests to avoid downloading when possible

## Description

- removes downloads from test_datelist

## Motivation and Context

- repeated download failures make it hard to test

## How Has This Been Tested?

- updated tests pass locally

## Type of change

- [ x] unit test update

## Checklist:


jlmaurer:
@bbuzz31 can you take a quick look at the error message on this? This PR is passing locally and I ran the test independently and completes just fine.

test/scenario_5/temp.yaml
bbuzz31, line 1: can / should this yaml file be created on the fly like the other ones?

bbuzz31: Also, can/should it point to the global testing WM_DIR (RAiDER/test/weather_files)?

test/test_GUNW.py
bbuzz31, line 1: This test will get skipped due to `@pytest.mark.long?` yes?
I think we'd prefer that not to happen, as the GUNW workflow will be cranking lots of products

bbuzz31 (approved):

PR #566, opened 2023-07-18
Fix fillna3d

jlmaurer:
Fix a bug in `fillna3D` that resulted in too large of integrated delays

## Description

`fillna3D` was designed to infill NaNs that are below topography with the nearest non-NaN value. However, there was previously no check for whether there were NaNs _above_ the highest elevation provided by the weather model. For most models this is not an issue because the weather model extends to the highest levels we use in RAiDER. However, HRRR only extends to ~20 km, which left up to 60 km of heights compared to our standard fixed heights (e.g., `LEVELS_137_HEIGHTS`). This fixes the issue by replace "upper" NaNs with 0s, except for temperature, which occurs in the denominator of the refractivity equation, so we replace it with a very high positive value. 

## How Has This Been Tested?

- verified that the issue is correctly handled and fixed in HRRR
- no existing unit tests tested this functionality. I will add unit tests to my other PR but think this is important enough to push now. 

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Latest updates:
++ Fix bug in fillna3D for NaNs at elevations higher than present in the weather model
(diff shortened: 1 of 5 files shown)


bbuzz31 (approved):
LGTM thanks @jlmaurer

PR #569, opened 2023-07-19
Tests for new fillna

bbuzz31:
This should have all tests back up and working.

staged weather model files had to be updated and golden delay values had to be adjusted by (by less than a millimeter) in a few places to account for new interpolation scheme


jlmaurer (approved):
Great!

tools/RAiDER/processWM.py
```diff
@@ -86,7 +86,7 @@ def prepareWeatherModel(
         )
 
         containment = weather_model.checkContainment(ll_bounds)
-        if not containment:
+        if not containment and not weather_model._Name.startswith('HRRR'):
```
jlmaurer, line 89: Is there a reason why HRRR here and not other models?

bbuzz31: Well the other models are global, so I figured if something is outside there's a problem somewhere (like aoi specified incorrectly)

jlmaurer: That's a good point, but what if we or someone else introduces another regional model? would there be a way to be more general?

bbuzz31: I put the ~inverse for now (If it the weather model is GMAO ERA5/T or HRES and its not contained it will crash).
Maybe we could add an attribute to the weather model class like `is_global` but I don't see an easy way of doing it smartly

PR #570, opened 2023-07-24
HRRR model levels 

### LINKED ISSUE #568: [BUG] fix era5 pressure levels
they're out of order

bbuzz31:
Use the native model levels rather than the pressure levels in order to cover a more complete height of the troposphere.

Closes #568


dbekaert:
Please check the unit test, specifically error related to the model levels not being defined.

tools/RAiDER/models/era5.py
```diff
@@ -29,7 +29,7 @@ def __init__(self):
         self._lag_time = relativedelta(months=lag_time)
 
         # Default, need to change to ml
-        self.setLevelType('ml')
+        self.setLevelType('pl')
```
dbekaert, line 32: This is a change on default behaviour not consitent with the PR. Where possible we want to be using model levels.

tools/RAiDER/models/hrrr.py
```diff
@@ -226,6 +232,25 @@ def __init__(self):
         self._proj = p1
 
         self._valid_bounds =  Polygon(((-125, 21), (-133, 49), (-60, 49), (-72, 21)))
+        self.setLevelType('ml')
```
dbekaert, line 235: This we should be coupling to the "nat" variable as that could be changed and this could still be coded up to say model levels.

 So i would catch it here in a statement and if prs is provided it should go differently and could throw and error saying this does not reach high enough altitude or so?

tools/RAiDER/models/hrrr.py
```diff
@@ -341,6 +365,25 @@ def __init__(self):
...
+        '''Set the level type to model levels or pressure levels'''
+        if levelType in ['ml', 'pl']:
+            self._model_level_type = levelType
+        else:
+            raise RuntimeError(f'Level type {levelType} is not recognized')
+
+        if levelType == 'ml':
+            self.__model_levels__()
+        else:
+            raise NotImplementedError('Pressure levels do not go high enough for HRRR.')
```
dbekaert, line 381: good

tools/RAiDER/models/hrrr.py
```diff
@@ -341,6 +365,25 @@ def __init__(self):
...
+
+
+    def __model_levels__(self):
+        self._levels  = 137
+        self._zlevels = np.flipud(LEVELS_137_HEIGHTS)
```
dbekaert, line 386: 50 or 137? Believe HRRR only has 51 model levels? https://rapidrefresh.noaa.gov/faq/HRRR.faq.html

tools/RAiDER/models/model_levels.py
```diff
@@ -505,3 +505,17 @@
     -300,
     -500,
 ]
+
```
dbekaert, line 508: We should be documenting also A and B for HRRR to stay consistent. Look here: https://rapidrefresh.noaa.gov/faq/HRRR.faq.html

Were the average model height computed using these values?

tools/RAiDER/models/era5.py
bbuzz31: yeah i forgot to change it back after testing

tools/RAiDER/models/hrrr.py
```diff
@@ -15,12 +15,12 @@
     WeatherModel, TIME_RES
 )
 from RAiDER.models.model_levels import (
-    LEVELS_137_HEIGHTS,
+    LEVELS_50_HEIGHTS_CONUS,
```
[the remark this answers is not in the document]
dbekaert: Don't expect much differences, so why complicate? Also these are specific for HRRR using its a and b values, hence a better naming might be needed?

tools/RAiDER/models/ecmwf.py
```diff
@@ -321,39 +321,24 @@ def _load_pressure_level(self, filename, *args, **kwargs):
...
 
         # re-assign lons, lats to match heights
         self._lons, self._lats = np.meshgrid(lons, lats)
 
         # correct heights for latitude
-        self._get_heights(_lats, geo_hgt)
+        self._get_heights(self._lats, geo_hgt)
 
-        self._p = np.broadcast_to(levels[:, np.newaxis, np.newaxis],
+        self._p = np.broadcast_to(levels[np.newaxis, np.newaxis, :],
```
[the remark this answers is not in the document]
bbuzz31: It was broken before in numerous places actually because variables had been renamed and shapes realigned. I shouldve issued a separate PR but had to get it working to compare with the HRRR.

It results in sub-mm differences with ERA5 model levels.

tools/RAiDER/models/hrrr.py
bbuzz31: We don't allow for this in any of the other models as of now; i.e., we always force model levels.
I added a change to for coupling to 'nat' though to support it in the future if we want

tools/RAiDER/models/hrrr.py
bbuzz31: There are 51 sigma layers and 50 computational layers that are made available to the end user ([Table 7](https://journals.ametsoc.org/view/journals/mwre/144/4/mwr-d-15-0242.1.xml)).

tools/RAiDER/models/hrrr.py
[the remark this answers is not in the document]
bbuzz31: Updated to use the same for both and corrected the name. See comment on a/b values below

tools/RAiDER/models/ecmwf.py
```diff
@@ -321,39 +321,24 @@ def _load_pressure_level(self, filename, *args, **kwargs):
...
                                   self._zs.shape)
 
-        # Re-structure everything from (heights, lats, lons) to (lons, lats, heights)
-        self._p = np.transpose(self._p)
-        self._t = np.transpose(self._t)
-        self._q = np.transpose(self._q)
-        self._lats = np.transpose(_lats)
-        self._lons = np.transpose(_lons)
+        # Re-structure from (heights, lats, lons) to (lons, lats, heights)
```
jlmaurer, line 335: Is there a reason only q and t are transposed and not p or z?

tools/RAiDER/models/hrrr.py
```diff
@@ -228,12 +232,29 @@ def __init__(self):
...
+        self._proj = CRS(f'+proj=lcc +lat_1={lat1} +lat_2={lat2} +lat_0={lat0} '\
                  f'+lon_0={lon0} +x_0={x0} +y_0={y0} +a={earth_radius} '\
                  f'+b={earth_radius} +units=m +no_defs')
-        self._proj = p1
-
         self._valid_bounds = HRRR_CONUS_COVERAGE_POLYGON
+        self.setLevelType('ml')
+
+
+    def setLevelType(self, levelType):
```
jlmaurer, line 242: Isn't this defined in `weatherModel.py`? If not, we should move it over there so it can be available to other models.

tools/RAiDER/models/hrrr.py
```diff
@@ -228,12 +232,29 @@ def __init__(self):
...
+        '''Set the level type to model levels or pressure levels'''
+        if levelType in ['ml', 'pl']:
+            self._model_level_type = levelType
+        else:
+            raise RuntimeError(f'Level type {levelType} is not recognized')
+
+        if levelType == 'ml':
```
jlmaurer, line 249: We could also just easily change this to `if levelType in ['ml', 'nat']:`

jlmaurer: That would allow for @dbekaert 's comment earlier.

tools/RAiDER/models/hrrr.py
```diff
@@ -345,6 +367,26 @@ def __init__(self):
         # The projection information gets read directly from the  weather model file but we
         # keep this here for object instantiation.
         self._proj = HRRR_AK_PROJ
+        self.setLevelType('ml')
+
+
+    def setLevelType(self, levelType):
```
jlmaurer, line 373: With this function defined in the main weatherModel module, we don't need it defined twice. Then you would just overload the `self.__load_pressure_levels()` method in hrrr.py to raise the same `NotImplementedError` you have here.

jlmaurer (approved):
LGTM. I updated the `NotImplementedError` message on pressure levels to help me remember in the future why we aren't using them.

tools/RAiDER/models/model_levels.py
dbekaert: Check the documentation link above, I understand they are separated by the "A.B" for HRRR, equation is provided too?

tools/RAiDER/models/ecmwf.py
```diff
@@ -321,39 +308,24 @@ def _load_pressure_level(self, filename, *args, **kwargs):
         self._t = t
         self._q = q
 
-        geo_hgt = z / self._g0
+        geo_hgt = (z / self._g0).transpose(1, 2, 0)
```
bbuzz31, line 311: @jlmaurer here we transpose the geo_hgt, which then propagates to the z and p

tools/RAiDER/models/ecmwf.py
bbuzz31: p and z are indeed transposed just a bit earlier on. I marked it for you

tools/RAiDER/losreader.py
bbuzz31, line 1: if you don't specify height levels, then all the model layers will be used. the top of the model layer (ht) is by definition = high_ht at the top. So nothing is calculated and then the stacking crashes. It's a very tiny contribution to the integral.

bbuzz31:
@cmarshak any idea why the GUNW test could be failing? 

It writes the first delay successfully but then seems to kill itself during the next datetime: 
The end of stdout is as follows:

```Successfully wrote delay cube to: /root/project/test/GUNW/HRRR_tropo_20200130T135244_ray.nc

Starting to run the weather model calculation
Date: 20200124
Beginning weather model pre-processing
Weather model HRRR is available from 2016-07-15 to Present
Weather model HRRR is available from 2016-07-15 to Present
```
It should continue with, e.g.,  ```✅ Found ┊ model=hrrr ┊ product=nat ┊ 2020-Jan-30 13:00 UTC F00 ┊ GRIB2 @ aws ┊ IDX @ ```

But instead just stops and returns:

```subprocess.CalledProcessError: Command '['raider.py', '++process', 'calcDelaysGUNW', '-f', '/root/project/test/GUNW/S1-GUNW-D-R-071-tops-20200130_20200124-135156-34956N_32979N-PP-913f-v2_0_4.nc', '-m', 'HRRR', '-o', '/root/project/test/GUNW']' died with <Signals.SIGKILL: 9>```

Maybe a memory issue on circleci? I can't reproduce it locally, and a main change here was to use the `product=nat` which is bigger

cmarshak:
I would run the test on a linux server via `pytest test/test_GUNW.py`. I did this a few days ago and it ran to completion. I don't know what that sig kill means. I have seen that when `gdal` is not installed correctly. However, don't know. The integration tests are pretty hefty for HRRR.

bbuzz31:
Thanks yeah I think it has to be something weird with the memory. If I just run with HRRR (and not GMAO) it works.

PR #571, opened 2023-07-25
Azimuth time interpolation for Weather Models in GUNW Workflow

cmarshak:
Ports the azimuth time grid interpolation demonstrated [here](https://github.com/ACCESS-Cloud-Based-InSAR/s1_azimuth_time_grid) to RAiDER's GUNW workflow. Formerly the center time was used to interpolate across the entire weather cube for two weather model times. Now, it is based on the time in which the sensor passes the pixel at zero-doppler (using ISCE3 `geo2rdr`).

The main test will be to show that the `center_time` interpolation and the `azimuth_time_grid` interpolation doe not deviate more than 1 mm.

There is a change in the API for yaml (see [here](https://github.com/dbekaert/RAiDER/blob/azimuth-time-interpolation/tools/RAiDER/cli/raider.yaml#L61)) and CLI (see [here](https://github.com/dbekaert/RAiDER/blob/azimuth-time-interpolation/tools/RAiDER/cli/raider.py#L495-L502)). Specifically, updates `interpolate_time` options to: 
+ `'none'` (formerly `False`), 
+ `'center_time'` (formerly `True` and no longer `default`), and 
+ `azimuth_time_grid` (not implemented previously and now `default` only working for `HRRR` model)

This will make the default behavior for the GUNW workflow `azimuth_time_grid`, which *only* works for HRRR.


cmarshak:
@dbekaert

@sssangha and I worked on this repo together: https://github.com/ACCESS-Cloud-Based-InSAR/s1_azimuth_time_grid
I have integrated into RAiDER for HRRR in the PR. It performs an inverse weighting based on a per pixel difference between the azimuth acquisition time (zero-doppler) and the model time of the weather model. The difference between center-time interpolation and this per pixel interpolation for the troposphere phase are on the order of 1 mm (not sure if this is too large given the model times are much greater than the azimuth timing I am putting in). The test to verify this for all relevant layers for a single GUNW is illustrated here: https://github.com/dbekaert/RAiDER/blob/4e00730433b2e800cce31ebd7877584b29a408be/test/test_GUNW.py#L121-L150
 
The To Dos:
 
1. Remove the download requirement for testing models
2. Incorporate other models (only HRRR has been tested – GMAO looks doable, but not sure about others)
3. I think a few more tests for inverse weighting would be good too – namely an example where two different groups of models are used

Let us know your thoughts.

cmarshak:
@sssangha - the easiest way to use this PR is:

```
raider.py ++process calcDelaysGUNW -f <gunw_path> -m HRRR -interp azimuth_time_grid
```

cmarshak:
Here is an example GUNW [product](https://gunw-development-testing.s3.us-west-2.amazonaws.com/raider-gunw/S1-GUNW-A-R-064-tops-20210723_20210711-015000-00119W_00033N-PP-6267-v2_0_6.nc) that is also used in the test.

The command to generate is:

```
raider.py ++process calcDelaysGUNW -f S1-GUNW-A-R-064-tops-20210723_20210711-015000-00119W_00033N-PP-6267-v2_0_6.nc --weather-model HRRR -interp azimuth_time_grid
```

Below are some screenshots of the reference delay in radians for the above product.

image
image

bbuzz31:
> Here is an example GUNW [product](https://gunw-development-testing.s3.us-west-2.amazonaws.com/raider-gunw/S1-GUNW-A-R-064-tops-20210723_20210711-015000-00119W_00033N-PP-6267-v2_0_6.nc) that is also used in the test.
> 
> The command to generate is:
> 
> ```
> raider.py ++process calcDelaysGUNW -f S1-GUNW-A-R-064-tops-20210723_20210711-015000-00119W_00033N-PP-6267-v2_0_6.nc --weather-model HRRR -interp azimuth_time_grid
> ```
> 
> Below are some screenshots of the reference delay in radians for the above product.
> 
> image image

Could you please show the residual between this and the original (uncorrected for azimuth time)?

cmarshak:
I can do this in the afternoon. Please note the test suite for GUNWs include a test that shows the residual of said product above is below 1 mm. Do you want to see spatial pattern of said residual?

cmarshak:
When I run the command line argument for GUNWs, note I get the following output in the log:

```
Processing slice 1 / 20: -500.0
Processing slice 2 / 20: 0.0
Processing slice 3 / 20: 500.0
Processing slice 4 / 20: 1000.0
Processing slice 5 / 20: 1500.0
Processing slice 6 / 20: 2000.0
Processing slice 7 / 20: 2500.0
Processing slice 8 / 20: 3000.0
Processing slice 9 / 20: 3500.0
Processing slice 10 / 20: 4000.0
Processing slice 11 / 20: 4500.0
Processing slice 12 / 20: 5000.0
Processing slice 13 / 20: 5500.0
Processing slice 14 / 20: 6000.0
Processing slice 15 / 20: 6500.0
Processing slice 16 / 20: 7000.0
Processing slice 17 / 20: 7500.0
Processing slice 18 / 20: 8000.0
Processing slice 19 / 20: 8500.0
Processing slice 20 / 20: 9000.0
CRITICAL: There are missing delay values. Check your inputs.          
```

I also have weird nodata values in the HRRR.

image

Will open an issue ticket to illustrate.

cmarshak:
From @sssangha:

Here is a series of quick figures. I’ll also paste the input GUNWs + sequence of few commands I used to extract these layers. I will go ahead and pass the latter into a formal notebook later.
 
Troposphere hydrostatic layer at 500 m elevation directly from the overlapping GUNWs, for the reference date:
image

Troposphere wet layer at 500 m elevation directly from the overlapping GUNWs, for the reference date:
image

Reference date extracted and stitched tropo total via ARIA-tools:
image

Secondary date extracted and stitched tropo total via ARIA-tools:
image

Everything overall looks seamless, but upon closer inspection in the GUNWs, I found surprisingly large mean/min values.
 
E.g. for the overlap region of the troposphere hydro layer:
STATISTICS_MAXIMUM=2.3955078125
STATISTICS_MEAN=-0.000690513815912
STATISTICS_MINIMUM=-3.829345703125
STATISTICS_STDDEV=0.067453400601181
 
For the overlap region of the troposphere wet layer:
STATISTICS_MAXIMUM=2.4646682739258
STATISTICS_MEAN=5.98180550712e-06
STATISTICS_MINIMUM=-1.6872520446777
STATISTICS_STDDEV=0.03005129298574
 
When I looked at the actual residual fields for the hydro and wet layers, respectively, I found a few strange, very narrow rows with some non-zero residual:
image

image


Here are the commands I used:
# GUNWs used:
S1-GUNW-A-R-064-tops-20210723_20210711-015001-35393N_33512N-PP-6267-v2_0_4.nc
S1-GUNW-A-R-064-tops-20210723_20210711-014936-33900N_32016N-PP-53c5-v2_0_4.nc

```
# using raider to populate the GUNWs
raider.py ++process calcDelaysGUNW -f products/S1-GUNW-A-R-064-tops-20210723_20210711-015001-35393N_33512N-PP-6267-v2_0_4.nc -m HRRR -interp azimuth_time_grid
 
raider.py ++process calcDelaysGUNW -f products/S1-GUNW-A-R-064-tops-20210723_20210711-014936-33900N_32016N-PP-53c5-v2_0_4.nc -m HRRR -interp azimuth_time_grid  
 
# using ARIA-tools to extract and stitch the tropo layers
ariaExtract.py -f 'products/S1*' -l 'unwrappedPhase,troposphereTotal' -tm HRRR -d Download -w output_dir -of ISCE
```

bbuzz31:
@sssangha do you have a function handy for cropping to the overlap?

bbuzz31:
I am not seeing this residual in the delays within the GUNW from RAiDER (prior to running ARIA-tools and the intersection). I'll see if I can recreate after running AT and debug, but as the problem does not appear to be in RAiDER, I think this PR can be merged. What do you think @dbekaert?

Before #575 
image

After #575 
image

bbuzz31 (approved):
LGTM nice work @cmarshak @sssangha

bbuzz31 (changes requested):
on closer inspection, there needs to be a fail safe to set a default value for interpolation grid in the case that one does not exist in the users .yaml file (when running raider not in the GUNW workfloW)

cmarshak:
> on closer inspection, there needs to be a fail safe to set a default value for interpolation grid in the case that one does not exist in the users .yaml file (when running raider not in the GUNW workfloW)

Hey @bbuzz31 - I feel like this ask would be a separate PR. The above PR is really focused on the GUNW. Did this PR change an existing functionality for the non-GUNW workflow? I don't think interpolation is an option (yet) in the non-GUNW workflow. Regardless, this is a pretty big PR largely because of testing.

If you wanted specific changes, please give some indication of what you wanted.

bbuzz31:
> > on closer inspection, there needs to be a fail safe to set a default value for interpolation grid in the case that one does not exist in the users .yaml file (when running raider not in the GUNW workfloW)
> 
> Hey @bbuzz31 - I feel like this ask would be a separate PR. The above PR is really focused on the GUNW. Did this PR change an existing functionality for the non-GUNW workflow? I don't think interpolation is an option (yet) in the non-GUNW workflow. Regardless, this is a pretty big PR largely because of testing.
> 
> If you wanted specific changes, please give some reference in the code of what you wanted.

No it should be in here. There is no default behavior if there isn't an `interp_method` specified by the user, and no clear warning given. So prior config files will fail for nonGUNW raider workflow. Line `353` should be changed to set a  default method (I forget what we decided on) with a warning message perhaps.

tools/RAiDER/cli/raider.py
```diff
@@ -245,12 +246,27 @@ def calcDelays(iargs=None):
         # Grab the closest two times unless the user specifies 'nearest'
         # If the model time_delta is not specified then use 6
         # The two datetimes will be combined to a single file and processed
-        times = get_nearest_wmtimes(t, [model.dtime() if \
-                    model.dtime() is not None else 6][0]) if params['interpolate_time'] else [t]
```
cmarshak, line 249: @bbuzz31 - this is the default behavior currently i.e. when the `interpolate_time` is not specified. What more are you asking?

tools/RAiDER/cli/raider.py
```diff
@@ -245,12 +246,27 @@ def calcDelays(iargs=None):
...
+        interp_method = params['interpolate_time']
```
cmarshak, line 249: @bbuzz31 - when you read a yaml file with a given field and it is not specified - then he returns None so the default behavior is not changed. I will change this to `get` for further clarity.

cmarshak:
> > > on closer inspection, there needs to be a fail safe to set a default value for interpolation grid in the case that one does not exist in the users .yaml file (when running raider not in the GUNW workfloW)
> > 
> > 
> > Hey @bbuzz31 - I feel like this ask would be a separate PR. The above PR is really focused on the GUNW. Did this PR change an existing functionality for the non-GUNW workflow? I don't think interpolation is an option (yet) in the non-GUNW workflow. Regardless, this is a pretty big PR largely because of testing.
> > If you wanted specific changes, please give some reference in the code of what you wanted.
> 
> No it should be in here. There is no default behavior if there isn't an `interp_method` specified by the user, and no clear warning given. So prior config files will fail for nonGUNW raider workflow. Sections starting on Line `353` in `raider.py` should set a default method (I forget what we decided on) with a warning message perhaps.

All the tests are passing. I am still unclear what you mean by default for the non-GUNW workflow or how that is called. I am hoping that the non-GUNW workflow just reads or modifies the yaml file [here](https://github.com/dbekaert/RAiDER/blob/azimuth-time-interpolation/tools/RAiDER/cli/raider.yaml#L61)? Is there an example?

My opinion is there should be a test for the control flow that you are asking and therefore a separate PR. I assume you mean the `NotImplementedError` in Raider [here](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/cli/raider.py#L353) for 353?

This seems problematic to me: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/cli/raider.py#L286-L289 - unexpected behavior when you specify a non-nearest interpolation and only one file is returned.

I will add some additional control flow, but again - please be as specific as possible with what you want here. It's very intricate control flow and I tried to modify it as little as possible while including these changes.

bbuzz31:
Lines 251 to 264 in raider.py are as follows

```        
interp_method = params['interpolate_time']
if interp_method in ['none', 'center_time']:
    times = get_nearest_wmtimes(t, [model.dtime() if \
                                model.dtime() is not None else 6][0]) if interp_method == 'center_time' else [t]
elif interp_method == 'azimuth_time_grid':
    n_target_dates = 3
    step = model.dtime()
    time_step_hours = step if step is not None else 6
    # Will yield 3 dates
    times = get_n_closest_datetimes(t,
                                    n_target_dates,
                                    time_step_hours)
else:
    raise NotImplementedError
```

`interp_method` just needs to be a string, currently its a boolean when it doesn't exist.
`.get('interp_method', 'none')` as you suggested should be fine

bbuzz31 (approved):

bbuzz31:
@cmarshak thanks for taking care of these last bits. feel free to merge if you are finished

PR #573, opened 2023-08-04
Avoid duplicate datetime keys

sssangha:
## Description
Fix bug where `Datetime` column is duplicated in file of GNSS delays, which leads to an error when attempting to subset the dataframe WRT to the CSV containing WM delays.

## Type of change


bbuzz31 (approved):

bbuzz31 (changes requested):
This produces same error as previously because the other datetime column is a string not datetime

jlmaurer (approved):
LGTM

bbuzz31 (approved):

bbuzz31:
@sssangha this should be merged, is there a reason you closed it without doing so?

Issue #584, opened 2023-08-25
[BUG] Lack of Availability of HRRR Weather Models Means 3-Date Azimuth Grid Interpolation Fails for GUNW workflow


cmarshak:
```
INFO:RAiDER:Extent of the input is (xmin, ymin, xmax, ymax): -78.05, 41.60, -74.40, 43.55
Traceback (most recent call last):
  File "/opt/conda/envs/RAiDER/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/conda/envs/RAiDER/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/conda/envs/RAiDER/lib/python3.10/site-packages/RAiDER/cli/__main__.py", line 46, in <module>
    main()
  File "/opt/conda/envs/RAiDER/lib/python3.10/site-packages/RAiDER/cli/__main__.py", line 42, in main
    process_entry_point.load()()
  File "/opt/conda/envs/RAiDER/lib/python3.10/site-packages/RAiDER/cli/raider.py", line 620, in calcDelaysGUNW
    cube_filenames = calcDelays([path_cfg])
  File "/opt/conda/envs/RAiDER/lib/python3.10/site-packages/RAiDER/cli/raider.py", line 388, in calcDelays
    raise NotImplementedError(f'The {interp_method} with {n} retrieved weather model files was not well posed '
NotImplementedError: The azimuth_time_grid with 1 retrieved weather model files was not well posed for the dela current workflow.
```

is raised from these inputs:

```
{'execution_started': True,
 'expiration_time': '2023-10-25T00:00:00+00:00',
 'job_id': '00312fde-8f89-4c5c-a0d6-ec2df678f316',
 'job_parameters': {'compute_solid_earth_tide': True,
                    'estimate_ionosphere_delay': True,
                    'frame_id': 16421,
                    'granules': ['S1A_IW_SLC__1SDV_20170210T225922_20170210T225949_015228_018F0A_787A',
                                 'S1A_IW_SLC__1SDV_20170210T225947_20170210T230014_015228_018F0A_10FF'],
                    'secondary_granules': ['S1A_IW_SLC__1SDV_20170129T225922_20170129T225949_015053_018996_BEDD',
                                           'S1A_IW_SLC__1SDV_20170129T225947_20170129T230014_015053_018996_6EA8'],
                    'weather_model': 'HRRR'},
[...]

PR #585, opened 2023-08-26
tighter control flow logic for calcDelays

jlmaurer:
- some products fail due to incorrect flow control in the main raider.py CLI 

## Description

- Instead of running the command, the CLI errors out at the weather model / date checking stage 
- Custom exceptions allow for more nuanced handling of errors
- Control flow is more readable by parsing out specific tasks to their own functions
- azimuth time interpolation only works if all three files can be found, center_time will work with at least one file

## Motivation and Context

- with three options now for time interpolation (or not), we need to have better-defined flow control for processing

## How Has This Been Tested?

- Raider-docs runs now
- some additional unit tests written and pass (more currently in the works)

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/cli/__main__.py
+++ b/tools/RAiDER/cli/__main__.py
@@ -29,7 +29,14 @@ def main():
     
     sys.argv = [args.process, *unknowns]    
 
-    from RAiDER.cli.raider import calcDelays, downloadGNSS, calcDelaysGUNW
+    if args.process == 'calcDelays':
+        from RAiDER.cli.raider import calcDelays
+    elif args.process == 'downloadGNSS':
+        from RAiDER.cli.raider import downloadGNSS
+    elif args.process == 'calcDelaysGUNW':
+        from RAiDER.cli.raider import calcDelaysGUNW
+    else:
+        raise NotImplementedError(f'Process {args.process} has not been fully implemented')
 
     try:
         # python >=3.10 interface
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -241,7 +241,8 @@ def calcDelays(iargs=None):
     ):
 
         ###########################################################
-        # weather model calculation
+        # Weather model calculation
+        ###########################################################
         logger.debug('Starting to run the weather model calculation')
         logger.debug(f'Requested date,time: {t.strftime("%Y%m%d, %H:%M")}')
         logger.debug('Beginning weather model pre-processing')
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -256,7 +256,7 @@ def calcDelays(iargs=None):
             logger.warning('interp_method is not specified, defaulting to \'none\', i.e. nearest datetime for delay '
                            'calculation')
 
-        if (interp_method is not 'azimuth_time_grid'):
+        if (interp_method != 'azimuth_time_grid'):
             times = get_nearest_wmtimes(
                 t, [model.dtime() if model.dtime() is not None else 6][0]
             ) if interp_method == 'center_time' else [t]
@@ -516,7 +516,7 @@ def calcDelaysGUNW(iargs: list[str] = None) -> xr.Dataset:
         file_name = iargs.file.split('/')[-1]
         gunw_id = file_name.replace('.nc', '')
         if not RAiDER.aria.prepFromGUNW.check_hrrr_dataset_availablity_for_s1_azimuth_time_interpolation(gunw_id):
-            raise ValueError('The required HRRR data for time-grid interpolation is not available')
+            raise NoWeatherModelData('The required HRRR data for time-grid interpolation is not available')
 
     if not iargs.file and iargs.bucket:
         # only use GUNW ID for checking if HRRR available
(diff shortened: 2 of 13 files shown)


tools/RAiDER/cli/__main__.py
```diff
@@ -29,7 +29,14 @@ def main():
     
     sys.argv = [args.process, *unknowns]    
 
-    from RAiDER.cli.raider import calcDelays, downloadGNSS, calcDelaysGUNW
+    if args.process == 'calcDelays':
```
jlmaurer, line 32: This control flow reduces the total import load

tools/RAiDER/cli/raider.py
```diff
@@ -313,6 +313,10 @@ def calcDelays(iargs=None):
         elif len(wfiles)==1 and len(times)==2:
             logger.warning('Time interpolation did not succeed, defaulting to nearest available date')
             weather_model_file = wfiles[0]
+        
+        elif (interp_method == 'center_time')  and len(times)==1:
```
jlmaurer, line 317: This fixes the interp_time option

tools/RAiDER/llreader.py
```diff
@@ -185,6 +185,8 @@ def set_output_xygrid(self, dst_crs=4326):
             out_proj = CRS.from_epsg(dst_crs.replace('EPSG:', ''))
         except pyproj.exceptions.CRSError:
             out_proj = dst_crs
+        except AttributeError:
```
jlmaurer, line 188: This is a small bug that was giving me grief.

cmarshak:
Please do not merge until we sort out control flow for GUNWs - this is more urgent.

jlmaurer:
> Please do not merge until we sort out control flow for GUNWs - this is more urgent. See #584

@cmarshak the error message mentioned in #584 is the same as what I was getting for the time-interp case. I suspect the issue is the same/similar, most likely having to do with the specific number of files returned and having an exact time. `get_nearest_wmtimes` is the relevant function.

cmarshak:
I agree - but unless this is tested on the exact GUNW that I mentioned in the error then this should not be merged...

cmarshak:
To be clear, the control flow in the main script requires concentration/thought. If we merge this and the GUNW logic still fails, then this affects the GUNW production.

We can in theory always re-run the RAiDER docs to fix this issue. This will fixed EOD tomorrow.

cmarshak:
I cannot necessarily recreate the error from the cloud here if this is merged first.

cmarshak:
I had a headache writing #593 in part because of the try/except handling [here](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/cli/raider.py#L273-L292) in `calcDelays` in `raider.py`.

I am concerned as an outsider that generally if `prepareWeatherModel` doesn't complete shouldn't that be an error as there are numerous reasons it couldn't whether it be a download issue (transient) or something intrinsic to RAiDER that someone hasn't considered in the weather model preparation? Or if not, is there are more explicit error (i.e. a RAiDER exception that could be made) that can be caught to ensure proper completion as you expect? The code would become a lot simpler too allowing the code to error out. Not to mention unintended consequences from my vantage point.

To elaborate on the latter point, if I am user of Raider and I specify as interpolation as `none` or `center_time`, then I can get different results for the same inputs if prepareWeatherModel has transient download issues. Is that desired? Why is it OK for weather model times to be ignored? Can you explain? This is particularly concerning when the model step times are 3 hours or 6 hours, in which you definitely do not just want to get the datetime that was available. Is there another error that you also are considering?

Additionally, having these try/except without failure leads to confusing errors like those in #587 (or at least confusing to me). The error doesn't indicate that input data was not available (which in the GUNW case, should have led to an error or exit much sooner).

Also, if the control flow is being added, I would encourage the use [pytest-mock](https://pypi.org/project/pytest-mock/) (or something similar) to make sure whatever logic you want (and errors) is happening as you expect without doing any actual delay calculations.

cmarshak (approved):

jlmaurer:
Hey @cmarshak, sorry about your headaches! This control flow logic isn't the cleanest, and I've updated this PR to try to at least make the code more readable. 

You have a good point about the failures. I usually think from a scientific user point of view - I want a weather model to use, and I'd like to use time-interpolation, but if it doesn't happen I really don't care, just get me a weather model. I've used many codes that post lots of errors because of various detailed problems when I just want a quick-and-dirty result I can work with. For the GUNW workflow that's obviously not what we want, but for general RAiDER use my inclination is to give the user something to work with and raise warnings about what they got. 

For this PR, I've made it explicit for azimuth time interpolation that three files are required or the program will fail. `center_time` will still try to work under certain cases with a warning. I've defined several custom exceptions to be more explicit about types of failures, and RAiDER will only continue without failure in appropriate cases. Testing all this will take a bit of time though, so hopefully there is enough time to get it done right.

cmarshak:
Hey Jeremy - do whatever works for the multiple uses of RAiDER - no need for my opinions to block that. I hope that general thoughts can provide helpful pointers to a slightly different viewpoint. Any large code bases will have headaches. 

I think the ISCE2 way of thinking is to put everything into a log and carefully look at the logs after if there are failures. I personally think it's easier from a debugging point of view to have errors that are causing any time of failures to be captured at the end of standard i/o.

Thanks for merging that monolith of a PR - I will test these new updates this morning and post on their respective issue tickets if they are success.

Also, maybe something that wasn't clear that *definitely* should be flagged is not the azimuth time interpolation uses 2 or 3 times. 2 times are used when the acquisition time occurs between two model times with a 5 minute buffer. Otherwise 3 times are used. That means simply looking at the number of times retrieved alone could lead to more unexpected behaviors.

Finally, paraphrasing from [`import this`](https://stackoverflow.com/q/43577404):

```
Errors should never pass silently.
Unless explicitly silenced.
In the face of ambiguity, refuse the temptation to guess.
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
Now is better than never.
Although never is often better than *right* now.
```

PR #588, opened 2023-08-30
v0.4.4

cmarshak:
Fixes the error related to convergence of geo2rdr when multiple orbit files are needed for azimuth time interpolation of different weather models dates.


cmarshak:
Please make sure to merge only after all tests pass - I am going to re-run CircleCI.

The `x` is due to coveralls which I believe is not strictly required.

PR #591, opened 2023-09-06
patch for model timing=requested time

bbuzz31:
When the center time is specified, but the requested time is at the same time as a model time, raider control logic fails. 
Added more logic that enables it to proceed

### DIFF
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -256,8 +256,11 @@ def calcDelays(iargs=None):
         # The two datetimes will be combined to a single file and processed
         # TODO: make more transparent control flow for GUNW and non-GUNW workflow
         if (interp_method in ['none', 'center_time']):
-            times = get_nearest_wmtimes(t, [model.dtime() if \
-                                        model.dtime() is not None else 6][0]) if interp_method == 'center_time' else [t]
+            if interp_method == 'center_time':
+                times = get_nearest_wmtimes(t, [model.dtime() if \
+                                        model.dtime() is not None else 6][0])
+            else:
+                times = [t]
         elif interp_method == 'azimuth_time_grid':
             n_target_dates = 3
             step = model.dtime()
@@ -311,7 +314,7 @@ def calcDelays(iargs=None):
             else:
                 ## check if requested time is coincident with model time
                 valid_hours = np.arange(0, 24, model._time_res)
-                if model._time.hour in valid_hours:
+                if tt.hour in valid_hours:
                     weather_model_file = wfiles[0]
                 else:
                     n = len(wfiles)
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -305,8 +305,18 @@ def calcDelays(iargs=None):
 
         # nearest weather model time via 'none' is specified
         # When interp_method is 'none' only 1 weather model file and one relevant time
-        elif (interp_method == 'none') and (len(wfiles) == 1) and (len(times) == 1):
-            weather_model_file = wfiles[0]
+        elif len(wfiles) == 1 and len(times) == 1:
+            if interp_method == 'none':
+                weather_model_file = wfiles[0]
+            else:
+                ## check if requested time is coincident with model time
+                valid_hours = np.arange(0, 24, model._time_res)
+                if model._time.hour in valid_hours:
+                    weather_model_file = wfiles[0]
+                else:
+                    n = len(wfiles)
+                    raise NotImplementedError(f'The {interp_method} with {n} retrieved weather model files was not well posed '
+                                      'for the dela current workflow.')
 
         # only one time in temporal interpolation worked
         # TODO: this seems problematic - unexpected behavior possibly for 'center_time'
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -78,7 +78,12 @@ def tropo_delay(
         else:
             height_levels = wm_levels
 
-    zref = float(zref)
+    if isinstance(zref, str):
+        try:
+            zref = float(zref)
+        except:
+            log.warning('Cannot convert zref={%s} to a number', zref)
+
     if not zref:
         zref = toa
 
--- a/tools/RAiDER/delay.py
+++ b/tools/RAiDER/delay.py
@@ -78,6 +78,7 @@ def tropo_delay(
         else:
             height_levels = wm_levels
 
+    zref = float(zref)
     if not zref:
         zref = toa
 
--- a/tools/RAiDER/processWM.py
+++ b/tools/RAiDER/processWM.py
@@ -83,11 +83,6 @@ def prepareWeatherModel(
     f = weather_model.load()
 
     if f is not None:
-        logger.warning(
-            'The processed weather model file already exists,'
-            ' so I will use that.'
-        )
-
         containment = weather_model.checkContainment(ll_bounds)
         if not containment and weather_model.Model() in 'GMAO ERA5 ERA5T HRES'.split():
             msg = 'The weather model passed does not cover all of the input ' \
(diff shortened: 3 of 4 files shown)


tools/RAiDER/cli/raider.py
```diff
@@ -305,8 +305,18 @@ def calcDelays(iargs=None):
 
         # nearest weather model time via 'none' is specified
         # When interp_method is 'none' only 1 weather model file and one relevant time
-        elif (interp_method == 'none') and (len(wfiles) == 1) and (len(times) == 1):
-            weather_model_file = wfiles[0]
+        elif len(wfiles) == 1 and len(times) == 1:
```
cmarshak, line 308: Why do you remove logic related to `interp_method`? If this is specifically for `center_time` please include this otherwise there is a risk this could effect other workflows.

tools/RAiDER/cli/raider.py
```diff
@@ -305,8 +305,18 @@ def calcDelays(iargs=None):
...
+            if interp_method == 'none':
+                weather_model_file = wfiles[0]
+            else:
```
cmarshak, line 311: See comment above.

tools/RAiDER/cli/raider.py
```diff
@@ -305,8 +305,18 @@ def calcDelays(iargs=None):
...
+                ## check if requested time is coincident with model time
+                valid_hours = np.arange(0, 24, model._time_res)
+                if model._time.hour in valid_hours:
+                    weather_model_file = wfiles[0]
+                else:
+                    n = len(wfiles)
+                    raise NotImplementedError(f'The {interp_method} with {n} retrieved weather model files was not well posed '
+                                      'for the dela current workflow.')
```
cmarshak, line 319: Now confused more - because shouldn't this be strictly for center time?

tools/RAiDER/delay.py
```diff
@@ -78,6 +78,7 @@ def tropo_delay(
         else:
             height_levels = wm_levels
 
+    zref = float(zref)
```
cmarshak, line 81: This seems like this fix unrelated to PR that you are putting here. Please explain.

tools/RAiDER/processWM.py
```diff
@@ -83,11 +83,6 @@ def prepareWeatherModel(
     f = weather_model.load()
 
     if f is not None:
-        logger.warning(
-            'The processed weather model file already exists,'
-            ' so I will use that.'
-        )
-
```
cmarshak, line 90: Again, why are you removing this? Should it be included in this PR.

cmarshak (changes requested):
It sounds like we need to make the control flow more complex. I really encourage us to zero-in on that for this or other PRs. Also, there needs to be tests with mocks to make sure logic is correctly being implemented. Be explicit as possible given how many cases are handled.

tools/RAiDER/delay.py
bbuzz31: a string can be entered instead of a float/int by user. catching here. given our timelines, it's easier to just include here

tools/RAiDER/processWM.py
bbuzz31: it's misleading because if the processed weather model already exists, the warning is raised on line 58.

bbuzz31:
> We need to make the control flow more complex. I really encourage us to zero-in on that for PR. Also, there needs to be tests with mocks to make sure logic is correctly being implemented. Be explicit as possible given how many cases are handled.

Agreed, if you could help with some tests it would be great

cmarshak:
I think I understand better what you are trying to accomplish. 

I really ask when you ask me to review to try to keep PRs focused on a particularly set of problems and communicate all these problems. Please don't make changes that you think are needed, unrelated to stated problem being solved. It's a confusing to me (reviewer) and a good way to introduce new bugs. Agree with all your changes, but it requires more time to parse what you are doing. Also it requires more time to figure it out.

I think tests are a good way to make sure what you are doing works and communicate the set of problems being fixed. I tried to provide examples previously using `pytest-mock` in the `test_GUNW.py` so that one could run `raider.py` without running any of the pre-processing or integration. This is what ASF does regularly. Happy to answer questions on this front. 

I am particularly picky (maybe overly so) because I was confused when trying read the control flow previously. I also am generally concerned of the scenario when 1 weather model file is downloaded and the process proceeds. I believe what you are proposing is a good idea and generally does not produce unintended side-effects.

I am still concerned as to why `t` is not being used to determine how many valid times are used.

tools/RAiDER/cli/raider.py
```diff
@@ -305,8 +305,18 @@ def calcDelays(iargs=None):
...
         # When interp_method is 'none' only 1 weather model file and one relevant time
-        elif (interp_method == 'none') and (len(wfiles) == 1) and (len(times) == 1):
-            weather_model_file = wfiles[0]
+        elif len(wfiles) == 1 and len(times) == 1:
+            if interp_method == 'none':
+                weather_model_file = wfiles[0]
+            else:
+                ## check if requested time is coincident with model time
+                valid_hours = np.arange(0, 24, model._time_res)
+                if model._time.hour in valid_hours:
```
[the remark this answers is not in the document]
cmarshak: And is there an hour floor happening with this logic? Why?

tools/RAiDER/delay.py
```diff
@@ -78,6 +78,12 @@ def tropo_delay(
         else:
             height_levels = wm_levels
 
+    if isinstance(zref, str):
+        try:
+            zref = float(zref)
+        except:
+            log.warning('Cannot convert zref={%s} to a number', zref)
+
```
cmarshak, line 86: Don't think you need try/except. Python builtin error will be explicit enough for users.

cmarshak, line 86:
```suggestion
    if isinstance(zref, str):
        zref = float(zref)
```

Think built in python error is better than `try`/`except`.

tools/RAiDER/cli/raider.py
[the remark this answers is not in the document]
bbuzz31: yeah model._time is equivalent to tt. this may be a redundant check

tools/RAiDER/delay.py
bbuzz31: this shouldn't be an error; often time it will be None[bool] or 'null'

bbuzz31:
> I think I understand better what you are trying to accomplish.
> 
> I really ask when you ask me to review to try to keep PRs focused on a particularly set of problems and communicate all these problems. Please don't make changes that you think are needed, unrelated to stated problem being solved. It's a confusing to me (reviewer) and a good way to introduce new bugs. Agree with all your changes, but it requires more time to parse what you are doing.
> 
> I think tests are a good way to make sure what you are doing works and communicate the set of problems being fixed. I tried to provide examples previously using `pytest-mock` in the `test_GUNW.py` so that one could run `raider.py` without running any of the pre-processing or integration. This is what ASF does regularly. Happy to answer questions on this front.
> 
> I am particularly picky (maybe overly so) because I was confused when trying read the control flow previously. I also am generally concerned of the scenario when 1 weather model file is downloaded and the process proceeds. I believe what you are proposing is a good idea and generally does not produce unintended side-effects.
> 
> I am still concerned as to why `t` is not being used to determine how many valid times are used. It might be because I don't understand the weather model class...
> 
> Jeremy should be the final say on this anyways. Please ignore anything that is getting in the way.

I agree under normal circumstances, but given our deadlines and bandwidth this will have to suffice

cmarshak:
Ok - whatever works for Jeremy, David, and yourself. I am not in line with everything RAiDER is being used for - mainly just getting GUNWs generated.

I am being very careful because we are working on related catches for HRRR availability and don't want unexpected side-effects down stream. 

Another solution could be if this fix is for independent anlaysis, maybe utilize your own fork and merge in the latest RAiDER? And make PRs for changes that affect the relevant deliverable workflows? But do whatever you need to do.

My opinion is very, very low in order of importance and much too verbose.

jlmaurer:
@bbuzz31 I apologize didn't realize you were working on this. I think my PR #585 should address this issue. I've already added some tests as well. I added your option if interp_time=='none'.

jlmaurer:
Closing this per conversation with @bbuzz31 as replaced by #585.

PR #601, opened 2023-10-03
Code maintenance

jlmaurer:
## Description

- Updates unit tests
- Removes unused functions
- Adds docstrings
- Update documentation as needed

## Motivation and Context

- Increase code coverage to 60%
- Add docstrings for important functions
- Clean up old code

## How Has This Been Tested?

- tests pass locally

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have added an explanation of what your changes do and why you'd like us to include them.
- [x] I have written new tests for your core changes, as applicable.
- [x] I have successfully ran tests with your changes locally.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #602, opened 2023-10-10
make weather file directory when it doesnt exist

bbuzz31:
tiny fix to make the weather model directory; if user calls processWM directly it may not have already been created


jlmaurer (approved):
LGTM

jlmaurer:
@bbuzz31 the same two tests are failing for you that are failing for me in #601. Any ideas what's going on?

jlmaurer:
> @bbuzz31 the same two tests are failing for you that are failing for me in #601. Any ideas what's going on?

I think you can still merge this though; clearly this is not the reason those tests are failing.

PR #604, opened 2023-10-18
Insure data/alaska.geojson.zip is included in package build

jhkennedy:
This fixes the build issue seen in conda-forge/raider-feedstock#11 and conda-forge/raider-feedstock#12.

In both of those PRs, the `tools/RAiDER/models/data/alaska.geojson.zip` file was not being included in the built package because we're building from a src tarball, which does not have the `.git` data inside it that is used by `setuptools_scm` to ensure all checked-in files are in the distribution (likewise, this is not an issue seen when building from a clone).

You can reproduce the missing zip file like:
```shell
wget https://github.com/dbekaert/RAiDER/archive/v0.4.3.tar.gz 
tar -zxvf v0.4.3.tar.gz
export SETUPTOOLS_SCM_PRETEND_VERSION=0.4.3
python -m pip install ./RAiDER-0.4.3
```
And look at the installed package.

Then, if you make the changes here, you'll see the zip file show up in the installed package.


jhkennedy:
CircleCI appear to be a Herbie problem and not related to this change, so I'm content to punt that to a follow on issue/pr.

PR #606, opened 2023-10-19
make isce3 an optional import for `s1_azimuth_timing.py`

jhkennedy:
Cleans up the ICSE3 imports across RAiDER and adds some information on handling optional imports in the contributing guide. 

Notably, following this pattern, you can still import at the top of the file and use typehints. Notably, if you forget the `if isce is None` block, you'll just get an `AtributeError` like:
```python
>>> isce = None
>>> isce.any()
AttributeError: 'NoneType' object has no attribute 'any'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixes
 * make weather file directory when it doesn't exist
 * Ensures the `models/data/alaska.geojson.zip` file is packaged when building from the source tarball
+* Make ISCE3 an optional dependency in `s1_azimuth_timing.py`
 
 ## [0.4.5]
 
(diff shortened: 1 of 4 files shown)


jhkennedy:
I'm tempted to clean up some of the other optional imports across RAiDER to match this pattern, but can't decide if I should do it here, or in a follow on PR. 

Also, there are some python standard library imports that are guarded like they are optional imports... is there a reason for this? For example:
https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/utilFcns.py#L16-L19

jhkennedy:
Cicle-CI failure is due to a redirect error (possibly some of the test data has moved) which is not related to this PR,

jlmaurer (approved):
LGTM

jlmaurer:
> I'm tempted to clean up some of the other optional imports across RAiDER to match this pattern, but can't decide if I should do it here, or in a follow on PR.

Doesn't matter to me, this looks fine to me so we can merge it now

> Also, there are some python standard library imports that are guarded like they are optional imports... is there a reason for this? For example: https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/utilFcns.py#L16-L19

@jhkennedy Not sure there is a reason, although I feel like I remember this being an issue at some point in the past I can't see how importing would result in an error, so it can probably be removed.

jhkennedy:
Sounds good, let's go ahead and merge this now, and I'll circle back later for the other imports.

PR #609, opened 2023-11-06
Update unit tests

### LINKED ISSUE #607: Need to upgrade importlib.metadata.entry_points for Python 3.12
Right now, RAiDER is incompatible with Python 3.12 due to how we handle entrypoints here:
https://github.com/dbekaert/RAiDER/blob/v0.4.5/tools/RAiDER/cli/__main__.py#L34-L40

Resulting in an error that looks like:
```python
  File .../bin/raider.py", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File ".../python3.12/site-packages/RAiDER/cli/__main__.py", line 36, in main
    process_entry_point = entry_points(group='console_scripts', name=f'{args.process}.py')[0]
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File ".../python3.12/importlib/metadata/__init__.py", line 287, in __getitem__
    raise KeyError(name)
KeyError: 0
```

We'll need to adjust how we select the process entrypoint to support python 3.12. It's worth reading the compatibility note for `importlib_metadata.entry_points` for both python 3.11 and 3.12:
* 3.11: https://docs.python.org/3.11/library/importlib.metadata.html#entry-points
* 3.12: https://docs.python.org/3.12/library/importlib.metadata.html#entry-points

### LINKED ISSUE #610: [BUG] Sentinel-1 ESA Orbit Availability and Rate Limiting
I am writing this issue ticket totally *second hand*. (based on coversation with @jhkennedy and @asjohnston-asf). Moreover, this is an issue for *cloud computing* when we have numerous parallel requests to get orbits so is not a bug for processing run locally.

Here is the issue:

1. We are utilizing https://github.com/scottstanie/sentineleof which utilizes the new ESA interface described here: https://github.com/scottstanie/sentineleof#update-2023-10-31-changes-to-sentinel-1-orbit-files-source. Here is the line we are using this: https://github.com/dbekaert/RAiDER/blob/1f2065501ef516d7a644c6ca0a08a1d9877a77ad/tools/RAiDER/aria/prepFromGUNW.py#L281
2. From my understanding the new ESA interface is severly *rate limited*: 4 requests / minute (for a given user).

Here are the following questions that need to be answered before the next processing campaign:

1. Will https://github.com/scottstanie/sentineleof error out if there are too many parallel requests to ESA even if it checks both ESA and ASF? @scottstanie will know.
2. Should we just use `force-asf` option if 1 is true? https://github.com/scottstanie/sentineleof/blob/master/eof/download.py#L48
3. Are there any downsides to using `force-asf`?
4. Or we we could use https://github.com/ASFHyP3/hyp3-lib/blob/develop/hyp3lib/get_orb.py#L146 but this is a much heavier lift not within the code itself but due to mocking the tests.

This is not urgent, but will need to be resolved before the next processing campaign.

jhkennedy:
This PR:
* updates CircleCI config to test across multiple Python versions (3.9-3.12)
  * Note: only runs `coveralls` on Python 3.12 
* removes unnecessary imports in `cli/__main__.py`
* Ensures all type hints are compatible with Python 3.9
* updates loading entrypoints for Python 3.12, fixes #607
* re-orders `environment.yml` so that packages are in the correct section and listed alphabetically
* :exclamation: marks all tests in `test_temporal_interpolate.py` as "long" to prevent them from running in CI due to #611
* Updates `sentineleof` to version 9 or greater, fixes #610 
  * this means orbits will be downloaded from **ESA CDSE, which will require credentials**
  * `sentineleof` also now returns paths instead of strings, so `_get_acq_time_from_gunw_id` and tests were updated accordingly
* Adds an `_ensure_orbit_credentials` function in `prepFromGUNW.py` to ensure ESA CDSE credentials are in `~/.netrc` or provided via environment variables

  Note: For this one, I wanted to put this in `models/credentials.py` so all credentials were handled in the same place, but everything there is related to acquiring weather models so it didn't make sense to stuff it in there without a significant refactor.
  
* drops `hyp3lib` in favor of `sentineleof` as `get_orb` was the only thing being used from `hyp3lib` 
* Adds a `setup_from_env` function to `models/credentials.py` which will pull *all* credentials needed for acquiring weather model data from environment variables and ensure the correct config file is written. This makes setting up credentials in CI pipelines significantly easier. 
  
  Note: This *could* also have a command line interface, but I couldn't decide if it was worth it.


tools/RAiDER/models/credentials.py
```diff
@@ -69,7 +69,7 @@ def _check_envs(model):
         key = os.getenv('RAIDER_ECMWF_ERA5_API_KEY')
         host = API_URLS['cdsapirc']
 
-    elif model in ('HRES'):
+    elif model in ('HRES', 'ERAI'):
```
dbekaert, line 72: @jhkennedy dont think these should be sharing the same?

tools/RAiDER/cli/__main__.py
```diff
@@ -29,18 +29,9 @@ def main():
...
 
-    if args.process == 'calcDelays':
-        from RAiDER.cli.raider import calcDelays
-    elif args.process == 'downloadGNSS':
-        from RAiDER.cli.raider import downloadGNSS
-    elif args.process == 'calcDelaysGUNW':
-        from RAiDER.cli.raider import calcDelaysGUNW
-    else:
-        raise NotImplementedError(f'Process {args.process} has not been fully implemented')
-
```
jhkennedy, line 40: All these imports are done dynamically in the try block below, so these are unessary.

Likewise, argparse already ensures `args.process` is one of the implemented processes.

tools/RAiDER/models/credentials.py
jhkennedy: @dbekaert I wasn't quite sure what to do with this one. 

For my [`setup_from_env`](https://github.com/dbekaert/RAiDER/blob/70a663d3a95c046cdbb31d7402b40306f22d0e9e/tools/RAiDER/models/credentials.py#L150-L152), I want to loop through [all supported models](https://github.com/dbekaert/RAiDER/blob/70a663d3a95c046cdbb31d7402b40306f22d0e9e/tools/RAiDER/models/credentials.py#L18-L24) and ensure their credentials are setup with the existing [`_check_envs`](https://github.com/dbekaert/RAiDER/blob/70a663d3a95c046cdbb31d7402b40306f22d0e9e/tools/RAiDER/models/credentials.py#L66) function.

`ERAI` is in the `API_FILENAME` dictionary and specifies that it uses the `ecmwfapirc` file (the same as `HRES`) but wasn't handled in `_check_envs` at all, and when passed to it, would just return `None` causing `check_api` to throw. 

Notably, `check_api` only ensures the config file exists, not that the correct credentials are in the config file.

So this change ensures all the config files are written and sufficient to pass testing. 

Really, if the crednetials for `ERAI` are different than `HRES`, the correct way to handle this would be to either:
1. drop support for `ERAI`, or
2. add an `ERAI` block to `_check_envs` to look for `RAIDER_ERAI_*` environment variables *and* update `check_api` to ensure the right credentials are in `ecmwfapirc`. This would mean appending credentials for `HRES`/`ERAI` if the other is present, changing the `update_flag` option to actually *update* (check + append) not *replace*, and probably a few other things.

jhkennedy:
@dbekaert the checks will pass once the ESA_CDSE_USERNAME and ESA_CDSE_PASSWORD environment variables are added to CircleCI so that orbits can be downloaded. 

I also decided to *not* use GitHub Actions as CircleCI handles forks better. You can remove the repository secrets I had you add earlier.

environment.yml
```diff
@@ -34,36 +31,39 @@ dependencies:
...
  - pydap>3.2.2
  - pyproj>=2.2.0
  - pyyaml
  - rasterio>=1.3.0
- - rioxarray
  - requests
+ - rioxarray
  - s3fs
  - scipy>1.10.0
+ - sentineleof>=0.9.0
```
cmarshak, line 42:
```suggestion
 - sentineleof>=0.9.5
```
So that it resolves issue 6.10

cmarshak (commented):
I think once that's integrated it resolves the other issue, too.

Feeding all of this orbit junk at once.

CHANGELOG.md
```diff
@@ -9,7 +9,17 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+* Adds an `ensure_orbit_credentials` function in `prepFromGUNW.py` to ensure ESA CDSE credentials are in `~/.netrc` or provided via environment variables
+* Adds a `setup_from_env` function to `models/credentials.py` which will pull *all* credentials needed for acquiring weather model data from environment variables and ensure the correct config file is written. This makes setting up credentials in CI pipelines significantly easier
+
+### Changed
+* `sentineleof` upgraded to version 9 or later to fetch orbits from ESA CDSE
```
cmarshak, line 17:
```suggestion
* `sentineleof` upgraded to version 0.9.5 or later to fetch orbits from ESA CDSE and ensure that if CDSE fails the fetch resorts to ASF
```

jlmaurer (approved):
LGTM

tools/RAiDER/models/credentials.py
jhkennedy: Since `ERAI` is not in the [allowed model's list](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/models/allowed.py), and it sounds like `ERAI` has been moved from the ECMF API, I've dropped it from the credential checking module.

It could be added back in in a later PR that brings back `ERAI` support.

dbekaert:
> @dbekaert the checks will pass once the ESA_USERNAME and ESA_PASSWORD environment variables are added to CircleCI so that orbits can be downloaded.


Done

PR #615, opened 2023-11-20
v0.4.6

jlmaurer:
See CHANGELOG


jhkennedy (approved):
I switched the label from minor -> patch; as the changelog indicates we're bumping the patch part of the version number. 

> Given a version number MAJOR.MINOR.PATCH,

https://semver.org/

PR #629, opened 2024-02-08
Fix merra2

royagrace:
Replaced writeWeather2netCDF function with new function titled writeWeatherVarsXarray in utilFncs. 
## Description

Updated merra2, weatherModel, and gmao to call the correct function that was changed in utilFncs. 
## Motivation and Context

Tried to fix the read error and update the file to the latest xarray function. 
## How Has This Been Tested?

Tested using the fetch function in merra2, but kept receiving errors. There was a few errors that showed up no matter what I tried and upon further consideration it is possible it is some type of authentication issue. I kept getting 'errno -90 file not found', also 'syntax error expected WORD_WORD, access denied'. The URLs would work in browser but they would not download or open when running the test. 

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/models/gmao.py
+++ b/tools/RAiDER/models/gmao.py
@@ -9,7 +9,7 @@ from pyproj import CRS
 
 from RAiDER.models.weatherModel import WeatherModel, TIME_RES
 from RAiDER.logger import logger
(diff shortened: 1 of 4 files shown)


jlmaurer:
#630 replaces this.

PR #630, opened 2024-02-14
Fix MERRA-2 access

jlmaurer:
This PR updates the calls to the MERRA-2 API, which seem to keep changing...

## Description

- updated API calls to the GES DISC
- updated the rest of `merra2.py` module to reflect the changes in variable types and names
- replaced some outdated NETCDF4 writing and reading routines with calls to xarray

## Motivation and Context

MERRA-2 was not working due to recent changes to the API, which involved both a change in the dataset short name and a change in the way variables are accessed. This fix is needed as GMAO is also currently not working, and is used for several unit tests. This PR also replaces GMAO with MERRA-2 for those tests. 

## How Has This Been Tested?

- all unit tests pass locally

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Latest Changes
+### Fixed
+* [620](https://github.com/dbekaert/RAiDER/issues/620) - Fix MERRA-2 access because API was updated
+
 
 ## [0.4.7]
 ### Fixed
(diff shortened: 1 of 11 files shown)


jlmaurer:
@cmarshak I'm getting two failures saying that there is no "foo.json" files in `test_weather_model_availability_integration_using_valid_range`. Any idea what's up with that?

cmarshak:
Yes, `foo.json` does not exist. It's a mocked side-effect of all the calls made in the GUNW workflow.

This test makes sure that the attribute `model._valid_range` is checked and silently stops for the GUNW workflow. This for the situation of HRRR being called for present SLCs that do not have data available (usually we have a long time-series of GUNWs for which a small set of HRRR is not available).

It appears you swapped out `GMAO` with `MERRA2` for this test.

They have different date ranges:

GMAO; https://github.com/dbekaert/RAiDER/blob/8f41347cddd484cd59cdc8c4b365ba813a5e8aab/tools/RAiDER/models/merra2.py#L48
MERRA2: https://github.com/dbekaert/RAiDER/blob/8f41347cddd484cd59cdc8c4b365ba813a5e8aab/tools/RAiDER/models/gmao.py#L32

I am not sure what the differences between these two models are and why you changed the models out in the test. Honestly, you could remove `GMAO` from the test because I don't think we will use GMAO or MERRA2 for cloud processing. If you or David see this model being used, would be good to update the test accordingly (just pick dates outside the date range of MERRA2).

Hope that helps.

bbuzz31 (approved):

PR #631, opened 2024-02-20
Pulling refs/tags/v0.4.7 into dev

dbekaert:
Fast-forward of main to dev failed!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,13 +12,13 @@ on:
 
 jobs:
   call-version-info-workflow:
-    uses: ASFHyP3/actions/.github/workflows/reusable-version-info.yml@v0.8.3
+    uses: ASFHyP3/actions/.github/workflows/reusable-version-info.yml@v0.11.0
     with:
       python_version: '3.10'
 
   call-docker-ghcr-workflow:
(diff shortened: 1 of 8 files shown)


jlmaurer:
@dbekaert do you know if this is still needed?

jhkennedy:
@jlmaurer yes -- we want all the tags, which are applied to the `main` branch, to be direct descendants of the dev branch as well so that the version numbers, which are based on git tags, are picked up correctly.

jhkennedy:
@jlmaurer, since this is old, let me make sure the version numbers will be correct once merged. I'll merge it if so.

jhkennedy (approved):
Okay, so, right now, from the `dev` branch, if you check the version number you'll get:
```python
>>> python -m setuptools_scm
0.4.7.dev40+g3e04129
```
which is incorrect since v0.4.7 has already been released. Had the fast-forward merge worked, we would be getting:
`0.4.8.dev25+g9c10639`

If I merge this PR, then we'll get:
`0.4.8.dev30+gb14d1b9`

Which accounts for the stuff pushed to `dev` in between the release and the merge-back attempt. 

So, it looks like we're good to merge. :shipit:

PR #632, opened 2024-02-20
Hrrr aztime

bbuzz31:
bugfix to correctly set HRRR-AK name for azimuth grid interpolation (was missing the hyphen)


jlmaurer (approved):
LGTM

Issue #634, opened 2024-04-02
Orbit issues `RuntimeError: Access token creation failed. Reason: 401 Client Error: Unauthorized for url: https://identity.dataspace.copernicus.eu/auth/realms/CDSE/protocol/openid-connect/token`


cmarshak:
**Describe the bug**

It appears a new exception is being raised in the orbit download of raider. I am not sure if this is transient issue or something that is a error with `EOF` in how exceptions are being raised. In <1 hour, I saw several thousand jobs fail so am disabling processing.

```
 â•­â”€[48;2;255;255;255m[38;2;136;33;27mâ–Œ[0m[38;2;12;53;118m[48;2;240;234;210mâ–Œ[38;2;0;0;0m[1mHerbie[0mâ”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â•®
 â”‚ INFO: Created a default config file.                 â”‚
 â”‚ You may view/edit Herbie's configuration here:       â”‚
 â”‚ [38;2;255;153;0m     /home/raider/.config/herbie/config.toml      [0m   â”‚
 â•°â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â•¯
Downloading s3://hyp3-a19-jpl-contentbucket-1wfnatpznlg8b/f8e84e33-4fe3-4aca-8d2d-b9cb302be80c/S1-GUNW-D-R-115-tops-20240223_20230204-141542-00124W_00040N-PP-48c1-v3_0_1.nc to S1-GUNW-D-R-115-tops-20240223_20230204-141542-00124W_00040N-PP-48c1-v3_0_1.nc
/opt/conda/envs/RAiDER/lib/python3.12/site-packages/RAiDER/s1_azimuth_timing.py:242: FutureWarning: 'H' is deprecated and will be removed in a future version, please use 'h' instead.
[...]

PR #635, opened 2024-04-10
dl orbits from asf first

bbuzz31:
download orbits from asf first then fall back to esa

### DIFF
--- a/tools/RAiDER/s1_orbits.py
+++ b/tools/RAiDER/s1_orbits.py
@@ -58,6 +58,8 @@ def get_orbits_from_slc_ids(slc_ids: List[str], directory=Path.cwd()) -> List[Pa
     missions = [slc_id[0:3] for slc_id in slc_ids]
     start_times = [re.split(r'_+', slc_id)[4] for slc_id in slc_ids]
     stop_times = [re.split(r'_+', slc_id)[5] for slc_id in slc_ids]
+    # logger.warning('st+stop %s', start_times+stop_times)
+    # logger.warning('missions %s', missions)
 
     orb_files = download_eofs(start_times + stop_times, missions * 2, str(directory))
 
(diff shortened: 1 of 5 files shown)


tools/RAiDER/s1_orbits.py
```diff
@@ -58,7 +58,19 @@ def get_orbits_from_slc_ids(slc_ids: List[str], directory=Path.cwd()) -> List[Pa
...
+def download_eofs(dts:list, missions:list, save_dir:str):
+    """ Wrapper to first try downloading from ASF """
+    import eof.download
+    try:
+        orb_files = eof.download.download_eofs(dts, missions, save_dir=save_dir, force_asf=True)
+    except:
+        logger.warning('Could not download orbit from ASF, trying ESA...')
+        orb_files = eof.download.download_eofs(dts, missions, save_dir=save_dir, force_asf=False)
+
+    return orb_files
```
dbekaert, line 76: @bbuzz31 is this a list of dates or single date. We might want to try and pull whatever we can pull from ASF and then what fails from ESA?

bbuzz31:
@dbekaert we're now pulling each orbit individually, first checking if it's ASF then falling back to ESA. It's a bit ugly because we can't leverage the simultaneous jobs as `sentineleof` has built in, but ESA limits to 4 anyway. I tried checking against existing orbit files so that we could keep this functionality, but it is too complicated to figure out the different combinations that we currently support

bbuzz31:
@cmarshak I had to change the orbit reader test to support the new multiple calls. it works but I have no real idea what I'm doing with this mocker stuff. let me know if it should be changed

dbekaert:
Once we have the fix on the sentinel reader for the coockie we can just
revert back to our original approach which would be indeed cleanest.

jlmaurer (commented):
LGTM but when I pulled and tried to run locally I'm getting `ValueError: get_sv: I cannot parse the statevector file ["[PosixPath('/Users/jlmd9g/Documents/raider_test/aria_products/orbits/S1A_OPER_AUX_POEORB_OPOD_20220213T081624_V20220123T225942_20220125T005942.EOF')]"]`. Any ideas?

cmarshak:
> @cmarshak I had to change the orbit reader test to support the new multiple calls. it works but I have no real idea what I'm doing with this mocker stuff. let me know if it should be changed

The idea is to slightly increase the runtime of tests - have the data staged and mock/inject into the workflow. Also can make sure you are calling certain functions and not calling others e.g. if you want to make sure you never enter a particular part of a workflow because say HRRR is not available, you can do that. You can let me know if you have specific questions.

cmarshak:
> @cmarshak I had to change the orbit reader test to support the new multiple calls. it works but I have no real idea what I'm doing with this mocker stuff. let me know if it should be changed

The idea is to slightly increase the runtime of tests - have the data staged and mock/inject into the workflow. Also can make sure you are calling certain functions and not calling others e.g. if you want to make sure you never enter a particular part of a workflow because say HRRR is not available, you can do that. You can let me know if you have specific questions.

cmarshak:
Resolves #634.

jlmaurer (approved):
Looks to be passing now!

PR #636, opened 2024-04-11
Allow calcDelaysGUNW to process HyP3 jobs independently

jhkennedy:
Right now, the calcDelaysGUNW workflow assumes that it's the second step in a multi-step INSAR_ISCE HyP3 job, so it only looks for GUNW files under the same prefix it will upload final products to.

However, if we want a RAiDER-only job that processes previous INSAR_ISCE jobs (that either failed on the RAiDER step or skipped the RAiDER step), we'll need to download products from the previous job's prefix and upload the final products to the current RAIDER's job prefix. 

This adds the option to take a different input prefix than the final output prefix. 

Supports ASFHyP3/hyp3#2192

--- 

Note: HyP3 jobs are all assigned a unique UUID as the Job ID, and we organize the HyP3 content bucket under a prefix, which *is* the job ID, so Job ID and prefix can be used interchangeably.

### DIFF
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -510,7 +510,7 @@ def calcDelaysGUNW(iargs: list[str] = None) -> xr.Dataset:
     iargs = p.parse_args(iargs)
 
     if not iargs.input_bucket_prefix:
-        iargs.input_bucket_prefix = iargs.input_bucket_prefix
+        iargs.input_bucket_prefix = iargs.bucket_prefix
 
     if iargs.interpolate_time not in ['none', 'center_time', 'azimuth_time_grid']:
         raise ValueError('interpolate_time arg must be in [\'none\', \'center_time\', \'azimuth_time_grid\']')
@@ -547,7 +547,7 @@ def calcDelaysGUNW(iargs: list[str] = None) -> xr.Dataset:
             #       we include this within this portion of the control flow.
             print('Nothing to do because outside of weather model range')
             return
-        json_file_path = aws.get_s3_file(iargs.bucket, iargs.bucket_prefix, '.json')
+        json_file_path = aws.get_s3_file(iargs.bucket, iargs.input_bucket_prefix, '.json')
         json_data = json.load(open(json_file_path))
         json_data['metadata'].setdefault('weather_model', []).append(iargs.weather_model)
         json.dump(json_data, open(json_file_path, 'w'))
--- a/tools/RAiDER/cli/raider.py
+++ b/tools/RAiDER/cli/raider.py
@@ -461,7 +461,15 @@ def calcDelaysGUNW(iargs: list[str] = None) -> xr.Dataset:
 
     p.add_argument(
         '--bucket-prefix', default='',
-        help='S3 bucket prefix containing ARIA GUNW NetCDF file. Will be ignored if the --file argument is provided.'
+        help='S3 bucket prefix which may contain an ARIA GUNW NetCDF file to calculate delays for and which the final '
+             'ARIA GUNW NetCDF file will be upload to. Will be ignored if the --file argument is provided.'
+    )
+
+    p.add_argument(
+        '--input-bucket-prefix',
+        help='S3 bucket prefix that contains an ARIA GUNW NetCDF file to calculate delays for. '
+             'If not provided, will look in --bucket-prefix for an ARIA GUNW NetCDF file. '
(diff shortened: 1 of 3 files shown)


tools/RAiDER/cli/raider.py
```diff
@@ -501,6 +509,9 @@ def calcDelaysGUNW(iargs: list[str] = None) -> xr.Dataset:
 
     iargs = p.parse_args(iargs)
 
+    if not iargs.input_bucket_prefix:
+        iargs.input_bucket_prefix = iargs.bucket_prefix
+
```
cmarshak, line 514: So hyp3 takes care of the creation of all the necessary prefixes.

Generally, we should never supply both `input_bucket_prefix` and `bucket_prefix` right?

cmarshak:
Thanks, Joe - this "trivial" update is looking good. I am almost there in terms of understanding, but I think you are three steps ahead of me.

The big question is: How do you envision calling this from the hyp3 api? i.e. `job_type` and relevant `parameters`.

I had another question, but is was answered below:

> Note: HyP3 jobs are all assigned a unique UUID as the Job ID, and we organize the HyP3 content bucket under a prefix, which is the job ID, so Job ID and prefix can be used interchangeably.

Anyways, looks fantastic. Thank you for doing this. Know we will work together to get all those products generated!

cmarshak (approved):

jhkennedy:
> The big question is: How do you envision calling this from the hyp3 api? i.e. job_type and relevant parameters.

@cmarshak it'll look similar to submitting any HyP3 job. In pseudo-code,  I'd expect usage to roughly look like:
```python
import hyp3_sdk as sdk

hyp3 = sdk.HyP3('https://hyp3-a19-jpl.asf.alaska.edu', prompt=True)

...

# get all the campaign jobs
isce_jobs = hyp3.find_jobs(job_type='INSAR_ISCE', start='...', status_code='FAILED')

# filter to jobs that failed the raider step -- TBD; possibly by looking at the runtimes?
failed_jobs = [job for job in isce_jobs if ...]

job_definition = {
    'name': ...,
    'job_type': 'ARIA_RAIDER',
    'job_parameters': {},
}

import copy

prepared_jobs = []
for failed_job in failed_jobs:
    new_job = copy.deepcopy(job_definition)
    new_job['job_parameters']['granules'] = [failed_job.job_id]
    new_job['job_parameters']['weather_model'] = [failed_job.job_parameters['weather_model']]
    prepared_jobs.append(new_job)

raider_jobs = sdk.Batch()
for batch in sdk.util.chunk(prepared_jobs):
    raider_jobs += hyp3.submit_prepared_jobs(batch)
```

Or you could hard-code the weather model like:
```python
job_definition = {
    'name': ...,
    'job_type': 'ARIA_RAIDER',
    'job_parameters': {
        'weather_model': 'HRRR',
    },
}

cmarshak:
That looks excellent. Tagging @mgovorcin too.

tools/RAiDER/cli/raider.py
jhkennedy: @cmarshak for any particular job, HyP3 will provide `bucket` as the HyP3 content bucket and `bucket_prefix` will be *that* job's ID. That's job's files will be uploaded to `s3://bucket/bucket-prefix`.

Becuase we ran RAiDER as a second step in an INSAR_ISCE job, the first step already uploaded a GUNW product to `s3://bucket/bucket-prefix`, and so RAiDER just looked there to download the GUNW.

Or in steps:
1. get GUNW from `s3://bucket/bucket-prefix`
2. add weather model correction to GUNW
3. upload modified GUNW to `s3://bucket/bucket-prefix` clobbering the old one. 

This PR makes it so we can do the above or do:

1. get GUNW from `s3://bucket/input-bucket-prefix`
2. add weather model correction to GUNW
3. upload modified GUNW to `s3://bucket/bucket-prefix`, which is empty


So:
* within HYP3_ISCE jobs where RAiDER is a step, We'll provide `--bucket` + `--bucket-prefix`
  https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/INSAR_ISCE.yml#L138-L141
* as a standalone ARIA_RAIDER job, we'll provide `--bucket` + `--bucket-prefix` + `--input-bucket-prefix`
  https://github.com/ASFHyP3/hyp3/blob/raider-job/job_spec/ARIA_RAIDER.yml#L42-L49
*

jhkennedy:
@cmarshak feel free to merge anytime you're happy/ready to do so!

PR #637, opened 2024-04-24
Ensure orbit credentials for ASF as well as CDSE

### LINKED ISSUE #613: Allow RAiDER to fall back to ASF for Sentinel-1 orbits
Right now, our orbit credential checking function:
https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/s1_orbits.py#L19-L46

* only checks ESA CDSE credentials
* throws an error if ESA credentials aren't found

But `sentineleof` is capable of calling back to ASF via NASA EDL credentials to download orbits if it cannot find them from ESA.

So we should probably instead:
* check for ESA CDSE credentials
* check for NASA EDL credentials
* warn if one exists but not the other
* throw if neither exists

jhkennedy:
This fixes #613 by requiring credentials for _both_  NASA Earthdata and ESA CDSE. 

#613, however, does suggest we should only require one and warn if the other doesn't exist. I can switch to that easily if desired.


bbuzz31 (approved):
Thanks @jhkennedy !

PR #640, opened 2024-04-26
For HyP3 jobs, also download and upload browse image

jhkennedy:
When running RAiDER independently in HyP3 on previous INSAR_ISCE HyP3 jobs, all three product files will need to be downloaded from the previous HyP3 job and re-uploaded so that when this job is ingested, all the files are available. 

Previously, only the `.nc` and `.json` files were downloaded and uploaded. This downloads and uploads the remaining browse image.


cmarshak:
Looks good to me.

So you download/upload png regardless if you are running Raider separately with failed `job_id` or with full workflow.

Hopefully the s3 upload function clobbers the old browser image. That would be my only conern

cmarshak (approved):

jlmaurer (approved):

PR #641, opened 2024-04-26
Release v0.5.0 -- orbits and stand-alone HyP3 job support

jhkennedy:

PR #642, opened 2024-04-27
Adjust tests for png upload/download

jhkennedy:
Notably, the tests aren't run on a fork, which is why we see them failing in #641 but not in #640.

I'm not using a fork here.


jlmaurer (approved):
Looks like everything passed this time

PR #645, opened 2024-05-01
Use hyp3-lib for RAiDER Downloading of orbits

cmarshak:
I believe long-term we will want to use `sentineleof`. To ensure the processing campaign can be successfully completed, I am using `hyp3lib` pinning major version 3 as done for our topsapp plugin: 

- https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/localize_orbits.py
- https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/environment.yml#L19

When we are ready, should be as simple as swapping out `get_orbits_from_slc_ids_hyp3lib` with `get_orbits_from_slc_ids`.

### DIFF
--- a/tools/RAiDER/s1_orbits.py
+++ b/tools/RAiDER/s1_orbits.py
@@ -97,7 +97,7 @@ def get_orbits_from_slc_ids_hyp3lib(
 
     orbits = []
     for scene in slc_ids:
-        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=('ASF', 'ESA'))
+        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=('ESA', 'ASF'))
         orbits.append(orbit_file)
 
     orbits = sorted(list(set(orbits)))
--- a/tools/RAiDER/s1_orbits.py
+++ b/tools/RAiDER/s1_orbits.py
@@ -97,7 +97,7 @@ def get_orbits_from_slc_ids_hyp3lib(
 
     orbits = []
     for scene in slc_ids:
-        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=('ESA', 'ASF'))
+        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=('ASF', 'ESA'))
         orbits.append(orbit_file)
 
     orbits = sorted(list(set(orbits)))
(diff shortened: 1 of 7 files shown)


tools/RAiDER/s1_orbits.py
```diff
@@ -78,6 +79,32 @@ def get_orbits_from_slc_ids(slc_ids: List[str], directory=Path.cwd()) -> List[Pa
...
+
+    orbit_directory = orbit_directory or 'orbits'
+    orbit_dir = Path(orbit_directory)
+    orbit_dir.mkdir(exist_ok=True)
+
+    orbit_fetcher =  get_orb.downloadSentinelOrbitFile
+
+    orbits = []
+    for scene in slc_ids:
+        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=('ASF', 'ESA'))
```
jhkennedy, line 100:
```suggestion
        orbit_file, _ = orbit_fetcher(scene, str(orbit_dir), esa_credentials=esa_credentials, providers=( 'ESA', 'ASF'))
```
ASF's orbit enpoint, s1qc, cannot handle the ARIA processing load, even in the constrained setup we have right now. Since hyp3lib releases tokens when it's done, I'd suggest hitting ESA first

cmarshak: Thinking a little more - I am reverting to ASF first again.

https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/blob/dev/isce2_topsapp/localize_orbits.py#L36

I think that's what we do for the topsapp step and that is doing ok. I feel like it shouldn't matter about order, but rather keep it as-is.

tools/RAiDER/s1_orbits.py
```diff
@@ -78,6 +79,32 @@ def get_orbits_from_slc_ids(slc_ids: List[str], directory=Path.cwd()) -> List[Pa
...
+    # Populates env variables to netrc as required for sentineleof
+    _ = ensure_orbit_credentials()
+    esa_username, _, esa_password = netrc.netrc().authenticators(ESA_CDSE_HOST)
+    esa_credentials = esa_username, esa_password
+
+    orbit_directory = orbit_directory or 'orbits'
+    orbit_dir = Path(orbit_directory)
+    orbit_dir.mkdir(exist_ok=True)
+
+    orbit_fetcher =  get_orb.downloadSentinelOrbitFile
```
jhkennedy, line 96:
```suggestion
    orbit_fetcher = get_orb.downloadSentinelOrbitFile
```

jhkennedy (approved):
I didn't do a thorough review, but this looks good to me.

PR #656, opened 2024-06-24
Re-add example run configurations

garlic-os:
## Description

This change adds example commands back to the CLI, something that was missing since the switch to the new YAML format. Now, each one's name is listed in `raider.py -h`, and can be generated/copied to your working directory just like the default `raider.yaml`. Further, that same `raider.yaml` is now called `template.yaml` to fit its new context among these example configurations that are based off of it.

## Motivation and Context

When RAiDER's CLI configuration was changed from arguments to YAML files, it lost [the portion of the help blurb that gives example commands](https://github.com/dbekaert/RAiDER/blob/ce0a0a6f5cce54abb86db968bdb28efc5a52a57e/tools/RAiDER/runProgram.py#L23).  I personally found them useful when familiarizing myself with RAiDER, so I was sad to see them go. As part of my recent work updating dbekaert/RAiDER-docs, I found myself needing to update those examples to the new format anyway, so since I already have them, integrating them back into the main repo was just the next logical step.

## How Has This Been Tested?

Tests were added in a new file `test_generate_config.py`:
- One test for each example to ensure it and any associated data files are properly produced
- Tests to ensure that RAiDER stops to confirm before overwriting any files

image

## Type of change

## Checklist:


jlmaurer (approved):
LGTM

PR #672, opened 2024-07-15
Linting RAiDER codebase

garlic-os:
## Description

This PR aims to eliminate all warnings produced by `ruff` in the `tools/RAiDER/` directory. Ideally, when `$ ruff check tools/RAiDER` is run, it should produce no output.

## Motivation and Context

Linting is an important part of preserving a project's code quality, making the code appear more consistent and readable, as well as helping to reveal potential mistakes. Clearer code also generally improves our experience as developers on the project and helps mitigate technical debt. With ruff's help I even identified a small bug (#669). As a preliminary step toward instituting a lint check in RAiDER's CI pipeline, the whole project needs to be linted through.

## Type of change

## Checklist:

- [ ] ~~I have written new tests for your core changes, as applicable.~~
- [ ] ~~My change requires a change to the documentation.~~
- [ ] ~~I have updated the documentation accordingly.~~


garlic-os:
This is the work I have so far. Tests are all passing for me except the ones that check RAiDER's numerical outputs. Looking into what could have affected the numbers.

garlic-os:
`calcDelays` CSV output is missing the wetDelay, hydroDelay, and totalDelay columns. Investigating

garlic-os:
Hm, delay value checks still failing.

garlic-os:
All tests passing. Phew! Just merged with latest version of dev too. I will have to stop short of the whole tools directory though, because my employment period with Dr. Maurer is set to end when the S&T semester begins again. Hopefully this can already improve the project in the ways I mentioned and set it up for further improvement down the road.

jlmaurer:
@garlic-os looking pretty good! One final request, can you add a line or two to the [CONTRIBUTING.md](https://github.com/dbekaert/RAiDER/blob/dev/CONTRIBUTING.md) document on how to format using ruff prior to pushing a PR? Thanks!

test/test_gnss.py
```diff
@@ -40,11 +38,11 @@ def temp_file():
 
 
 def test_getDateTime():
-    f1 = '20080101T060000'
-    f2 = '20080101T560000'
-    f3 = '20080101T0600000'
-    f4 = '20080101_060000'
-    f5 = '2008-01-01T06:00:00'
+    f1 = Path('20080101T060000')
```
jlmaurer, line 41: @garlic-os I don't think these should be Paths, aren't they string inputs?

test/test_gnss.py
```diff
@@ -59,10 +57,10 @@ def test_addDateTimeToFiles1(tmp_path, temp_file):
     df = temp_file
 
     with pushd(tmp_path):
-        new_name = os.path.join(tmp_path, 'tmp.csv')
-        df.to_csv(new_name, index=False)
-        addDateTimeToFiles([new_name])
-        df = pd.read_csv(new_name)
+        new_path = tmp_path / 'tmp.csv'
```
jlmaurer, line 60: Is this compatible with running on Win11?

jlmaurer: Meaning the "/" symbol. I think windows and *nix systems have opposite conventions for which slash they use as directory delimiters.

jlmaurer: There's a number of other locations where this happens too, so same comment on those.

jlmaurer (commented):
@garlic-os looks very good overall, just had a few comments you can find in the review.

test/test_gnss.py
garlic-os: These are actually file names, and getDateTime does path operations on them. In usual execution, these will be GPS Zenith delay CSV files with a timestamp like this in their full filename.

test/test_gnss.py
garlic-os: Yes, this is compatible with all platforms. This is actually the Python division operator, which pathlib.Path overloads as its way of joining Paths/strings together. On Linux/Mac/other Unix-based platforms, the resulting path will have forward slashes ("/"), and on Windows it will have backslashes ("\\").

jlmaurer (approved):

PR #685, opened 2024-09-13
Fix check containment

### LINKED ISSUE #684: [BUG] checkContainment fails on global zenith HRES
**Describe the bug**
when running on a global HRES file (e.g., bounding box: -90 90 -180 179.9296785) raider checkContainment fails on zenith.
I suspect if fails on slants as well.

[run_global.yaml.txt](https://github.com/user-attachments/files/16935778/run_global.yaml.txt) 

**Full error message**
```
Invalid extension NetCDF for cube. Defaulting to .nc
Output cube spacing: 0.09 degrees
WARNING: Bounds extend past +/- 180. Results may be incorrect.
Output SNWE: [-90.0, 90.0, -180.18, 180.09]
Starting to run the weather model c
alculation
Requested date,time: 20200101, 00:00
Beginning weather model pre-processing
WARNING: Processed weather model already exists, please remove it ("weather_files/HRES_2020_01_01_T00_00_00_90S_90N_180W_180E.nc") if you want to download a new one.
WARNING: The processed weather model file already exists, so I will use that.
Extent of the weather model is (xmin, ymin, xmax, ymax):-179.52, -89.52, 179.51, 89.51
Extent of the input is (xmin, ymin, xmax, ymax): -180.18, -90.00, 180.09, 90.00
Weather model point bounds are -90.00/90.00/-180.18/180.09
Query datetime: 2020-01-01 00:00:00
ERROR: The weather model passed does not cover all of the input points; you may need to download a larger area.
```

I am able to change line 92 in processWM.py to make it run by adding HRES to the exception (in addition to HRRR). There are `nans` in the result that need to be checked though.

I also forced the interpolated cube datatype to float32 to conserve memory (line 200 in delay.py: `np.zeros(..., dtype=np.float32)`)

Note that the giant processed weather model can be compressed as follows with xarray:
```
src = 'HRES_2020_01_01_T00_00_00_90S_90N_180W_180E.nc'
dst = 'HRES_2020_01_01_T00_00_00_90S_90N_180W_180E_compressed.nc'
ds = xr.open_dataset(src)
comp = dict(zlib=True, complevel=9)
encoding = {var: comp for var in ds.data_vars}
ds.to_netcdf(dst, encoding=encoding)

### PR DESCRIPTION
**Pull request by jlmaurer**
## Description

Attempts to fix #684 

## Motivation and Context

If the global bounding box is larger than the world box [-90, 90, -180, 180], then the current code gets unhappy (see #684). This checks whether the world box fits inside the requested box and sets the requested box to equal the world. 

## How Has This Been Tested?

- added unit tests to check the basic functionality 
- Made a bunch of formatting tweaks to fix ruff issues

## Type of change

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,32 +12,36 @@ jobs:
       - checkout
 
       - run:
-          name: Install micromamba
+          name: Install MiniForge
           shell: /bin/bash -l
           command: |
             apt update --yes && apt-get upgrade --yes
             apt install -y --no-install-recommends wget ca-certificates git
 
             cd ${HOME}
-            curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba
-            eval "$(${HOME}/bin/micromamba shell hook -s posix)"
+            wget -O  Miniforge3.sh "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
+            bash Miniforge3.sh -b -p "${HOME}/conda"
+            source "${HOME}/conda/etc/profile.d/conda.sh"
+            source "${HOME}/conda/etc/profile.d/mamba.sh"
 
       - run:
           name: Create RAiDER environment
-          shell: /bin/bash -l
+          shell: /bin/bash -xl
           command: |
-            eval "$(${HOME}/bin/micromamba shell hook -s posix)"
+            source "${HOME}/conda/etc/profile.d/conda.sh"
+            source "${HOME}/conda/etc/profile.d/mamba.sh"
           
             PYTHON_VERSION="<< parameters.python-version >>"
             sed -i "/python>=/c\ - python=${PYTHON_VERSION}" environment.yml
-            micromamba create -f environment.yml
+            mamba env create -f environment.yml
 
       - run:
           name: Install raider and check environment
-          shell: /bin/bash -l
+          shell: /bin/bash -xl
           command: |
-            eval "$(${HOME}/bin/micromamba shell hook -s posix)"
-            micromamba activate RAiDER
+            source "${HOME}/conda/etc/profile.d/conda.sh"
+            source "${HOME}/conda/etc/profile.d/mamba.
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: .circleci/config.yml [hunk 1]
```diff
@@ -12,32 +12,36 @@ jobs:
...
 
       - run:
           name: Install raider and check environment
-          shell: /bin/bash -l
+          shell: /bin/bash -xl
           command: |
-            eval "$(${HOME}/bin/micromamba shell hook -s posix)"
-            micromamba activate RAiDER
+            source "${HOME}/conda/etc/profile.d/conda.sh"
+            source "${HOME}/conda/etc/profile.d/mamba.
```
jhkennedy, line 43:
```suggestion
            source "${HOME}/conda/etc/profile.d/mamba.sh"
```

This should get you off and running again

tools/RAiDER/interpolator.py
```diff
@@ -119,20 +130,89 @@ def fillna3D(array, axis=-1, fill_value=0.0):
...
+    dem_path: Path to the DEM file.
+    points: List of (latitude, longitude) tuples.
+
+    Returns:
+    List of elevation values corresponding to the input points.
+    """
+    import rasterio
+
+    # with rasterio.open(dem_path) as src:
+    breakpoint()
```
jhkennedy, line 168:
```suggestion
```
This is causing the tests to hang and fail due to inactivity

test/test_intersect.py
```diff
@@ -57,6 +58,7 @@ def test_cube_intersect(tmp_path, wm):
         latf = os.path.join(SCENARIO_DIR, "lat.rdr")
         lonf = os.path.join(SCENARIO_DIR, "lon.rdr")
 
+        breakpoint()
```
jhkennedy, line 61:
```suggestion
```

jlmaurer:
@jhkennedy thank you I think we got it! If you're good to merge we can start rebasing the other PRs

jhkennedy (approved):
This looks good to me! I haven't done a highly detailed review, but I don't see anything here that would require it.

PR #695, opened 2025-01-08
Add in logic to not fail when no HRRR weather model is available

jacquelynsmale:
## Type of change

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
 
       - run:
           name: Install MiniForge
-          shell: /bin/bash -l
+          shell: /bin/bash -xl
           command: |
             apt update --yes && apt-get upgrade --yes
             apt install -y --no-install-recommends wget ca-certificates git
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 [Unreleased]
+### Added
+* [695](https://github.com/dbekaert/RAiDER/pull/695) - Add `AUTO` option for `raider.py` to avoid failure when there is no `HRRR` weather model available
+
 ### Changed
 * [672](https://github.com/dbekaert/RAiDER/pull/672) - Linted the project with `ruff`.
 * [683](https://github.com/dbekaert/RAiDER/pull/683) - Fixed a naive datetime and added default template to template generation argument
--- a/docs/WeatherModels.md
+++ b/docs/WeatherModels.md
@@ -100,3 +100,5 @@ Reference: __[The Modern-Era Retrospective Analysis for Research and Application
     **Note**: this step has been included in the conda install of RAiDER, thus can be omitted if one uses the recommended conda install of RAiDER
    
     **Note**: PyDAP v3.2.1 is required for now (thus specified in the above pip install command) because the latest v3.2.2 (as of now) has a known [bug](https://colab.research.google.com/drive/1f_ss1Oa3VzgAOd_p8sgekdnLVE5NW6s5) in accessing and slicing the GMAO data. This bug is expected to be fixed in newer versions of PyDAP.
+
+# TODO : ADD IN DOCUMENTATION FOR THE AUTO SELECT
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,7 @@ dependencies:
  - lxml
  - matplotlib
  - netcdf4
- - numpy<2
+ - numpy<2  # https://github.com/dbekaert/RAiDER/issues/691
  - pandas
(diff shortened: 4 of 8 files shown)


asjohnston-asf:
FYI, here are the test cases we itemized for this feature:

1. Install RAiDER
    1. `git clone git@github.com:dbekaert/RAiDER.git`
    2. `cd RAiDER`
    3. `mamba env create -f environment.yml`
    4. `mamba activate RAiDER`
    5. `python -m pip install -e .`
2. Download the two test input files
    1. `aws s3 cp s3://asj-dev/raider/ . --recursive`
3. Add a `.netrc` entry for `dataspace.copernicus.eu`
    1. `machine dataspace.copernicus.eu login *** password ***`
4. Run the tests
    1. `raider.py ++process calcDelaysGUNW --weather-model {model} --file {file}`
    2. "adds data layers" takes 10ish minutes and adds four data layers to the .nc file. Since this edits the file in place, be sure to grab a fresh copy before running subsequent tests
    3. "nothing to do" takes ~3 seconds and doesn’t change the input .nc file
    4. "invalid choice" takes ~3 seconds and doesn’t change the input .nc file
    5. "location is unavailable" takes ~15 seconds and doesn’t change the input .nc file

v0.5.3 | HRRR | None | AUTO
-- | -- | -- | --
S1-GUNW-D-R-071-tops-20241204_20241029-135046-00118W_00042N-PP-5700-v3_0_1.nc | adds data layers | nothing to do | invalid choice
S1-GUNW-A-R-091-tops-20241205_20241123-233558-00074W_00037S-PP-014f-v3_0_1.nc | location is unavailable | nothing to do | invalid choice

New Version | HRRR | None | AUTO
-- | -- | -- | --
S1-GUNW-D-R-071-tops-20241204_20241029-135046-00118W_00042N-PP-5700-v3_0_1.nc | adds data layers | nothing to do | adds data layers
S1-GUNW-A-R-091-tops-20241205_20241123-233558-00074W_00037S-PP-014f-v3_0_1.nc | location is unavailable | nothing to do | nothing to do

asjohnston-asf:
I think a more complete changelog entry for this feature would be:

### Added
* [695](https://github.com/dbekaert/RAiDER/pull/695) - Added a new `AUTO` option to the `--weather-model` argument of `raider.calcDelaysGUNW`. Selecting this option will automatically select the weather model to use based on the established strategy for standard ARIA S1 GUNW products. The current strategy is to use `HRRR` where available and `None` otherwise.

@cmarshak @dbekaert does that communicate our intent well enough? I'm open to keywords other than `AUTO` if there's one you prefer.

asjohnston-asf:
There's code that strongly suggests the `--weather-model=HRRR` was intended to do what we want when run via HyP3. The if block at https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/cli/raider.py#L604-L632 is run for jobs run via HyP3, and the availability checks at L620 and L627 are both intended to exit with a success status code without making any changes to the GUNW product.

However, [check_weather_model_availability](https://github.com/dbekaert/RAiDER/blob/dev/tools/RAiDER/aria/prepFromGUNW.py#L98) invoked on L627 raises `ValueError` on L137 when the scene is outside the HRR's spatial extent (rather than returning False), causing the program to exit with a failure status code. This mismatch between the calling code expecting True/False and the function raising an exception looks like an unintended bug to me.

This is the only place `check_weather_model_availability` is called. I propose we update the function to return False instead of raising ValueError on L627. Then HyP3 can submit jobs using weather_model=HRRR and we'll get the behavior we want for standard products.

cmarshak:
Yes - it's honestly, quite confusing because it supports different use cases. Auto seems ok, but there are some nuances that not being considered. The HRRR data is actually not continuous in that sometimes we expect it to be available (because a product is over US CONUS or AK), but the date is not available: https://github.com/ACCESS-Cloud-Based-InSAR/East-Coast-VLM-Sprint/blob/dev/check_hrrr_availability/Check%20HRRR%20CONUS.ipynb. That is, spatially we expect HRRR but temporally it is not available.

The "bug" that @asjohnston-asf you are seeing is the result of having it exit successfully there.

I did a lot of the logic to help, but not sure what the best way to proceed.

ALSO! to keep in mind that eventually "auto" should use the HRES data that is being stored at ASF - this should be implemented at some point.

cmarshak:
> I think a more complete changelog entry for this feature would be:
> 
> ### Added
> * [695](https://github.com/dbekaert/RAiDER/pull/695) - Added a new `AUTO` option to the `--weather-model` argument of `raider.calcDelaysGUNW`. Selecting this option will automatically select the weather model to use based on the established strategy for standard ARIA S1 GUNW products. The current strategy is to use `HRRR` where available and `None` otherwise.
> 
> @cmarshak @dbekaert does that communicate our intent well enough? I'm open to keywords other than `AUTO` if there's one you prefer.

I think this makes sense - see my previous comments for the caveats regarding exiting...

PR #702, opened 2025-02-04
exit successfully with no changes when run via hyp3 with weather_model=HRRR and outside HRRR coverage

asjohnston-asf:
Pivoting to a shorter-term change than https://github.com/dbekaert/RAiDER/pull/695 , I believe this is sufficient to let ASF provide standard ARIA_S1_GUNW products on demand via HyP3. Standard Jobs will always be submitted with weather_model=HRRR. Jobs within HRRR coverage will have the data layers added. Jobs outside HRRR temporal or spatial coverage will exit successfully without making any changes to the GUNW file.

Expanding the logic for a weather model for a standard product in the future, say to add the use of additional weather models, would require more in-depth changes such as those in #695.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -37,6 +37,8 @@ def test_gunw_path_factory() -> Callable:
             file_name = 'S1-GUNW-D-R-071-tops-20200130_20200124-135156-34956N_32979N-PP-913f-v2_0_4.nc'
         elif location == 'alaska':
             file_name = 'S1-GUNW-D-R-059-tops-20230320_20220418-180300-00179W_00051N-PP-c92e-v2_0_6.nc'
+        elif location == 'philippines':
+            file_name = 'S1-GUNW-D-R-032-tops-20200220_20200214-214625-00120E_00014N-PP-b785-v3_0_1.nc'
         else:
             raise NotImplementedError
         return TEST_DIR / 'gunw_test_data' / file_name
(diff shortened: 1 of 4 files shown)


asjohnston-asf:
@cmarshak @jhkennedy I believe this one line change is sufficient to let ASF produce standard products on-demand and make the correct weather model choice on the user's behalf. Are you (or others) available to review and merge to `dev`? That would let ASF validate the changes end-to-end in one of our non-production HyP3 environments.

I'll take a stab at resolving https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/issues/191 today or tomorrow if that helps take something off the ARIA's team plate, since ASF will also need that fix for an end-to-end test.

test/conftest.py
```diff
@@ -37,6 +37,8 @@ def factory(location: str = 'california-t71') -> Path:
             file_name = 'S1-GUNW-D-R-071-tops-20200130_20200124-135156-34956N_32979N-PP-913f-v2_0_4.nc'
         elif location == 'alaska':
             file_name = 'S1-GUNW-D-R-059-tops-20230320_20220418-180300-00179W_00051N-PP-c92e-v2_0_6.nc'
+        elif location == 'philippines':
+            file_name = 'S1-GUNW-D-R-032-tops-20200220_20200214-214625-00120E_00014N-PP-b785-v3_0_1.nc'
```
cmarshak, line 41: Thanks for including this.

cmarshak (approved):
Nice work including the added tests and test data. I think this is ready to be merged and released.

asjohnston-asf:
I'm not a maintainer and don't have permissions to merge to `dev` myself. @cmarshak or @jhkennedy could you do the honors?

PR #713, opened 2025-02-28
Pulling refs/tags/v0.6.0 into dev

dbekaert:
Fast-forward of main to dev failed!


jhkennedy:
Bah, woops, forgot to check the tag vs version; fixing.

PR #715, opened 2025-03-01
Update raiderCombine and fix coveralls

jlmaurer:
- `raiderCombine.py` was not working when the raider time was not an integer multiple of 5 minutes (the GNSS sample rate)
- coveralls (evidently???) changed its behavior at https://coveralls.io/builds/69674378. Now `coverage run -m pytest ...` seems to be required. 

## Description

- Coveralls was reporting zero % coverage (0/0). Changing the pytest line to include "coverage run -m pytest..." seems to have solved the issue and coverage is now back up. 
- raiderCombine was assuming that raider would have the same time as the GNSS delays; however, this is not the case with the new time interpolation that allows for any time to be returned. I added logic to round to the nearest 5 minutes (the GNSS sample rate) as well as fixed up the column naming logic

## Motivation and Context

Coveralls is back and at record coverage!!!

## How Has This Been Tested?

Unit tests are passing

## Type of change

- [x] Bug fix [TO COVERALLS] (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
             mamba activate RAiDER
 
             COV_OPTIONS=`python -c "import importlib;print(*(' --cov='+p for p in importlib.util.find_spec('RAiDER').submodule_search_locations))"`
-            pytest -m "not long" test/ ${COV_OPTIONS} --cov-report=
+            coverage run -m pytest -m "not long" test/ ${COV_OPTIONS} --cov-report=
 
       - run:
           name: Report coverage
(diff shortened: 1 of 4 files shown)


jlmaurer:
@jhkennedy I think I found the issue; although I haven't tracked down why `coverage run -m` is now needed whereas it wasn't before. See main PR description for details.

cmarshak:
I don't know what raiderCombine does nor do I have any experience with coveralls. I trust Joe with the latter. Good luck.

PR #716, opened 2025-03-13
use s1_orbits to fetch orbit files from AWS

asjohnston-asf:
Uses [s1_orbits](https://pypi.org/project/s1-orbits/) to fetch orbit files from the [Registry of Open Data on AWS](https://registry.opendata.aws/s1-orbits/), which will be faster and less error prone than using hyp3lib.

Similar treatment as https://github.com/ACCESS-Cloud-Based-InSAR/DockerizedTopsApp/releases/tag/v0.4.0.

I'll take some time this afternoon to run `calcDelaysGUNW` for a handful of GUNW products with the old and new code and confirm they both ultimately download the same orbit files.


tools/RAiDER/gnss/downloadGNSSDelays.py
```diff
@@ -18,7 +18,7 @@
 
 
 # base URL for UNR repository
-_UNR_URL = 'http://geodesy.unr.edu/'
+_UNR_URL = 'https://geodesy.unr.edu/'
```
asjohnston-asf, line 21: Unrelated to the orbit changes, geodesy.unr.edu is not responding over http today. Requests hang indefinitely, causing the GNSS related tests to fail. The test worked when they last ran on Mar 4.

Requests over https work fine, so updating the scheme here resolved my issues.

asjohnston-asf:
I'm not inclined to take action on the coveralls check. Coverage changed from 3,326/3,518 to 3,204/3,395. Untested lines dropped from 192 to 191, and every line I changed is tested, but I removed enough code that the untested 191 lines are a slightly bigger percent of the total.

jlmaurer (approved):
LGTM!

PR #731, opened 2025-06-17
Fix GMAO downloader

### LINKED ISSUE #673: [BUG] `models/gmao.py:149` Undefined names dt, crs
**Describe the bug**
Ruff reports that the variables `dt` and `crs`, used below, are not defined.
https://github.com/dbekaert/RAiDER/blob/d617dacf585004c4b20ac138f93366013c3efbf6/tools/RAiDER/models/gmao.py#L147-L151

RAiDER proceeds without any error because this line is wrapped in a bare try-except. I suspect this try-except was not meant to catch this error in particular, but since it didn't specify an exception it must have slipped through.

 - RAiDER git tag:
commit 2a348023801fc1f7fb21427d3ae63150530e2382 (HEAD -> dev, tag: v0.5.2, upstream/main, upstream/dev, upstream/HEAD)
Merge: 223d862 2b7190c
Author: Jeremy Maurer <maurer.jeremy@gmail.com>
Date:   Mon Jul 15 11:54:29 2024 -0500

    Merge pull request #670 from dbekaert/dev
    
    v0.5.2
 - OS: Arch Linux 5.15.153.1-microsoft-standard-WSL2 on Windows 10

nate-kean:
## Description

Fixes `GMAO._fetch()`. With this, RAiDER should now be able to download data from GMAO.

> [!IMPORTANT] 
> RAiDER is downloading GMAO data now, I think verifying the _correctness_ of the data is out of my wheelhouse. I would like to ask someone with expertise in this project's math for help in checking the outputs as part of the review.

1. Adds missing logic in `GMAO._fetch()` to extract q from the fetched data, to the case where the input time is past `T0`.
2. Adds a line to perform `np.meshgrid()` on the lats and lons, [as in the MERRA2 implementation of `_fetch()`](https://github.com/dbekaert/RAiDER/blob/b837793/tools/RAiDER/models/merra2.py#L90), as this appeared necessary to present the lats and lons to `writeWeatherVarsXarray()` with the expected dimensions.  
    - **This is the portion that I am the most unsure about.**
3. Defers loading `.netrc` credentials to pydap:
    - The credentials appeared to be hard-coded dummy credentials before.
    - Explicit credential passing was deprecated anyway, and this actually also resolves a deprecation warning.
    - I added this change to the MERRA2 class as well.
    - This obsolesces `read_EarthData_loginInfo()`, so this PR also removes `read_EarthData_loginInfo()` and its tests.
4. Makes `GMAO._fetch()`'s call to `writeWeatherVarsXarray()` raise exceptions it encounters instead of suppressing them.

Fixes https://github.com/dbekaert/RAiDER/issues/673.

## Motivation and Context

This PR is meant to fix the issues surrounding RAiDER's GMAO class's ability to download data. From my investigation, I found that these things stood in RAiDER's way of properly fetching GMAO data:
1. As the GMAO class evolved over time, it came to be that it would work for some requested dates, but not others; through at least a couple commits, one branch was tested but not the other.
2. A certain line that had errors in it was wrapped in a try-catch block that suppressed all exceptions coming from it, making tests against it look like they were passing, when really they were not, and no data was being saved.
3. Dummy API credentials were hardcoded in, making it so that requests would fail even if everything else was correct.

And so this PR was aimed primarily at fixing those issues.

## How Has This Been Tested?

All test continue to pass, and I can now see RAiDER saving an .nc file when requested to download GMAO data. Work still needs to be done to verify that the data it is saving is what would be expected.

## Type of change

## Checklist:


jlmaurer:
@garlic-os one quick comment on this. Is there a reason to remove the Earthdata creds testing from the test suite? There may be DEMs or other datasets that RAiDER might use that require Earthdata credentials.

nate-kean:
My reasoning for deleting it was that this fix makes it so RAiDER doesn't have a need for `read_EarthData_loginInfo()` anymore. Part of the fix is moving away from a deprecated pydap function, `pydap.cas.urs.setup_session()`, and over to pydap's new method for loading EarthData credentials where it handles it all on its own. In doing so, the project's one use of `read_EarthData_loginInfo()` is removed.

But I see what you mean; if we believe the ability to read EarthData credentials ourselves would be useful in the future, I wouldn't be opposed to keeping it.

jlmaurer (approved):

nate-kean (commented):
We would also want to restore `read_EarthData_loginInfo()` itself in `utilFcns.py`

Oop, my bad, spoke too soon, there it is

jlmaurer (approved):

jlmaurer:
> We would also want to restore `read_EarthData_loginInfo()` itself in `utilFcns.py`
> 
> Oop, my bad, spoke too soon, there it is

Ah yeah sorry, was just updating on the fly. If this passes I'll merge it in!

nate-kean:
Had a meeting with Jeremy, and I have reverted the commits of Jeremy's that restored `read_EarthData_loginInfo()` per his request. We decided it would be better after all to remove it for now, since its functionality is now completely taken care of by the netrc and pydap packages, rather than leave code in RAiDER that would go unused.

PR #733, opened 2025-06-18
Fix Coveralls

nate-kean:
## Description

Several things were leading to an incorrect coverage number:
1. The `--cov` argument to pytest was not being generated properly. It is possible that what did this used to work and was broken in a Python update. This is 0f5359e
2. Some changes were still left to be made to make RAiDER fully compatible with the new CDS API, specifically when it came to `setup_from_env()`, which the CircleCI runner uses. Those updates were assigned to me, so this one is my bad. Fixed in 7d6b109
3. After (1) was fixed, pytest began having a different problem where test runs would end with a warning saying "no data was collected". Running coverage.py through pytest-cov instead of directly fixed this. I must admit, I don't know why this fixes it, but pytest-cov does give a "slightly different `sys.path`", so that is my guess. Fixed in 20258d

## Motivation and Context

This change will correct RAiDER's test coverage number to more accurately reflect the number of lines of project code the test suite is testing. The benefits of that include awareness of what parts of the code are yet not verified to be correct and a greater and greater confidence in the correctness and robustness of the program over time.

## How Has This Been Tested?

Testing for this kind of change happens mostly in comparing the outputs of the CI runners to what they were producing previously.
### Fixes in build process
The project's build output used to show an error relating to setting up the command to run pytest, as seen [here](https://app.circleci.com/pipelines/github/dbekaert/RAiDER/2063/workflows/c49c77b7-505d-4551-be43-4a01a97eaead/jobs/3286) (Run Unit Tests, line 1643) (from #728).  
image
The build reports in this PR no longer have that error.

### Fixes in Coveralls report
The project's coverage reading used to be erroneously based off the number of lines covered in `test/`, as can be found here](https://coveralls.io/builds/74178366). Now, coverage readings are generated with respect to `RAiDER/tools/`, as can be found [here](https://coveralls.io/builds/74226516), which means coverage is now counting the actual project code, and not other folders.

Before
image

After! :)
image

## Type of change

## Checklist:


nate-kean:
Comment on the failing checks:
1. The drastic coverage/coveralls change is to be expected, since coveralls is now checking `tools/RAiDER/` instead of `test/` and `.circleci/`, as it was before.
2. ~~build-3.9 failed because it timed out installing the conda environment. Although, it has been _reliably_ timing out for me lately... that might be another issue?~~ build 3.9 failure is now #739

jlmaurer (approved):

PR #743, opened 2025-06-25
Use DAP4 where possible, suppress warning otherwise

### LINKED ISSUE #736: UserWarning: PyDAP was unable to determine the DAP protocol defaulting to DAP2
**Describe the bug**
Use of HTTPS instead of DAP2 or DAP4 as the protocol for URLs opened in pydap throws this warning.
```
... pydap/handlers/dap.py:134: UserWarning: PyDAP was unable to determine the DAP protocol defaulting to DAP2.
DAP2 is consider legacy and may result in slower responses. 
Consider replacing `http` in your `url` with either `dap2` or `dap4` to specify the DAP protocol
(e.g. `dap2://<data_url>` or `dap4://<data_url>`).  For more 
information, go to https://www.opendap.org/faq-page.
```

Unfortunately, trying to use dap2 or dap4 when accessing data from opendap.nccs.nasa.gov, as we do, causes a myriad of strange errors within pydap. First, it can't find the file, then if it does it fails to parse due to malformed XML, and then even if you fix that, the data in the response still is not structured the way pydap expects. Unless this is user error on my part, this may be a problem that could only be fixed through changes to pydap or switching to an alternative.

**To Reproduce**
```py
import pydap.client
url = 'https://opendap.nccs.nasa.gov/dods/GEOS-5/fp/0.25_deg/assim/inst3_3d_asm_Nv'
pydap.client.open_url(url)
```

 - RAiDER git tag: dev [b837793](https://github.com/dbekaert/RAiDER/commit/b837793e2a74254894dcd32b6e46aebbeb48fd72)
 - OS: Linux 6.6.87.2-microsoft-standard-WSL2x86_64

**Additional context**
Add any other context about the problem her

nate-kean:
## Description

1. Changes the template for MERRA2 URLs to use DAP4, since the data store MERRA2 uses gives OPeNDAP responses that pydap is able to handle.
2. Suppresses warnings from pydap while fetching GMAO data, since the data store GMAO uses gives OPeNDAP responses that pydap can't handle.

## Motivation and Context

Resolves #736, in which pydap throws warnings when requesting data using HTTPS instead of DAP2 or DAP4; with these changes, RAiDER no longer emits this warning.

It turns out that we can use DAP4 for MERRA2 without issue, but GMAO's data store and pydap do not work together when using either DAP protocol, so continuing to use HTTPS and suppressing the warning is the best we can do.

## How Has This Been Tested?

No new tests needed to be added, and all existing tests continue to pass.

## Type of change

## Checklist:


jlmaurer:
@garlic-os this seems to have some changes that have already been incorporated and doesn't seem to show the full changelog. Can you just check that this correctly rebased on the current dev branch? Thanks!

nate-kean:
That's funny, the line I added had [disappeared](https://github.com/dbekaert/RAiDER/pull/743/commits/d3dfebdad798705eb730a9a0a65cd73bc9550737) from this PR's commit history... wrote it back again.

jlmaurer (approved):

PR #751, opened 2025-08-01
Fix for ERA-5 API change

### LINKED ISSUE #750: [BUG] Change to ERA-5 API causing ERA-5 runs to fail
**Describe the bug**
Runs using ERA-5 are failing with a NoWeatherModelData error. Digging deeper, the underlying cause of this exception is that `RAiDER.models.ecmwf._makeDataCubes` is failing while trying to access `t`, because the received dataset has stopped containing `t` or `q`:  
https://github.com/dbekaert/RAiDER/blob/43d4036/tools/RAiDER/models/ecmwf.py#L324

It turns out that the Climate Data Store has changed the behavior/data RAiDER receives from a request for ERA-5 reanalysis data, and now datasets that used to contain `lnsp`, `z`, `t`, and `q`, at model levels 1-137 now only contain `lnsp`, `z`, at just model level 1. I have found out that you can still get `t` and `q` (at all 137 levels) if you request them separately.

The ECMWF has documentation on this: [ERA5: compute pressure and geopotential on model levels, geopotential height and geometric height](https://confluence.ecmwf.int/display/CKB/ERA5%3A+compute+pressure+and+geopotential+on+model+levels%2C+geopotential+height+and+geometric+height). It indicates that now we indeed have to request `lnsp` and `z` separate from `t` and `q`, and we must compute `z` at all levels beside surface level ourselves.

**To Reproduce**
Configuration to reproduce in a full run:
```yaml
weather_model: ERA5
date_group:
  date_start: "20200103"
time_group:
  time: "23:00:00"
aoi_group: 
  lat_file: "test/scenario_1/geom/lat.dat"
  lon_file: "test/scenario_1/geom/lon.dat"
los_group:
  zref: 20000
```
Throws NoWeatherModelData.

 - RAiDER git tag: [43d4036](https://github.com/dbekaert/RAiDER/commit/43d40361342210143c43a0f58a2f1782c1f3e721)

nate-kean:
## Description

(Fixes #750)  
Modifies the fetch routine for ERA-5 to:
1. request lnsp/z separate from t/q, and
2. compute z for all model levels ourselves.

## Type of change

## Checklist:


nate-kean:
Here is a fix for the "model levels" level type. Untested for pressure levels and not incredibly efficient (there are some calculations that are being done multiple times unnecessarily), but I'm pushing this out due to the urgency of this issue. If this is merged in its current state I will file another pull request later with more tests and optimization.

tools/RAiDER/delay.py
```diff
@@ -73,8 +73,8 @@ def tropo_delay(
             wm_proj = CRS.from_epsg(4326)
 
     # get heights
-    with xr.load_dataset(weather_model_file) as ds:
-        wm_levels = ds.z.values
+    with xr.open_dataset(weather_model_file) as ds:
```
jlmaurer: @nate-kean is this a breaking change in xarray or just a different way to load data?

tools/RAiDER/interpolator.py
```diff
@@ -156,13 +156,14 @@ def interpolate_elevation(dem_path: Union[Path, str], x: np.ndarray, y: np.ndarr
     Interpolates elevation values from a DEM to scattered points.
```
jlmaurer: I'd recommend pulling out the files that are purely formatting issues and submitting those in a separate PR.

tools/RAiDER/losreader.py
```diff
@@ -71,6 +71,8 @@ def is_Projected(self):
     def ray_trace(self):
```
jlmaurer: Same thing here. We can very quickly submit these as a separate PR just for formatting and typo changes.

tools/RAiDER/models/era5t.py
```diff
@@ -4,14 +4,14 @@
```
jlmaurer: Same thing here in terms of formatting. you don't want to have too many lines of code changed in one PR.

jlmaurer (changes requested):

tools/RAiDER/delay.py
nate-kean: This is actually mostly an optimization. `load_dataset` and `open_dataset` work the same, but `open` lazy loads the data, which in places where only a small part of a large dataset is being used, this can save a lot of memory.

jlmaurer (approved):
LGTM

PR #757, opened 2025-08-14
Switch to micromamba to fix CircleCI timeout

nate-kean:
## Description

Fixes two problems that were preventing the CircleCI runners from passing/were causing them to time out:
1. **Problem:** A mysterious configuration issue began occurring while trying to run mamba commands. This happened from no change on our end, and appears to be caused by an update to MiniForge mamba.  
**Solution:** Switched to micromamba. This turns out to be the preferred conda/mamba flavor for automated runners because it is extremely simple and downloads and installs very fast. It also does not suffer from the problem that we were experiencing with MiniForge mamba. We continue to use only the `conda-forge` channel, as micromamba comes preconfigured with just it.
2. **Problem:** The "Create RAiDER environment" step was attempting to activate the RAiDER environment before it had been created, through the profile.d script.  
**Solution:** Delayed adding the line to activate the RAiDER environment to after it is created, so the runner only starts to attempt to activate it in the steps after when it has been created.

## Motivation and Context

Fixes the problem that has been occurring in our CI builds lately that has caused them all to fail or time out. After this is merged, we will want to re-run the CI tests for recent pull requests that were failing.

## How Has This Been Tested?

The `ci/circleci` checks in this PR are now passing.

## Type of change

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,10 +33,9 @@ jobs:
           shell: /bin/bash -xl
           command: |
             PYTHON_VERSION="<< parameters.python-version >>"
-            sed -i "/python>=/c\ - python=${PYTHON_VERSION}" environment.yml
             # Create RAiDER enivronment
-            micromamba env create --yes -f environment.yml
-            # Activate the RAiDER env for the rest of this script's steps
+            micromamba env create --yes -f environment.yml python=${PYTHON_VERSION}
+            # Activate the RAiDER env for the rest of the script's shells
             echo "micromamba activate RAiDER" >> /etc/profile.d/mamba-raider-circleci.sh
 
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,21 +12,20 @@ jobs:
       - checkout
 
       - run:
-          name: Install MiniForge
+          name: Install micromamba
           shell: /bin/bash -xl
           command: |
             apt update --yes && apt upgrade --yes
             apt install --yes --no-install-recommends wget ca-certificates git
 
-            cd ${HOME}
-            wget -O Miniforge3.sh \
-              "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
-            bash Miniforge3.sh -b -p "${HOME}/conda"
-
-            # Put mamba's shell init into a profile.d file, in order to activate
-            # it even in non-interactive terminals
-            "${HOME}/conda/bin/mamba" shell hook --shell bash --root-prefix "${HOME}/conda" \
-              > /etc/profile.d/mamba-raider-circleci.sh
+            # Put micromamba's shell init script into a profile.d file instead
+            # of .bashrc, in order to activate it even in CircleCI's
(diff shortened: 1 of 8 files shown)


jlmaurer:
@nate-kean thanks for this!
@jhkennedy @dbekaert do you see any issues with pinning major versions? The idea is to speed up package resolving and prevent breaking changes from coming into RAiDER without being able to address them first.

jhkennedy:
> do you see any issues with pinning major versions? 

@jlmaurer yes, pinning major version guarantees RAiDER will break in the future, and is generally considered a bad practice, e.g., https://iscinumpy.dev/post/bound-version-constraints

And pinning the development environment here, but not the package dependencies, will mean you won't follow the typical user install path and miss dependency issues that users will hit. 

Generally, if you want a known working environment, tools like conda-lock and pixi are designed to give you a locked environment you can use for that. You should still have an unlocked testing workflow so you hit user issues in CI though. 

Unfotunately, there isn't a good way to automatically open PRs to update dependencies in conda/pixi lock files like dependabot does, which would be the right way to provide a locked environment and ensure you're seeing dependency issues as they appear, but there is a workaround for pixi:
https://pixi.sh/latest/integration/ci/updates_github_actions/

It's _fine_ short term to add major upper-bounds to the dev environment, but you it should be a _very_ short term. Ideally, in order of importance:
* `pyproject.toml` and, if provided, `environment.yml` would list dependencies and only exclude _known incompatibilities_. With core conda dependencies, pixi is the best way to provide this
* a lockfile that provides a known working environment
* a CI tool that opens PRs to test new dependency and upadate the lock file
* CI that regularly tests the minimum bounds of dependencies to make sure check the lower bounds and find new incompatibilities

nate-kean:
Wow, you bring up a lot of good points I didn't consider. You have convinced me, I shouldn't have added these upper bounds where we don't need them. Thanks for your input, @jhkennedy, I will remove the upper bounds I added.

jlmaurer:
@nate-kean I think you can rebase on dev and hopefully that last test will pass. I think #759 and/or #760 updated the `test_GUNW` test.

nate-kean:
Yes, I have done that. But now the runners are taking over an hour again. I'm starting to doubt whether my changes made any difference...

nate-kean:
New upper bounds removed. I also had to make a couple fixes to tests to make all the tests pass. The tests in question even failed for me on dev, so I suspect it may have been upstream changes in dependencies which only started applying to me after I had reinstalled/updated my environment for this PR.

But now the CircleCI runners are taking over an hour to solve their environment again sometimes. It's making me wonder if these bouts of slowdowns are even being caused by us, especially since they have historically started after changes that shouldn't affect environment creation at all, and "solved" themselves after just trying again in a week. I might split these test fixes into their own pull request and close this one, because it doesn't appear to have really fixed the issue.

I am considering a new approach now: add a lock file ([conda/conda-lock](https://github.com/conda/conda-lock)) to the project, and the dependency solutions for each Python version we test will be computed ahead of time and no longer have to be done by the CircleCI runners.

nate-kean:
Closing; moved progress to #762

PR #761, opened 2025-08-22
Add Python 3.13 compatibility

nate-kean:
## Description

Adds Python 3.13 as a tested version in the CircleCI configuration, and introduces some small changes to the following tests to make them work in Python 3.13:
- `test_GUNW_hyp3_metadata_update`
- `test_GUNW_hyp3_metadata_update_different_prefix`
- `test_cube_timemean`
- `test_cube_weighting`

## Motivation and Context

`test_GUNW_hyp3_metadata_update` series: updated `gunw_schema.json` to the version of the JSON Schema spec that `jsonschema` 4+ uses. `jsonschema` upgraded from `==3.2.0` to `>=4` to make it compatible with Python 3.13.

`test_cube_` series: Was also failing in lower Python versions. I cherry-picked the MERRA2-relevant bit from #752, and the issue fixed itself. The code for interacting with the dataset seemed invalid for one returned from `pydap`, but it fit perfectly with one from `xarray`.

I also included a refactor to `test_temporal_interpolate.py` that helped me debug this. `calcDelays` is now called directly instead of through `subprocess.run`, so the debugger will break at exceptions that occur within it.

## How Has This Been Tested?

Tests now all pass in CircleCI on all our supported Python versions, including Python 3.13. I can confirm that all tests pass locally on Python 3.13, as well.

## Type of change

## Checklist:


jlmaurer (approved):
LGTM. @cmarshak you may want to take a quick look before this gets merged. Tests all pass so I assume this is consistent with existing GUNW file structure.

PR #762, opened 2025-08-22
Add lockfiles to RAiDER

nate-kean:
## Description

Adds lockfiles to RAiDER via [conda/conda-lock](https://github.com/conda/conda-lock). Makes creating a RAiDER environment at least 10-40x faster.[^1]
1. Adds lockfiles generated against the current dev branch for every version of Python the project currently supports.
    - `conda-lock.yml` at project root is for our highest supported Python version, currently 3.12.
    - I put the others in `.circleci/`, since I expect they will be chiefly used for the CircleCI runners.
2. Updates README and CONTRIBUTING to reflect how the addition of the lockfile system will affect developers.
3. Configures the CircleCI runners to take advantage of the lockfiles.

[^1]: The "Create RAiDER environment" step used to take the CircleCI runners at least 15 minutes, and sometimes would time out after solving for 60 minutes. Now it takes about 1m30s including the extra time to install `conda-lock`.

## Motivation and Context

- This is primarily meant as an alternative solution to #757 to **fix the present CircleCI timeouts**. Giving the CI runners access to a lockfile allows them to create their RAiDER environment without solving it each time, which for this project is a step that sometimes would take so long that it would singlehandedly elapse the runners' 60 minute timeout.
- This will also allow developers, particularly those who are setting up RAiDER for the first time, to **create local environments faster**.
- **One caveat** is that this will add friction while submitting changes to `environment.yml`: any change to this file will now require also regenerating the lockfiles to keep them up to date with it. In other package managers/languages, this is done automatically, but unfortunately with conda we will have to remember to do this ourselves.  
I wrote and included a script that aids in doing this for us. See https://github.com/dbekaert/RAiDER/pull/762/files#diff-a590e97322c09b68d2481c457c63f1611410317c627f1019375a93aa3d39c2d3

## How Has This Been Tested?

The runners now all pass.

## Type of change

- [x] Bug fix (cures CircleCI timeouts)
- [x] New feature (makes CircleCI tests and new env installs faster)
- [x] Breaking change (developers have an extra thing to do when modifying `environment.yml`)

## Checklist:


tools/RAiDER/models/merra2.py
```diff
@@ -109,7 +109,7 @@ def _fetch(self, out: Path) -> None:
             + '.nc4'
         )
 
-        stream = pydap.client.open_url(url)
+        stream = xr.open_dataset(url, decode_times=False, engine='pydap')
```
jlmaurer, line 112: I think this snuck in here from #752

nate-kean: Good eye. Yes, I included part of #752 here to make all of the tests pass, because without it, tests that used `MERRA2._fetch()` were failing. For Python versions 3.9-3.12, this is the bare-minimum of the PR that is necessary. 

Would you rather I take it out here? Just know that the tests will probably start to display as failing.

jlmaurer: Got it. 👍 Nope that's fine

jlmaurer (approved):

PR #770, opened 2025-09-15
Bugfix/hack: Always use EPSG:4326 with GeocodedFile to fix crash

nate-kean:
## Description

Adds a line that transforms all bounding boxes the GeocodedFile AOI receives to EPSG:4326.

## Motivation and Context

Rest of the code is not properly set up to handle GeocodedFile AOIs that are not in EPSG:4326, which causes problems with EPSG:32617 interferograms like from ASF On Demand. I lack the time to fix this properly, but I did end up doing this much so that RAiDER would work with my code for the gemlab ASF On Demand notebook jlmaurer/gemlab#11. I offer what I have in case you might want to accept it and/or use it as a jumping-off point for a more in-depth solution.

In the case when the user inputs a geocoded file with projection other than EPSG:4326, this change fixes a crash, at the cost of more transformations in this code path, and any precision errors or other drawbacks that may come with it. A more thoughtful solution may be able to avoid doing a conversion here.  
In all other cases (other projections, other input AOI types), nothing is changed/affected.

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/llreader.py
+++ b/tools/RAiDER/llreader.py
@@ -332,7 +332,7 @@ class GeocodedFile(AOI):
 
         self._filename = path
         self.p = rio_profile(path)
-        self._bounding_box = transform_bbox(rio_extents(self.p), dest_crs=4326, src_crs=32617)
+        self._bounding_box = transform_bbox(rio_extents(self.p), dest_crs=4326, src_crs=self.p['crs'])
         self._is_dem = is_dem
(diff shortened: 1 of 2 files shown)


dbekaert:
Hi Nate,

Suggest we create/update a ticket where this support is flagged to be added
or add reference to this pr change here. If i understand this here refers
purely to the bounding box and the export projection as chosen by user is
still supported.

D.

PR #771, opened 2025-11-05
Update concat script to handle multiple subdirs + add noise-corrected residual

sssangha:
## Description
Support in concatenation script to pass wildcard for `--raiderDir` and `--gnssDir` added back in. Before proposed changes, only single path arguments were supported. Now both single path arguments and wildcards are supported.

I figured this was a change that needed to be reintroduced after legacy bash scripts collating data for global analyses I had between multiple directories were failing.

In addition, now the noise correction residual column is directly passed and added to the final output files. Before I had separate wrapper scripts to add this in

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have successfully ran tests with your changes locally.

### DIFF
--- a/tools/RAiDER/gnss/processDelayFiles.py
+++ b/tools/RAiDER/gnss/processDelayFiles.py
@@ -437,11 +437,6 @@ def main(
     # estimate residual
     dfc['ZTD_minus_RAiDER'] = dfc['ZTD'] - dfc[raider_delay]
 
-    # estimate noise corrected residual using GNSS sigZTD
-    dfc["Residual_minus_sigZTD"] = (
-        (dfc["ZTD_minus_RAiDER"] ** 2) - (dfc["sigZTD"] ** 2)
-    ) ** 0.5
-
     print('Total number of rows in the concatenated file: ' f'{dfc.shape[0]}')
     print(f'Total number of rows containing NaNs: {dfc[dfc.isna().any(axis=1)].shape[0]}')
     print('Merge finished')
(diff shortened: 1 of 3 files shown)


tools/RAiDER/gnss/processDelayFiles.py
```diff
@@ -429,6 +437,11 @@ def main(
     # estimate residual
     dfc['ZTD_minus_RAiDER'] = dfc['ZTD'] - dfc[raider_delay]
 
+    # estimate noise corrected residual using GNSS sigZTD
```
jlmaurer: This is not quite right because it assumes that the standard deviation of this column will be the same as the standard deviation of the residuals minus the mean standard deviation of the ZTD, which is not true:
```
Var[X - Y] = Var[X] + Var[Y] - 2Cov(X, Y)
```
Here "X" can be the residuals and "Y" can be the sigZTD. What we want (or at least what we say in the paper) is 
```
\sigma_{model} = \sigma_X - \sigma_Y
```
which is quite different. You can't calculate the difference on a per-observation basis, only for the overall population variances. I'd suggest to leave this part off.

jlmaurer (dismissed):
See comment. Otherwise LGTM

tools/RAiDER/gnss/processDelayFiles.py
sssangha: Ok sounds good, yes I think this was passed as a vestige from before that we ended up dropping anyways

sssangha:
> See comment. Otherwise LGTM

Sounds good! I reverted that change and left everything else

jlmaurer (approved):

jlmaurer:
I'll merge once all tests pass

PR #774, opened 2025-11-06
In combine workflow, consistently map GNSS coordinates

sssangha:
In the raider combine workflow, if there are any inconsistencies in the Lat/Lon values between the two input files, then this will lead to dropped matches downstream as localtime estimation would be inconsistent. Thus here changes are introduced to consistently map Lat/Lon coordinates for each GNSS ID between both CSVs.

Also, I had encountered this warning when applying changes to a given dataframe:
```
SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame. 
```

This arises because `pandas` can’t be sure whether the change is applied to the original DataFrame or just a temporary view. To circumvent this, before applying changes an independent dataframe is made (e.g. `dfr = dfr.copy()`) to essentially "refresh" the dataframe

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * [743](https://github.com/dbekaert/RAiDER/pull/743) - Switched from HTTPS to DAP4 for retrieving MERRA2 data, and suppressed a warning for using DAP4 for GMAO data where doing so is not possible.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


sssangha:
@jlmaurer Not sure what to do about the coveralls failure.

I read that you could relax fail conditions in `.coveragerc` or the CI config (e.g., require a minimum coverage of 50% or only fail if it drops >0.5%), since here `0.06%` is very small

jlmaurer (approved):

PR #775, opened 2025-11-06
Bugfix for handling lat/lon input in downloadGNSS

sssangha:
When a user specifies a bounding box for the `raiderDownloadGNSS.py`, there was a bug with the necessary internal -180/180 to 0/360 conversion to match with the UNR source convention.

I've pushed a patch and an explicit modification to the parser such that there is no ambiguity in that a 0 longitude centered convention is enforced for the bounding box input

## How Has This Been Tested?

You can run this test, where download data globally over a week

raiderDownloadGNSS.py --date 20250930 20251006 1 -b '-90 90 -180 180' --returntime 00:00:00 --out GNSS_obs_00 --cpus 8

Before this patch, only the western hemisphere got downloaded since the -180/180 to 0/360 conversion wasn't handled properly internally (see the screenshots of the station distribution before vs after the patch).
Screenshot 2025-11-05 at 7 06 09 PM

Screenshot 2025-11-06 at 1 56 31 PM

## Type of change

## Checklist:

### DIFF
--- a/test/test_downloadGNSS.py
+++ b/test/test_downloadGNSS.py
@@ -50,7 +50,7 @@ def test_fix_lons_positive_to360():
 
 def test_fix_lons_negative_to360():
     lon = -100.0
-    assert fix_lons(lon, to360=True) == 800.0
+    assert fix_lons(lon, to360=True) == 80.0
 
 
 def test_fix_lons_0_to180():
--- a/test/test_downloadGNSS.py
+++ b/test/test_downloadGNSS.py
@@ -43,24 +43,24 @@ def test_in_box_outside():
     assert not in_box(lat, lon, llbox)
 
 # Test fix_lons with various longitudes
-def test_fix_lons_positive():
-    lon = 200.0
-    assert fix_lons(lon) == -160.0
+def test_fix_lons_positive_to360():
+    lon = 80.0
+    assert fix_lons(lon, to360=True) == 260.0
 
 
-def test_fix_lons_negative():
-    lon = -220.0
-    assert fix_lons(lon) == 140.0
+def test_fix_lons_negative_to360():
+    lon = -100.0
+    assert fix_lons(lon, to360=True) == 800.0
 
 
-def test_fix_lons_positive_180():
-    lon = 180.0
-    assert fix_lons(lon) == 180.0
-
-
-def test_fix_lons_negative_180():
-    lon = -180.0
+def test_fix_lons_0_to180():
+    lon = 0.0
     assert fix_lons(lon) == -180.0
+
+
+def test_fix_lons_360_to180():
+    lon = 360.0
+    assert fix_lons(lon) == 180.0
 
 
 # Test get_ID with a valid line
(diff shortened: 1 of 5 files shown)


jlmaurer:
@sssangha this looks good to me but can you include the screenshots you mentioned? They don't seem to show up.

sssangha:
> @sssangha this looks good to me but can you include the screenshots you mentioned? They don't seem to show up.

Done, looks like I didn't add them in. Sorry about that.

Btw I had to tweak the unit tests given that I went back and overhauled and simplified the longitude conventions. Now the input bbox is always expected to be centered at 0 longitude (i.e. -180 to 180) and it is passed as such. Before there were a bunch of confusing conversions which converted back and forth between this convention and 0 to 360, it was very hard to track.

tools/RAiDER/gnss/downloadGNSSDelays.py
```diff
@@ -231,13 +229,56 @@ def in_box(lat, lon, llhbox):
...
-def fix_lons(lon):
-    """Fix the given longitudes into the range `[-180, 180]`."""
-    fixed_lon = ((lon + 180) % 360) - 180
-    # Make the positive 180s positive again.
-    if fixed_lon == -180 and lon > 0:
-        fixed_lon *= -1
-    return fixed_lon
+def fix_lons(lon, to360: bool = False):
+    """
+    Convert longitude values between [-180, 180] and [0, 360] conventions.
```
jlmaurer, line 234: Wouldn't it be -360 to 0?

test/test_downloadGNSS.py
```diff
@@ -43,24 +43,24 @@ def test_in_box_outside():
...
+    lon = 0.0
+    assert fix_lons(lon) == -180.0
 
 
-def test_fix_lons_negative_180():
-    lon = -180.0
-    assert fix_lons(lon) == -180.0
+def test_fix_lons_360_to180():
+    lon = 360.0
+    assert fix_lons(lon) == 180.0
```
jlmaurer, line 63: Why does 360 go to 180? shouldn't it be 0?

jlmaurer (commented):
@sssangha this looks fine to me, but now I'm confused because UNR have longitudes reported as -360-0 degrees. It looks like some of the original code used this convention and some did not. Does you new code handle both positive and negative longitude conventions? Also had a few questions on some of the details.

sssangha:
> @sssangha this looks fine to me, but now I'm confused because UNR have longitudes reported as -360-0 degrees. It looks like some of the original code used this convention and some did not. Does you new code handle both positive and negative longitude conventions? Also had a few questions on some of the details.

Yeah it's a bit confusing, but the part where the conversion from the UNR convention (-360-0) is handled separately in one line here:
https://github.com/dbekaert/RAiDER/blob/dd45383038da5d06538b03af534ac7b5e50e7e6a/tools/RAiDER/gnss/downloadGNSSDelays.py#L85

The `fix_lons` function wasn't even used in `downloadGNSSDelays.py` before, nor is it now. It seems it just is there in case somebody wanted to convert between -180,180 and 0,360 with their own inputs, or in other scripts.

jlmaurer (approved):

PR #792, opened 2026-01-14
Add support to subset files in time for stats analysis

sssangha:
Added support within the `raiderCombine.py` pipeline to support temporal subsetting, which is introduced as a new option `--timeinterval`.

If a valid argument is provided, temporal sampling is performed in 2 phases:
1. With a 1 day buffer encompassing the specified `--timeinterval` in order to not lose valid observations due to time zone changes.
2. With the explicit specified `--timeinterval` once 'Localtime' has been computed

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/gnss/types.py
+++ b/tools/RAiDER/gnss/types.py
@@ -1,6 +1,6 @@
 import argparse
 from pathlib import Path
-from typing import Optional, List
+from typing import List, Optional
 
 class RAiDERCombineArgs(argparse.Namespace):
     raider_file: Path
--- a/tools/RAiDER/gnss/types.py
+++ b/tools/RAiDER/gnss/types.py
@@ -1,14 +1,18 @@
 import argparse
 from pathlib import Path
-from typing import Optional
-
+from typing import Optional, List
 
 class RAiDERCombineArgs(argparse.Namespace):
     raider_file: Path
-    raider_folder: Path
-    gnss_folder: Path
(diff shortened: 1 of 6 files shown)


tools/RAiDER/gnss/types.py
```diff
@@ -1,14 +1,18 @@
...
-from typing import Optional
-
+from typing import Optional, List
 
 class RAiDERCombineArgs(argparse.Namespace):
     raider_file: Path
-    raider_folder: Path
-    gnss_folder: Path
+    raider_folder: List[Path]
+    gnss_folder: List[Path]
```
jlmaurer, line 8: @sssangha could this be `Union[List[Path], Path]`?

sssangha: I made the logic more consistent overall in the code, but the logic is as follows:
- Support wildcard/multiple path input in the parser
- Split into a list discrete individual paths and push them through as individual paths through the downstream function.

I verified the outputs I get now vs before this little clean-up are consistent though, so the files are being pushed through as we'd expect.

sssangha: Hello @jlmaurer , following up. I worked this out and enforced consistency in the workflow. Please let me know what you think. Thanks

sssangha:
Hello @jlmaurer , I think this PR is ready to merge. Please let me know what you think. Thanks

jlmaurer (approved):

PR #794, opened 2026-04-08
Fix GNSS data retrieval and downloading functions

sssangha:
Updated the get_stats_by_llh function to prioritize IGS20 coordinates over legacy data.

Enhanced the download_UNR function to include a fallback to the legacy archive when downloading tropospheric delays.

Context: There appears to be ongoing migration of data in the UNR archive such that data from much of the eastern hemisphere is disrupted after 2023. This PR is intended to circumvent this.

## Type of change

## Checklist:

### DIFF
--- a/tools/RAiDER/gnss/downloadGNSSDelays.py
+++ b/tools/RAiDER/gnss/downloadGNSSDelays.py
@@ -69,19 +69,35 @@ def get_station_list(
 
 def get_stats_by_llh(llhBox=None, baseURL=_UNR_URL):
     """
-    Function to pull lat, lon, height, beginning date, end date, and number of solutions for stations inside the bounding box llhBox.
-    llhBox should be a tuple in SNWE format.
+    Pull lat, lon, height for stations inside the bounding box from 
+    both legacy and IGS20 UNR holdings. Prioritizes IGS20 coordinates.
     """
     if llhBox is None:
         llhBox = [-90, 90, -180, 180]
-    S, N, W, E = llhBox
 
-    stationHoldings = f'{baseURL}NGLStationPages/llh.out'
-    # it's a file like object and works just like a file
+    url_legacy = f'{baseURL}NGLStationPages/llh.out'
+    url_igs20 = f'{baseURL}gps_timeseries/IGS20/llh/llh.out'
+    col_names = ['ID', 'Lat', 'Lon', 'Hgt_m']
 
-    stations = pd.read_csv(stationHoldings, sep=r'\s+', names=['ID', 'Lat', 'Lon', 'Hgt_m'])
+    # Read legacy list
+    try:
+        stat_leg = pd.read_csv(url_legacy, sep=r'\s+', names=col_names)
+    except Exception as e:
+        logger.warning("Failed to fetch legacy llh.out: %s", e)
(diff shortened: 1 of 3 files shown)


tools/RAiDER/gnss/downloadGNSSDelays.py
```diff
@@ -69,19 +69,35 @@ def get_station_list(
...
         llhBox = [-90, 90, -180, 180]
-    S, N, W, E = llhBox
 
-    stationHoldings = f'{baseURL}NGLStationPages/llh.out'
-    # it's a file like object and works just like a file
+    url_legacy = f'{baseURL}NGLStationPages/llh.out'
+    url_igs20 = f'{baseURL}gps_timeseries/IGS20/llh/llh.out'
+    col_names = ['ID', 'Lat', 'Lon', 'Hgt_m']
+
+    # Read legacy list
```
jlmaurer, line 82: @sssangha would it make more sense to wrap all of the legacy calls within a single try/except that checks for stations for which IGS20 was successful for all stations and then pulls legacy data only for the subset of stations that are not available? That way the IGS20 is still the default and all the work needed if legacy stations are required is skipped unless they are actually needed.

tools/RAiDER/gnss/downloadGNSSDelays.py
```diff
@@ -173,22 +189,45 @@ def download_UNR(statID, year, writeDir=".", download=False, baseURL=_UNR_URL):
...
-        filepath = download_url(URL, saveLoc)
-        if filepath == "":
-            raise ValueError("Year or station ID does not exist")
+        filename = f"{stat_upper}.{year}.trop.zip"
+        save_loc = os.path.abspath(os.path.join(writeDir, filename))
+
+        # Try IGS20 first
+        filepath = download_url(url_igs20, save_loc)
+
+        # If IGS20 is missing, try legacy
```
jlmaurer, line 215: Does this work if a subset of stations is available but others are not?

jlmaurer (commented):
@sssangha just a couple comments about workflow.

tools/RAiDER/gnss/downloadGNSSDelays.py
sssangha: Hello @jlmaurer I cleaned up the logic a bit here to streamline it and lead with igs20, however we don't know what is missing from igs20 from the subset alone since all references to missing stations are gone. So I need to also pull in the legacy file to be sure nothing slipped through the cracks

tools/RAiDER/gnss/downloadGNSSDelays.py
sssangha: Good catch! Before if a user requests a list of 10 stations and the 3rd one is missing from both archives, the script will crash immediately. It will fail to process the remaining 7 stations, even if they are perfectly valid and exist in the archives. This was the existing behavior.

Since we're iterating over stations/years, we should not raise an exception that halts the entire script. Instead, I tweaked it now to log a warning that the specific station/year failed, and use continue to move on to the next iteration.

sssangha:
@jlmaurer following up on this. Do my adjustments/comments address your concerns? Thanks

jlmaurer (approved):

PR #795, opened 2026-04-15
Use memory efficiently, GNSS geometric heights

jlmaurer:
## Description

xarray makes available two methods for opening datasets, open_dataset and load_dataset. Using open_dataset allows for lazy loading and memory efficiency. This PR change all the locations were `load_dataset` was used to `open_dataset', resulting in much better memory performance for RAiDER. This PR also wraps netcdf dataset opening with try/except/finally logic that ensures the datasets are closed properly. 

In addition, the CDS API call for ERA-5 model levels was doubling calls to the API because the variables were in two pairs. I think this was a holdover from a previous API version. Now all four variables (lnsp, z, t, and q) are called at once. 

Finally, GNSS heights were not getting converted to orthometric heights prior to interpolation in RAiDER. This PR implements a conversion from ellipsoid heights to the geoid only for GNSS stations. 

All unit tests pass locally. 

## Type of change

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ orbits/
 /GUNW*.yaml
 /*.tif
 .coverage*
+TODO.md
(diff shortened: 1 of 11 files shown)


jlmaurer:
@sssangha I think this PR is still good to go if you are able to review it. The integration tests that we were talking about are in #805

jlmaurer:
Closing this and replacing the memory-efficiency part with #808, still working on the geoid correction part.

PR #802, opened 2026-06-12
Multiday download from ECMWF CDS API

jlmaurer:
## Motivation and Context

RAiDER access ECMWF ERA-5 data from the ECMWF CDS API. Due to recent changes in how the `era5-reanalysis-complete ` data is stored (tape archive) data access for model levels is taking a very long for single api calls, on the order of hours. Combine that with two required calls per date and a multi-date request, and you're waiting for days. 

## Description

This PR addresses the issue by bundling the request for all the days together when requesting model levels. The complexity is that pressure level data is still stored on a harddrive, so access is faster, but the multi-day download API does not work the same way. So, this PR updates both the pressure level and model level requests to use the fastest possible data access method. 

This PR builds on #795, that PR should be merged first. 

## How Has This Been Tested?

Unit tests pass locally, and more unit tests have been added to test the new functionality. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have added an explanation of what your changes do and why you'd like us to include them.
- [x] I have written new tests for your core changes, as applicable.
- [x] I have successfully ran tests with your changes locally.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #805, opened 2026-07-09
Fix a couple of issues that deal with the RAiDER bias compared to GNSS

jlmaurer:
This PR came about through comparing RAiDER-derived ZTD to GNSS ZTD and tracking down the source of the differences with the help of Claude Fable. 

**NOTE**: This PR builds on #802; that PR should be integrated before this one. 

## Description

There were two funadmental changes needed, one of which was already in place but was not being called. 
1. RAiDER was using trapz integration, which introduces a systematic bias when the function is convex or concave. In the case of pressure-level data with only ~27 levels, the bias is especially high (~2 cm). The fix was to directly use the fact that the hydrostatic delay is exponential in shape and use the the log-mean rule, ∫N dz = Δz·(N₀−N₁)/ln(N₀/N₁)), which is exact for N(z) = N₀e^(−z/H). 
2. GNSS station heights are ellipsoidal heights, but the RAiDER calculation is done on geoidal heights. The conversion to geoid heights is built in to `llreader.py` in #795 , but it was never called. This PR adds the needed infrastructure so that a user can specify a CRS for the GNSS station heights, and those will get converted to geoid heights. 

## Motivation and Context

These two issues together all but eliminate the bias between GNSS and RAiDER ZTD estimates. 

## How Has This Been Tested?

Unit tests pass locally

The unit test for `getZTD` broke because it used a linear function to test. This PR changes the linear function to an exponential, since the new functionality assumes exponential decrease in hydrostatic delay with elevation. 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have added docstrings to the code to document these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ orbits/
 /GUNW*.yaml
 /*.tif
 .coverage*
+TODO.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Changedp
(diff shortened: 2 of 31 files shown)


jlmaurer:
Changing this to draft until I verify the integration method.

jlmaurer:
This is addressed by #822

PR #808, opened 2026-08-05
Update xarray usage in RAiDER to use lazy loading

jlmaurer:
xarray makes available two methods for opening datasets, open_dataset and load_dataset. Using open_dataset allows for lazy loading and memory efficiency. This PR change all the locations were load_dataset was used to `open_dataset', resulting in much better memory performance for RAiDER. This PR also wraps netcdf dataset opening with try/except/finally logic that ensures the datasets are closed properly.

The PR also fixes a small issue where ECMWF requests were being handled in two API requests instead of one. 

All unit tests pass locally.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 * [811](https://github.com/dbekaert/RAiDER/pull/811) - Fixed a north-south flip in the ECMWF/ERA-5 model-level reader. `_load_model_level` reversed only the level axis of the geopotential cube (`z[::-1]`) while `t`, `q`, and `lnsp` had their latitude axis reversed, leaving heights mirrored in latitude relative to the meteorology. Surface height and surface pressure were anti-correlated (r = -0.54 where physics requires ~+1); sea-level ZTD over a 3-degree box spanned 1621-3233 mm instead of 2239-2395 mm. The error is antisymmetric in latitude, so it largely cancels in a domain average while being severe at individual points.
 
+### Changed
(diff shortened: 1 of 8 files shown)


sssangha:
Hello @jlmaurer 

I reviewed and tested the changes locally, and added two small safety enhancements:

1. **Robust Resource Cleanup with `ExitStack`:**
   To guarantee that all opened datasets are cleanly closed—even if an exception occurs mid-loop or during `.to_netcdf()`—I wrapped the file loading using `contextlib.ExitStack`:

   ```python
   from contextlib import ExitStack

   with ExitStack() as stack:
       datasets = [stack.enter_context(xr.open_dataset(f)) for f in wfiles]
       # ... perform calculations and save to netcdf ...
    ```
2. **Preventing In-Place Variable Reassignment**
   Assigning `ds_out = datasets[0]` mutates `datasets[0]` directly while its disk handle is open. Updated this to explicit shallow copying:
   
   ```
   ds_out = datasets[0].copy(deep=False)
   ```


**Future Consideration on Dask Chunking**
Standard lazy loading via `open_dataset` handles the current memory performance improvements well. If processing ultra-large weather model grids ever hits RAM limits down the road, we can revisit introducing explicit chunking `(e.g., chunks={})` to leverage Dask streaming directly to disk.


LGTM overall! I'm ready to approve if these changes look good to you.

jlmaurer:
@sssangha thanks, looks good! I checked the PR again and found a couple small issues related to when I pulled it out from #810, so I fixed those (this is in `ecmwf.py` and related tests). If you want to just take a quick look at those, we should be good here.

sssangha (approved):
LGTM

PR #809, opened 2026-08-05
Add automatic datatime column as needed when running raiderCombine.py 

jlmaurer:
## Description

Running `raiderDownloadGNSS.py` and then `raiderCombine.py` directly fails because the datetime column may not exist. This PR allows for adding it on-the-fly if needed.

`readZTDFile` now builds the `Datetime` column from whichever time information the file actually provides, in this order:

1. a `Date` column, optionally combined with a `times` seconds-of-day column,
2. an existing `Datetime` column,
3. a `YYYYMMDDTHHMMSS` timestamp parsed from the filename.

When none of those are available it raises an error naming the columns it did find, instead of surfacing a cryptic pandas parsing failure. The filename fallback stamps every row with a single timestamp, which is correct for a single-epoch delay file and lossy for anything covering several epochs, so it logs a warning saying so.

### Two further behavior changes worth reviewing deliberately

These go beyond the datetime column and are easy to miss in the diff:

- **Coordinate backfill.** `Lat`/`Lon`/`Hgt_m` are copied into the GNSS frame from the RAiDER delay file, keyed on station ID, for GNSS ZTD files that carry no station coordinates (e.g. UNR per-station delay files). IDs absent from the RAiDER file get NaNs and are dropped downstream, since they could not have been matched anyway.
- **Hard failure on zero overlap.** The workflow now raises when the RAiDER and GNSS files share no common observations. Previously the empty frames propagated into the per-station variance analysis and surfaced much later as a `KeyError` on a column that was never created (e.g. `sigma_model_neg`). This turns a previously-completing — though meaningless — run into an explicit abort, so it is a behavior change for anyone whose inputs do not overlap.

## Type of change

## Checklist:


jlmaurer:
@sssangha this one is also ready; **note** that this is built on top of #808, so that one should be merged first.

sssangha:
Nice catch @jlmaurer handling missing `Datetime` columns on-the-fly!

I'll summarize things below just to avoid potential incoming redundant commits on my end.

**Key Takeaways:**
* **Datetime Parsing Hierarchy:** The fallback chain (`Date` + `times` -> `Datetime` -> filename timestamp -> explicit error) is clean and avoids cryptic pandas parsing failures.
* **Station Coordinate Backfilling:** Automatically pulling `Lat`/`Lon`/`Hgt_m` from the RAiDER file for UNR-style files keeps user scripts from breaking unnecessarily.
* **Zero-Overlap Guardrail:** Raising an explicit error early when there are no shared observations is much better than letting empty DataFrames trigger obscure downstream key errors.
* **Test Coverage:** Comprehensive additions in `test/test_gnss.py` verifying all fallback paths and boundary cases.

sssangha (approved):
LGTM!

PR #811, opened 2026-08-06
Fixes a north–south flip of the geopotential cube in the ECMWF/ERA-5 model-level reader.

### LINKED ISSUE #806: ERA5 model-level heights corrupted by a latitude-axis flip in `_load_model_level` (regression introduced by #751, released in v0.6.0)
## Summary

Since #751 (commit f81225b, first released in v0.6.0), the ERA5 model-level path produces corrupted height grids over terrain with north–south relief.
The surface geopotential passed to the hydrostatic integration in `_load_model_level` is the **lowest-full-level z with its latitude axis
mirrored** relative to `t`/`q`/`lnsp`. Every column's height coordinate is rigidly shifted by roughly `h(mirror row) - h(own row)`; in our test AOI (central Japan, 4.5° box) the shift reaches **-1.25/+1.27 km**, which corresponds to tens of hPa of apparent pressure error at a fixed height (>100 hPa near the surface at the extremes) and correspondingly corrupted delays. The pressure
*values* themselves are correct — pressure is computed from the a/b coefficients and `lnsp` only — they are just assigned to wrong heights.

Flat AOIs and AOIs with latitude-symmetric terrain show little to no error, which is presumably why this escaped notice.

## Root cause

#751 changed `ECMWF._fetch` to store a **full 137-level z cube** in the weather file (previously z was a single 2D surface field). The axis-flip block in `_load_model_level` was not updated for the new dimensionality
([ecmwf.py L246-L251 at f81225b](https://github.com/dbekaert/RAiDER/blob/f81225b87e2cae19d284a1c0032692458202a875/tools/RAiDER/models/ecmwf.py#L246-L251)):

```python
if lats[0] > lats[1]:
    z: FloatArray3D = z[::-1]        # z is now (level, lat, lon): flips the LEVEL axis
    lnsp: FloatArray2D = lnsp[::-1]  # 2D: flips latitude (correct)
    t: FloatArray3D = t[:, ::-1]     # 3D: flips latitude (correct)
    ...
```

The CDS files in our reproduction come with descending latitude, so this branch runs. `z[::-1]` reverses the model-level axis instead of latitude, and

```python
geo_hgt, p, hgt = self._calculategeoh(z[0, :], lnsp, t, q)
```

([L264](https://github.com/dbekaert/RAiDER/blob/f81225b87e2cae19d284a1c0032692458202a875/tools/RAiDER/models/ecmwf.py#L264))
then selects the lowest model level's z (~10 m above ground — nearly the
right magnitude, which makes the bug subtle) with its latitude axis
**un-flipped**, i.e. mirrored relative to the other fields.

## Evidence

Full analysis, scripts, and raw numbers are in a small public evidence repo, pinned at
[`8d943c1`](https://github.com/s-sasaki-earthsea-wizard/raider-arco-era5/tree/8d943c1cc6ca834fa9faa34d28f19dbfdc2fa054):
see
[`reports/ml_offset_regression.md`](https://github.com/s-sasaki-earthsea-wizard/raider-arco-era5/blob/8d943c1cc6ca834fa9faa34d28f19dbfdc2fa054/reports/ml_offset_regression.md).
Test case: 34.25–38.75N, 137.25–141.75E, 2026-03-01 00:00 UTC, 23×25 grid,
137 levels.

Reproduced against RAiDER `dev` at 1fb9e14 (`ecmwf.py` is unchanged between f81225b, v0.6.0, and current `dev`, so the analysis applies to all three).
The raw input is the netCDF written by RAiDER's own CDS fetch for this AOI and time — not a hand-built file. 
To reproduce, run
[`ml_offset_regression.py`](https://github.com/s-sasaki-earthsea-wizard/raider-arco-era5/blob/8d943c1cc6ca834fa9faa34d28f19dbfdc2fa054/ml_offset_regression.py)
in any environment with RAiDER installed:

```
python ml_offset_regression.py --raw <the ERA-5 ML netcdf RAiDER fetched>
# writes reports/ml_offset_regression.json and prints the verdict
```

1. **Exact identity.** Wrapping `_calculategeoh` to capture its `z_surface`
   argument and comparing bitwise against five candidates: it is
   `np.array_equal` to the **latitude-mirrored lowest-full-level z** over all
   575 columns, and matches no other candidate (own-orientation lowest level,
   model top, reconstructed true surface z, doubled surface z).

2. **Offset regression.** Regressing the per-column height offset
   `dh = (z_used - z_true_surface) / g0` on own and mirrored terrain height:

   | quantity | value |
   |---|---|
   | slope on h_own | -1.000243 |
   | slope on h_mirror | +0.999979 |
   | intercept | +9.87 m (lowest-level height above ground) |
   | R² | 0.99999987 |
   | offset range | -1252 m to +1272 m |

3. **Propagation.** Compared in consistent units (both sides through
   `geo_to_ht`), the error survives into the final ellipsoidal-height grid as
   a nearly rigid per-column shift (within-column std ≤ 7.4 m).

4. **Fix simulation.** Rerunning the same case with the latitude flip
   corrected and the lowest level passed as the surface term collapses the
   error from -1252/+1272 m to a near-constant **+9.8 m**
   (9.44–10.20 m across columns) — exactly the lowest-level height above
   ground, i.e. the regression intercept.

The corruption is deterministic and environment-independent: bit-identical results in two very different environments (pip py3.12/numpy 2.5 vs the repo's conda-lock py3.13/numpy 2.3; `reports/ml_debug_summary.md`).

## Scope

- The corruption fires only when the descending-latitude branch runs
  (`lats[0] > lats[1]` in `_load_model_level`); CDS delivers ERA5 ML with
  descending latitude by default, so in practice it always fires. A file
  already stored with ascending latitude would load correctly.
- **ERA5 ML** with files fetched after #751: silent height corruption as
  above. ERA5T uses the same fetch and loader and is presumably equally
  affected.
- **Files whose z is a 2D surface field** (HRES ML via MARS, where z is
  archived on model level 1 only; ERA5 ML files fetched before #751):
  `z[0, :]` on a 2D array yields a 1D row, and `calcgeoh`'s in-place
  `z_h += ...` raises `ValueError` (non-broadcastable output operand), so
  these likely fail to load at all. Flagged as suspected — we have not run a
  real HRES file through it — but it is a falsifiable prediction: the HRES
  ML path should crash with that `ValueError` inside `_load_model_level`.

## Suggested fix

The smallest correct change, in our view: **store the true surface z in the fetched file**. `calcgeoh`'s contract requires a 2D surface geopotential, and `_fetch` already holds exactly that (`ds_lnsp_z['z']`) before discarding it. 
Saving it as a separate variable (e.g. `z_surface`) alongside the full cube, with the loader preferring `z_surface` and falling back to a 2D `z` for old-format files, keeps HRES/pre-#751 compatibility in the same change and removes the 2D-vs-3D ambiguity that caused this bug.

Alternatives:

- Hotfix: flip z as 3D (`z[:, ::-1]` / `z[..., ::-1]`) and pass `z[-1]` instead of `z[0]`. Verified to collapse the error on our case, but bakes in
  the near-constant ~+10 m lowest-level bias measured above.
- Longer term: skip the load-time re-integration of heights entirely — the file already contains the full z cube computed at fetch time from the true surface geopotential (pressure still needs the a/b + `lnsp` computation, which never depended on `z_surface`).

Happy to open a PR for whichever direction you prefer, including a small network-free regression test (asymmetric-latitude grid, descending lats, asserting the flip acts on the latitude axis of a 3D z).

## Disclosure

The diagnosis and harness were written with assistance from AI agents (Claude Fable 5, GPT-5.6 Sol, Kimi-k3).
I ran the diagnostic scripts myself and reproduced the reported numbers
before posting.

jlmaurer:
## Description

Thanks to @s-sasaki-earthsea-wizard for catching this bug! 

Fixes a north–south flip of the geopotential cube in the ECMWF/ERA-5 model-level reader.

In `ECMWF._load_model_level`, the branch that corrects ECMWF's descending-latitude ordering reversed the wrong axis of `z`:

```python
if lats[0] > lats[1]:
    z    = z[::-1]        # z is (level, lat, lon) -> reverses LEVEL
    lnsp = lnsp[::-1]     # (lat, lon)             -> reverses lat
    t    = t[:, ::-1]     # (level, lat, lon)      -> reverses lat
    q    = q[:, ::-1]     # -> reverses lat
```

`z` needs **both** axes reversed: the latitude axis to stay aligned with `t`/`q`/`lnsp`, and the level axis because `z[0]` is subsequently passed to `_calculategeoh` as the surface geopotential. Reversing only axis 0 satisfied the second requirement while silently violating the first, leaving the height field mirrored in latitude against the meteorology.

The visible symptom is columns whose pressure profile is displaced vertically by however much terrain differs between a latitude and its mirror — e.g. a sea-level node reporting constant pressure from −500 m up to 1900 m, which is ~1900 m of fictitious sea-level-density air worth roughly +490 mm of spurious delay.

**Affects ERA-5, ERA-5T and HRES on model levels**, which is the default level type for those models. The pressure-level path is untouched.

### When this was introduced

Introduced in `f81225b` ("Update for ERA-5 API changes", #751, merged 2025-09-05, released in 0.6.0) — so the regression window is 0.6.0 onward, not the whole history.

The statement `z = z[::-1]` is older than the bug and was **correct** when written. What changed is the dimensionality of `z` underneath it:

| | before `f81225b` | after |
|---|---|---|
| `_makeDataCubes` | `np.squeeze(block['z'].values)[0, ...]` → **2D** (lat, lon) | `block['z'].values.squeeze()` → **3D** (level, lat, lon) |
| `z[::-1]` | reverses **latitude** ✓ | reverses **level** ✗ |
| call site | `_calculategeoh(z, lnsp)` | `_calculategeoh(z[0, :], lnsp, t, q)` |

The ERA-5 API change meant CDS would only return `z` at the surface, so RAiDER had to compute the full geopotential cube itself. That moved the `[0, ...]` surface-extraction out of `_makeDataCubes` and into the call site as `z[0, :]`, which made the level reversal newly load-bearing — it is what puts the surface at index 0 — at exactly the moment the latitude reversal silently stopped happening. One statement went from serving one purpose to serving a different one, and reads as deliberate either way.

Fixes #806 

### Why this was not caught earlier

The error is antisymmetric in latitude, so it very nearly cancels in a domain average while being severe at individual points. Over a 3-degree box in southern California the mean ZTD moves only 2360 → 2340 mm, but the spread across the domain collapses from physically impossible to correct:

| diagnostic | before | after |
|---|---|---|
| corr(surface height, −surface pressure) | **−0.5415** | **+0.9992** |
| sea-level ZTD range across the domain | 1621 – 3233 mm | 2239 – 2395 mm |
| domain-mean ZTD | 2360 mm | 2340 mm |

Physics requires the correlation to be ~+1 (high terrain carries low surface pressure).

Validated against UNR GNSS ZTD at four stations (FGNW, MHMS, WLHG, WORG) on four dates in 2020 and 2025, spanning both dry and high-water-vapour conditions. Per-station residuals go from **±400–500 mm to −13 to −66 mm**.

### Tests

`test/test_ecmwf_levels.py` builds a synthetic ERA-5 model-level file in ECMWF's native descending-latitude ordering, with terrain rising from 0 m in the south to 3000 m in the north so a reflection is unambiguous, and loads it through the real `_load_model_level`. `z` is generated with `calcgeoh` exactly as RAiDER's own `_fetch` writes it.

Two assertions are load-bearing and **both fail on the pre-fix code** (the correlation comes out at −0.9986):

- surface height per latitude row matches the terrain it was built from
- `corr(surface height, −surface pressure) > 0.99`

The domain mean is deliberately not asserted on, for the reason above; every assertion is per-column.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
-* [TBD](https://github.com/dbekaert/RAiDER/pull/TBD) - Fixed a north-south flip in the ECMWF/ERA-5 model-level reader. `_load_model_level` reversed only the level axis of the geopotential cube (`z[::-1]`) while `t`, `q`, and `lnsp` had their latitude axis reversed, leaving heights mirrored in latitude relative to the meteorology. Surface height and surface pressure were anti-correlated (r = -0.54 where physics requires ~+1); sea-level ZTD over a 3-degree box spanned 1621-3233 mm instead of 2239-2395 mm. The error is antisymmetric in latitude, so it largely cancels in a domain average while being severe at individual points.
+* [811](https://github.com/dbekaert/RAiDER/pull/811) - Fixed a north-south flip in the ECMWF/ERA-5 model-level reader. `_load_model_level` reversed only the level axis of the geopotential cube (`z[::-1]`) while `t`, `q`, and `lnsp` had their latitude axis reversed, leaving heights mirrored in latitude relative to the meteorology. Surface height and surface pressure were anti-correlated (r = -0.54 where physics requires ~+1); sea-level ZTD over a 3-degree box spanned 1621-3233 mm instead of 2239-2395 mm. The error is antisymmetric in latitude, so it largely cancels in a domain average while being severe at individual points.
 
 ## [0.6.0]
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Fixed
+* [TBD](https://github.com/dbekaert/RAiDER/pull/TBD) - Fixed a north-south flip in the ECMWF/ERA-5 model-level reader. `_load_model_level` reversed only the level axis of the geopotential cube (`z[::-1]`) while `t`, `q`, and `lnsp` had their latitude axis reversed, leaving heights mirrored in latitude relative to the meteorology. Surface height and surface pressure were anti-correlated (r = -0.54 where physics requires ~+1); sea-level ZTD over a 3-degree box spanned 1621-3233 mm instead of 2239-2395 mm. The error is antisymmetric in latitude, so it largely cancels in a domain average while being severe at individual points.
+
 ## [0.6.0]
 ### Removed
 * [764](https://github.com/dbekaert/RAiDER/pull/764) - Removed Python 3.8 support. Python 3.9 is now the minimum version officially required to run RAiDER.
--- a/test/test_ecmwf_levels.py
+++ b/test/test_ecmwf_levels.py
@@ -1,0 +1,125 @@
+"""Regression tests for the ECMWF/ERA-5 model-level reader.
+
+These cover the orientation of the geopotential cube relative to the other
+fields. ECMWF delivers model-level data with latitude descending, and the
+reader reverses it to ascending; the height cube has to be reversed on the
+same axis as t/q/lnsp, or heights end up mirrored north-south against the
+meteorology.
+
+That failure mode is antisymmetric in latitude, so it very nearly cancels in
+a domain average -- a mean-based check will not catch it. The assertions here
+are deliberately per-column.
+"""
+import numpy as np
+import pytest
+import xarray as xr
+
+from RAiDER.constants import _g0
+from RAiDER.models.era5 import ERA5
+from RAiDER.utilFcns import calcgeoh
+
+
+# Terrain rising steeply from south to north, so that a latitude reflection is
+# unambiguous: the southern rows are at sea level, the northern rows are high.
+_LATS_DESC = np.array([36.0, 35.0, 34.0, 33.0])   # ECMWF order: descending
+_LONS = np.array([240.0, 241.0, 242.0])           # 0-360, as ECMWF delivers
+_TERRAIN = {36.0: 3000.0, 35.0: 2000.0, 34.0: 1000.0, 33.0: 0.0}
+
+
+def _write_synthetic_ml_file(path):
+    """Write a minimal ERA-5 model-level file in ECMWF's native ordering.
+
+    Mirrors what RAiDER's own _fetch writes: z is the *full* geopotential cube
+    (computed with calcgeoh), not the surface-only field CDS returns.
+    """
+    model = ERA5()
+    nlev = model._levels
+    nlat, nlon = len(_LATS_DESC), len(_LONS)
+
+    terrain = np.array([_TERRAIN[la] for la in _LATS_DESC])[:, None] * np.ones((nlat, nlon))
+    z_surface = terrain * _g0
+
+    t = np.full((nlev, nlat, nlon), 280.0)
+    q = np.full((nlev, nlat, nlon), 1e-3)
+
+    # surface pressure consistent with the terrain (barometric)
+    psurf = 101325.0 * np.exp(-terrain / 8400.0)
+    lnsp = np.log(psurf)
+
+    z_full, _, _ = calcgeoh(
+        lnsp=lnsp, t=t, q=q, z_surface=z_surface,
+        a=model._a, b=model._b, R_d=model._R_d, num_levels=nlev,
+    )
+
+    dims = ('valid_time', 'model_level', 'latitude', 'longitude')
+    ds = xr.Dataset(
+        data_vars={
+            't': (dims, t[np.newaxis]),
+            'q': (dims, q[np.newaxis]),
+            'z': (dims, np.asarray(z_full)[np.newaxis]),
+            # the reader takes lnsp as [0, 0], so it must carry both leading dims
+            'lnsp': (dims, np.broadcast_to(lnsp, (1, nlev, nlat, nlon)).copy()),
+        },
+        coords={
+            'valid_time': [0],
+            'model_level': np.arange(1, nlev + 1),
+            'latitude': _LATS_DESC,
+            'longitude': _LONS,
+        },
+    )
+    ds.to_netcdf(path)
+    return model
+
+
+@pytest.fixture
+def loaded_model(tmp_path):
+    f = tmp_path / 'ERA-5_synthetic_ml.nc'
+    model = _write_synthetic_ml_file(f)
+    model._ll_bounds = (32.5, 36.5, -120.5, -117.5)
+    model._load_model_level(f)
+    return model
+
+
+def test_model_level_latitudes_are_ascending(loaded_model):
+    """The reader is expected to flip ECMWF's descending latitudes."""
+    lats = loaded_model._lats[:, 0]
+    assert lats[0] < lats[-1]
+    assert np.allclose(np.sort(lats), lats)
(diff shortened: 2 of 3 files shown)


s-sasaki-earthsea-wizard:
@jlmaurer Confirmed — this fixes the latitude mirror. I re-ran the reproduction harness from the evidence repo against this branch at its current head 919d6b6, with the same raw ERA-5 ML file and the same frozen dependency set as the `dev`@1fb9e14 and #805@e7c8187 runs reported in #806 — only the RAiDER checkout changed (`raider==0.6.1.dev136+g919d6b699`).

**Result: the mirror is gone.**

| diagnostic (575 columns, 4.5°×4.5° box over Kanto) | dev@1fb9e14 (= #805@e7c8187) | this PR @919d6b6 |
|---|---|---|
| `z_surface` passed to `_calculategeoh` (bitwise ID) | lowest full level, **latitude-mirrored** | lowest full level, **own latitude** |
| height offset vs true-surface reference | −1252.41 / +1271.91 m | +9.46 / +10.19 m |
| regression `dh = a + b·h_own + c·h_mirror` | b = −1.000243, c = +0.999979 | b = −0.000243, c = −0.000021 |

The fitted coefficients changed by exactly +1.000000 for `h_own` and −1.000000 for `h_mirror`, removing the mirror component (`−h_own + h_mirror`) while the residual component (intercept 9.867 m, rmse 0.149 m) is unchanged to the last digit, and the measured post-fix offset lands inside the minimal-fix simulation from the #806 report (predicted +9.44 to +10.20 m). This matches the simulated axis-only fix quantitatively.

I also ran the new `test/test_ecmwf_levels.py` in both environments: 4/4 pass on this branch, and on pre-fix `dev`@1fb9e14 the two load-bearing assertions fail (correlation comes out at −0.9986, matching the PR description). The descending-latitude fixture closes the test-suite gap I noted on #805 — this regression can't silently return now.

Updated results are in the evidence repo, pinned at [`67087c0`](https://github.com/s-sasaki-earthsea-wizard/raider-arco-era5/tree/67087c036cb7bbaeb554715643cf89172dd8bc0f): see [`reports/ml_offset_regression_pr811.md`](https://github.com/s-sasaki-earthsea-wizard/raider-arco-era5/blob/67087c036cb7bbaeb554715643cf89172dd8bc0f/reports/ml_offset_regression_pr811.md).

### The remaining ~10 m offset (tracked separately)

The remaining ~+9.5–10.2 m is a separate, much smaller correctness issue: after the fix, `z[0]` is the lowest *full* model level where `_calculategeoh` expects the true surface field, so the reconstructed columns retain a near-uniform vertical offset of that size (measured above). This is the "additional item" from my #806 proposal and I don't think it should block this focused fix — it is about two orders of magnitude below the bug fixed here and roughly uniform in space. Since merging this PR closes #806, I've opened #812 to track the separate `z_surface` + loader-fallback change, including how it interacts with the reworked fetch in #805. Happy to follow up with that PR once this merges.

sssangha:
Great catch on this and thanks @s-sasaki-earthsea-wizard for the detailed verification!

Quick review notes where I distilled everything for easy reference later:
- **The Root Cause**: Spot-on fix. When `z` became a 3D cube in #751, `z[::-1]` started quietly flipping the vertical level axis instead of latitude, leaving `z` mirrored North-South against `t`, `q`, and `lnsp`.
- **Validation**: The diagnostics here look good. Seeing UNR GNSS ZTD residuals drop from ±400–500 mm down to -13 to -66 mm, and the surface terrain/pressure correlation jump back to +0.9992 shows the distortion is addressed. 
- **Testing**: Good we captured the flip in `test/test_ecmwf_levels.py` and made it clear in the notes.

LGTM! I'll approve

sssangha (approved):
LGTM, see my comment

s-sasaki-earthsea-wizard:
Thanks @sssangha ! 
That summary is a nice reference to have in one place.

PR #822, opened 2026-09-02
Fix ellipsoidal/geoid height datum handling for query points and DEM

royagrace:
## Summary
Height datums are now explicit. Heights previously flowed into weather-model
sampling unconverted, silently treating every source as geoid-referenced; each
source is now kept in its own datum and converted only where a consumer needs
the other.

## Motivation
The weather-model cube's z-axis is geoid-referenced (the ERA-5 convention), but
GNSS station heights are ellipsoidal. Mixing them displaced the integration
surface by the geoid undulation — 15–35 m in CONUS — for a **~5–9 mm positive
ZTD bias**. Measured at TORP: 2.343549 → 2.332483 m, −11.07 mm, matching the
≈−10 mm predicted from ZHD and an 8 km scale height.

## Changes
- Conversion happens at one place — weather-model sampling — via EPSG:9518
  (WGS84 + EGM2008). `tropo_delay` reads heights twice so the line-of-sight
  path keeps the ellipsoidal heights its ECEF geometry requires.
- Two new `height_group` options: `height_datum` for the query points (a
  station file's `Hgt_m` column), default `ellipsoidal`, since that is what
  GNSS positions carry; and `dem_height_datum` for any DEM whether downloaded
  or supplied via `dem`, default `geoid`, the native datum of GLO-30 and SRTM.
  The latter exists so a custom DEM already on the ellipsoid can say so.
- `download_dem` keeps GLO-30 on its native geoid
  (`dst_ellipsoidal_height=False`). DEMs RAiDER writes carry a
  `RAIDER_HEIGHT_DATUM` tag; reusing an untagged cached DEM warns, since one
  written by an earlier version holds ellipsoidal heights.
- `StationFile.readZ()` writes an authoritative `Hgt_datum` column when it
  fills heights from a DEM.
- If the EGM2008 grid is available neither locally nor from the PROJ CDN, the
  conversion is skipped with a warning naming the consequence rather than
  silently reintroducing the bias.
- Corrected stale docstrings in `geo_to_ht`, `weatherModel._get_heights`,
  `transformPoints`, and the delay cube's `z` attribute, all of which called
  geoid-referenced heights ellipsoidal.

## Upgrade notes
Existing configs keep working, with one exception worth checking:
- **Station files holding MSL heights need `height_datum: geoid`.** An
  unlabelled `Hgt_m` column is now read as ellipsoidal and converted. Files
  already carrying an `Hgt_datum` column are unaffected — that column wins.
- A config that supplies its own `height_group` must spell the new keys out,
  since `write_yaml` merges only the top level of a config.

API changes, for anyone calling RAiDER as a library:
- `LOS.setPoints()`: third parameter renamed `heights` → `ellipsoidal_heights`.
- `readZ()` now returns ellipsoidal heights by default instead of the source
  heights unconverted; pass `geoid_heights=True` for the geoid-referenced ones.
  The new parameter is defaulted, so existing calls still work.

Requirements:
- `pyproj` floor rises `2.2.0` → `3.1.0`, for `pyproj.network` and
  `Transformer.to_proj4()`.

## Testing
`test_llreader.py` covers both conversion directions, datum classification
(9518 vs 4979), the missing-grid fallback, and per-class datum defaults,
including real-EGM2008-grid tests that self-skip when the grid is unavailable.
`test_dem.py` covers `download_dem` against a stubbed stitcher, plus a
`long`-marked live smoke test. `test_delayFcns.py` pins the order in which
`tropo_delay` requests each datum. The ERA5 golden in `test_intersect.py` moves
2.343549 → **2.332483** — that is the bias being removed, not a regression.

## Needs revisiting
- **`RasterRDR`'s single 2-band lat/lon raster is unimplemented.** `__init__`
  always routes through `bounds_from_latlon_rasters`, which can't accept
  `lon_file=None`, so construction fails before `readZ` runs and its
  `NotImplementedError` guard for that case is unreachable. Harmless as-is;
  fixing it means building the feature, not repairing a bug.
- **The cube and ray-tracing paths still disagree on datum.** Pre-existing and
  out of scope — this PR fixes the point-AOI path. Tracked in #824, and called
  out under "Not fixed here" in the CHANGELOG so the `z` attribute's new
  `height above geoid` label isn't read as a claim about the cube.

## Type of change

## Checklist:


tools/RAiDER/cli/examples/template/template.yaml
```diff
@@ -77,6 +77,12 @@ aoi_group:
   lon_file:
   station_file:
 
+  ## Height datum of station_file's Hgt_m column, as an EPSG code.
+  ## 4979 (default): WGS84 ellipsoidal heights (e.g. from GNSS/UNR MAGNET or IGS20).
+  ## 4326: geoid-referenced (MSL) heights, e.g. from a DEM.
+  ## Ignored if station_file has an Hgt_datum column (that takes precedence) or is not set.
+  station_file_crs: 4979
```
jlmaurer: I think maybe adding the station_file_crs isn't actually as useful as I had hoped. Instead, a "height_datum" or something like that might be better.

jlmaurer: In the `height_group` in the template file and the and `HeightGroupUnparsed` might be good to also add a `dem_height_datum`, accepting 'geoid' or 'ellipsoidal'. Default geoid, since that is the native datum of GLO-30 and SRTM and of anything RAiDER downloads itself. That will take care of any custom DEM that gets passed.

test/test_losreader.py
```diff
@@ -280,7 +280,7 @@ def test_Zenith():
     with pytest.raises(RuntimeError):
         z.setPoints(lats=None)
 
-    z.setPoints(lats=lats, lons=lons, heights = hgts)
+    z.setPoints(lats=lats, lons=lons, ellipsoidal_heights=hgts)
```
jlmaurer, line 283: Ray-tracing still mixes datums. The cube z is now labelled "height above geoid", but _build_cube_ray uses the levels as ellipsoidal heights for ECEF ray geometry. Similarly Geocube.readZ() with its default ellipsoidal semantics feeds cube levels at [delay.py:79](vscode-webview://1pimgqvhtup2u9cvlnnm5rn71uv40j1spra8utqm6naf02dggpqu/tools/RAiDER/delay.py#L79). Pre-existing, but the PR should say so rather than imply the cube is consistently geoid.

tools/RAiDER/cli/types.py
```diff
@@ -112,6 +112,7 @@ class AOIGroupUnparsed:
     lat_file: Optional[str] = None
     lon_file: Optional[str] = None
     station_file: Optional[str] = None
+    station_file_crs: Union[int, str] = 4979
```
jlmaurer: same comment here as above.

jlmaurer: The `height_datum` argument could go under the `height_group` instead of here.

tools/RAiDER/llreader.py
```diff
@@ -228,23 +534,49 @@ def readZ(self):
...
+            # _parse_crs guarantees self._crs has a vertical axis, so this is
+            # always transformable. When the Hgt_datum column outranks
+            # self._crs and says these are ellipsoidal, a self._crs that
+            # disagrees would make src == dst and no-op, so trust the column.
+            src_crs = _WGS84_3D_CRS if _is_geoid_crs(self._crs) else self._crs
+            return _ellipsoidal_to_geometric(lats, lons, heights, src_crs)
+        return _geometric_to_ellipsoidal(lats, lons, heights)
 
 
 class RasterRDR(AOI):
```
jlmaurer, line 587: This should be ellipsoidal heights.

CHANGELOG.md
```diff
@@ -7,10 +7,19 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
...
 
 ### Changed
+* [822](https://github.com/dbekaert/RAiDER/pull/822) - Height datums are now explicit. Heights previously flowed into weather-model sampling unconverted, silently treating every source as geoid; each source is now kept in its own datum -- ellipsoidal for GNSS positions and ISCE geometry rasters, geoid for GLO-30 DEMs -- and converted only where a consumer needs the other. **Behaviour changes for existing inputs:**
+  * An unlabelled `Hgt_m` column is assumed ellipsoidal and converted; a file holding MSL heights must now declare `station_file_crs: 4326` in the config's `aoi_group` (default `4979`). `write_yaml` merges only the top level, so a config with its own `aoi_group` must spell this key out.
```
jlmaurer: This would get updated with the new syntax.

jlmaurer (changes requested):
see detailed comments.

jlmaurer:
@mgovorcin @dbekaert  just a note on the OPERA-relevance of this PR from a Claude Opus 5 review that I thought was useful: 

-----------------------------------
For the OPERA TROPO product as it is generated today, this PR is a no-op. Not "probably fine" — provably zero numerical impact, because OPERA does not execute any of the code the PR changes. But the PR does make an assertion about RAiDER's height semantics that, if correct (and I believe it is), means the TROPO product's height coordinate is mislabeled in the spec and in the netCDF attributes. That is a pre-existing documentation defect the PR surfaces rather than creates.

Separately, I found a blocking bug in the PR that is not OPERA-specific and that I reproduced end-to-end. It silently returns inf heights in exactly the kind of network-restricted environment a NASA SDS runs in.

1. Why OPERA is untouched

opera_tropo bypasses RAiDER's entire delay-calculation front end. It drives the weather model object directly in src/opera_tropo/core.py:

HRES() → _calculategeoh → _get_heights → _find_e → _uniform_in_z
       → _checkForNans → _get_wet_refractivity → _get_hydro_refractivity
       → _adjust_grid → _getZTD

Its only RAiDER imports are RAiDER.models.HRES, RAiDER.models.model_levels, and RAiDER.__version__. It never calls tropo_delay, never constructs an AOI, never touches llreader.py, dem.py, losreader.py, validators.py, or the CLI. Those are where 100% of this PR's logic lives.

The PR does touch two files OPERA's path runs through, models/weatherModel.py and utilFcns.py. I parsed both revisions and compared the ASTs with docstrings stripped:

tools/RAiDER/models/weatherModel.py  AST identical ignoring docstrings: True
tools/RAiDER/utilFcns.py             AST identical ignoring docstrings: True

Every change in those two files is a docstring or autopep8 whitespace. The bit-for-bit output of get_ztd() is unchanged.

There is also a version gap: conda-env.yml pins raider-base==0.5.3, and dev is already at 0.6.0. Nothing here reaches production until that pin moves.

2. The real OPERA implication: the height coordinate is labeled wrong

This is the part I'd actually escalate.

product_info.py declares the height coordinate as long_name="Ellipsoidal Height", description="Height above ellipsoid WGS84", and the spec document agrees: the height variable is described as ellipsoidal height levels above the WGS84 ellipsoid.

Those values come from hres_model._zs, produced by _get_heights → geo_to_ht. The old docstring claimed the result was ellipsoidal, reasoning that "by calculating the ellipsoid here we directly reference to WGS84." That reasoning is wrong, and the PR is right to strike it. geo_to_ht uses a latitude-varying Earth radius only to convert geopotential height to geometric height. The radius model affects the curvature correction; it does not move the zero reference. ECMWF geopotential is referenced to mean sea level, so the output is orthometric height above the geoid.

So the TROPO height levels are geoid-referenced, and the product says ellipsoidal. The discrepancy is the geoid undulation: roughly −107 to +86 m globally, about −30 m over much of CONUS. For a user sampling the cube at a DEM surface, a 30 m datum error is about 8–9 mm of ZTD at the near-surface hydrostatic gradient. This is the same magnitude the PR measured at TORP (−11.07 mm).

Two things follow:

**For differential InSAR (DISP-S1 and similar), the practical impact is small.** A fixed datum offset produces a nearly time-invariant ZTD bias that largely cancels in the double difference. What survives is the second-order term from temporal variation of the vertical gradient.
**For Cal/Val, it matters directly.** If TROPO validation compares product ZTD against GNSS ZTD sampled at station heights, and GNSS heights are ellipsoidal (UNR MAGNET, IGS20), then the validation is sampling the cube at the wrong level by the full undulation. That is precisely the bias this PR exists to fix on the station path. I'd check whether the TROPO Cal/Val comparison carries a systematic few-mm offset that correlates with geoid undulation. If it does, you've found the cause.

My recommendation: fix the metadata, not the values. Change long_name/description to geoid-referenced (and name the geoid model), and correct the spec. Do not "fix" it by shifting the height levels, which would silently change the product.

-----------------------------------

This is exactly what I found in my testing and validation results for the RAiDER paper.

jlmaurer (approved):
Looks good!

Issue #315, opened 2021-08-03
Confirm model levels and pressures levels is computed correctly for all models


dbekaert:

dbekaert:
@kam3545 can you a test over a small region to verify the mean biass issue has been resolved?

kam3545:
Hi @dbekaert 
I used below commands with new RAiDER codes after update repo and environment

raiderDownloadGNSS.py --out products_ex2 --date 20210517 --returntime '18:00:00' --bounding_box '32 42 -125 -114' --cpus all

raiderDelay.py --model NCMR --date 20210517 --time 18:00 --station_file ./products_ex2/UNRcombinedGPS_ztd.csv --zref 30000 -v 

I found attached error

Thanks
[log.TXT](https://github.com/dbekaert/RAiDER/files/7110859/log.TXT)

kam3545:
Hi @jlmaurer and @dbekaert 
New codes are working at global calculation, but I found larger errors as compared to the last version for NCMRRF/NCUM model. 
Thanks

dbekaert:
@kam3545 Can we confirm that NCUM product structure identical to the ECMWF model? I understood its model is a modification of the ECMWF model hence; similar variable definitions etc.

kam3545:
@dbekaert Please provide me a sample file of ECMWF HRES to identify changes in NCUM vs ECMWF model's variables. Both centers are using unified model (UM), definitely with large differences in implementation level.

PR #826, opened 2026-09-18
Restore the half-level average in calcgeoh's full-level pressure (-42 mm on ZTD)

mgovorcin:
## Description

`calcgeoh` returns the **half-level pressure above** each full level instead of the mean of the two half levels that bracket it. The averaging is present in the source, commented out:

```python
p[idx_level] = ph_lev  # + Ph_levplusone) / 2  # average pressure at half-levels above and below
```

ECMWF hybrid coordinates define pressure on half levels; the full-level value is the mean of the bracketing pair. Two lines below, `geopotential[idx_level]` is a **full**-level height, so the pressure paired with it must be the full-level value. As shipped, every level is handed a pressure that belongs roughly half a model layer higher — about **145 m through the mid troposphere** — so every level is too light, and the integrated hydrostatic delay with it.

The fix restores the line that already existed in this repository's history.

## Motivation and Context

**Magnitude.** On a moist US Standard Atmosphere sampled onto the 137-level HRES grid and integrated on the native output grid, the defect costs **−42.24 mm of ZTD**. Generalising the choice to a blend `p = (1−α)·ph_above + α·ph_below`, the error is linear at **+84.5 mm per unit α**, passing through zero at the ECMWF definition α = 0.5. The shipped code sits at α = 0.

| | ZTD error vs analytic truth |
|---|---:|
| α = 0 — as shipped | **−42.24 mm** |
| α = 0.5 — ECMWF definition | 0.00 mm |

Confirmed on real data: the residual against an independent closed form (Saastamoinen) goes from **29.9 mm RMS to 1.0 mm RMS**, with the bias going from −23.8 mm to −0.9 mm and the terrain ramp from +3.6 to −0.4 mm per km. Independently, the bias reproduces against GNSS ZTD at ~540 sites using two unrelated analysis centres, which agree with each other to 0.3 mm.

**History.** The averaging was correct for six months and then lost inside a file move:

| date | commit | state |
|---|---|---|
| 2019-08-09 | `485f0184` | no averaging — present from the start |
| 2021-06-30 | `fc7d0f59` "Fix pressure level calculation" | **correct** |
| 2021-12-04 | `459763124` move to `utilFcns.py` | commented out — the regression |
| 2022-10-03 | `44ebc41b` | replayed onto `dev`, reaching every release since |

`v0.1.0` is the only release that ever shipped it correct.

**Why this matters more after #816.** #816 fixed two errors in the same path that both push the delay *up*: the `k1·P/T` → `k1·P/Tv` virtual-temperature correction, and trapezoid → PCHIP quadrature. Both are right, and both removed error that had been partially masking this one. On a synthetic column the three combine as −42.24, +3.98 and +0.36 mm, so `dev` today carries a **larger net bias than v0.5.3 did** (−42.2 mm against −37.9 mm) until this line is restored.

Relevant to #315, which asks whether model-level pressures are computed correctly — for the model-level path, this is the answer.

## How Has This Been Tested?

Two new modules, both failing before the change and passing after.

**`test/test_calcgeoh_pressure.py`** — asserts the returned pressure is exactly the mean of the bracketing half levels (an exact relation, independent of the atmosphere), guards the specific regression so a revert cannot pass silently, and quantifies the delay cost from the coefficients alone. Self-contained: no weather file, no network.

**`test/test_downstream_opera_tropo.py`** — a downstream contract test. OPERA L4 TROPO-ZENITH (`opera-adt/tropo`) pins a RAiDER release and drives the HRES model-level path through internal entry points (`_calculategeoh`, `_get_heights`, `_find_e`, `_uniform_in_z`, the refractivity getters, `_getZTD`) rather than `calcDelays`, so nothing in this suite currently covers that sequence. This test walks the same sequence and checks the result against **Saastamoinen's closed form** rather than a recorded golden — a golden regenerated from a broken run defends the breakage, a closed form cannot be regenerated.

It is not asking for those internals to be frozen; refactor them freely. It asks that the delay they produce stays physically correct, so a PR that would move an operational product says so in CI.

Fixture: `test/opera_tropo_data/hres_columns_9.npz`, **8 KB** — nine HRES columns on the 137-level grid spanning sea level to 6.3 km terrain (surface pressure 1020 down to 469 hPa).

```
                          RMS      bias     ramp
  before this PR       33.1 mm  -23.9 mm  +4.70 mm/km   <- all three tests fail
  after                 1.0 mm   -0.9 mm  -0.36 mm/km   <- all three pass
```

```
$ pytest test/test_calcgeoh_pressure.py test/test_downstream_opera_tropo.py
6 passed
```

## Notes for reviewers

* **Fixture provenance.** The 8 KB fixture holds real ECMWF HRES `t`/`q`/`lnsp`/`z` values at nine points. If redistributing HRES-derived values here is a problem, I can regenerate it from an analytic atmosphere on the public A/B coefficient table — it reproduces the defect identically. Say the word.
* **Unrelated breaking change spotted.** `_calculategeoh`'s signature changed after v0.5.3 from `(self, z, lnsp)` to `(self, z_surface, lnsp, t, q)`. `opera-adt/tropo` still calls the two-argument form and will raise `TypeError` on upgrade, independent of this bug. Flagging rather than fixing here; happy to open a separate issue.
* This PR changes delay values for **every** model-level weather model, so processed fixtures elsewhere may need regenerating the way #816 did.
* CHANGELOG entry to follow once this has a number.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Marked breaking because computed delays change for every model-level model.

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jlmaurer (approved):
@dbekaert @mgovorcin this looks good to me; we'll need to regenerate golden test values post merging this. #822 (making sure everything uses the same height datum) also slightly changes the golden values, so this can be merged and I will re-generate those post merging both PRs.

